PR #30, opened 2018-03-14
Centralize and reorganize application settings

### LINKED ISSUE #16: Flat list of TestFixtures not working
This is a setting on the Tree settings page but it has no effect. We should either make it work or remove the setting.

### LINKED ISSUE #23: Centralize definition of Settings
Settings are accessed on an adhoc basis throughout the app. In particular, key is specified at each use as is the default value. This should be centralized for the settings we use while still making it possible to get and set arbitrary settings as necessary.

In addition, it would be good to separate the TestCentric settings from the NUnit settings, either by key or in a different file.

CharliePoole:
Fixes #23 
Fixes #16 

This was aimed at fixing issue #23. Issue #16 fell into place while I was doing it.

Before this PR, the app was using settings provided through the engine directly. At each point in the code where a setting was needed, we had to correctly specify the internal path to the setting, the Type of it's value and the default. It was possible to indicate these differently in different places, leading to unintended results.

With this fix, Settings are accessed using a hierarchy of properties that mimics the internal path used for each one. The property itself decides how to get or save the setting, what it's type is and what default value it has when not specified.

In addition, I revised entirely the hierarchy of settings to be more compact, while however keeping the same paths for settings used by the console runner. Note that the engine itself does not use it's own settings. Rather, each runner interprets them as needed and passes the effective settings in the test package.

This is ready for review. Tests are passing locally. We don't yet have a CI build.


immeraufdemhund:
I'll look at it here after I put kids to bed

CharliePoole:
Since @immeraufdemhund is having trouble building, could some other @Test-Centric/gui-team member please review this?

CharliePoole:
This is holding me up so I'm going to go ahead and merge. Later comments are OK.

src/TestCentric/testcentric.gui/SettingsPages/ProjectEditorSettingsPage.cs
```diff
@@ -32,35 +32,32 @@ namespace TestCentric.Gui.SettingsPages
...
             }
         }
 
         public override void ApplySettings()
         {
-            if (useNUnitEditorRadioButton.Checked)
-                Settings.RemoveSetting(EDITOR_PATH_SETTING);
-            else
-                Settings.SaveSetting(EDITOR_PATH_SETTING, editorPathTextBox.Text);
+            Settings.TestCentric.ProjectEditorPath = useNUnitEditorRadioButton.Checked
```
mikkelbu: I think that this should be inverted. If we have selected `useNUnitEditorRadioButton` then we should use `nunit-editor.exe`, but if `useNUnitEditorRadioButton` is not selected then we should use the text in `editorPathTextBox`

src/TestCentric/testcentric.gui/SettingsPages/SettingsPage.cs
```diff
@@ -143,7 +143,7 @@ protected override void OnLoad(EventArgs e)
 
 			if ( !DesignMode )
 			{
-                Settings = (ParentForm as SettingsDialogBase)?.Settings;
+                Settings = new UserSettings((ParentForm as SettingsDialogBase)?.Settings);
```
mikkelbu: It can be the case that `ParentForm` will always be a `SettingsDialogBase`, but I would rather fail early here than sending null to `UserSettings` (and in turn let `SettingsGroup` fail with a NRE).

src/TestCentric/testcentric.gui/TestCentricMainForm.cs
```diff
@@ -941,14 +946,14 @@ private void decreaseFontMenuItem_Click(object sender, System.EventArgs e)
 
         private void applyFont( Font font )
         {
-            Font = font;
+            if (_displayFormat == "Mini")
```
mikkelbu: Both the cases below this `if` have the same behaviour?

src/TestModel/model/Settings/EngineSettings.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// </summary>
+    public class EngineSettings : SettingsGroup
+    {
+        public EngineSettings(ISettings settings)
+            : base(settings, "Options.TestLoader") { }
+
+        public bool ReloadOnChange
+        {
+            get { return GetSetting("ReloadOnChange", true); }
+            set { SaveSetting("ReloadOnChange", value); }
```
mikkelbu: Perhaps we should use `nameof(ReloadOnChange)` here or perhaps use a constant string (just to avoid problems with persisting with one key and retrieving with another key). Similar comment applies to all the settings below

src/TestModel/model/Settings/RecentProjectsSettings.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    public class RecentProjectsSettings : SettingsGroup
+    {
+        public RecentProjectsSettings(ISettings settings)
+            : base(settings, "Options") { }
+
+        public int MaxFiles
+        {
+            get { return GetSetting("MaxFiles", 24); }
+            set { SaveSetting("MaxFiles", value); } // TODO: Handle null
```
mikkelbu: `value` cannot be null due to its type.

src/TestModel/model/TestCentric.Gui.Model.csproj
```diff
@@ -44,12 +44,25 @@
       <Private>True</Private>
     </Reference>
     <Reference Include="System" />
+    <Reference Include="System.Drawing" />
```
mikkelbu: Is this necessary?

CharliePoole:
FYI, the default values of the settings were not preserved by the old code. I found several spots where the same settings were accessed specifying different defaults. 😄

src/TestCentric/testcentric.gui/Controls/ErrorDisplay.cs
```diff
@@ -83,6 +82,9 @@ protected override void Dispose( bool disposing )
 		}
 
 		#region Properties
+
+        private UserSettings UserSettings { get; set; }
```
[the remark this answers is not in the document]
CharliePoole: Changes are getting the current standard settings, but the original files are pretty old and I tried to avoid reformatting everything.

src/TestModel/model/Settings/TestTreeSettings.cs
```diff
@@ -0,0 +1,60 @@
...
+// ***********************************************************************
+
+using System.Drawing;
+using NUnit.Engine;
+
+namespace TestCentric.Gui.Model.Settings
+{
+    public class TestTreeSettings : SettingsGroup
+    {
+        private static readonly Font DefaultFixedFont = new Font(FontFamily.GenericMonospace, 8.0F);
```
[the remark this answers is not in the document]
CharliePoole: Fixed font is used in the error display and the console output.

src/TestCentric/testcentric.gui/Controls/NotRunTree.cs
```diff
@@ -36,14 +36,12 @@ public class NotRunTree : TreeView, IViewControl
 	{
         const string CHILD_IGNORED_MESSAGE = "One or more child tests were ignored";
 
-        private ISettings _userSettings;
+        private UserSettings _userSettings;
```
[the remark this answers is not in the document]
CharliePoole: Will drop.

src/TestCentric/testcentric.gui/Controls/TestSuiteTreeView.cs
```diff
@@ -987,8 +976,7 @@ private void HideTestsUnderNode(TestSuiteTreeNode node)
         /// <returns>DisplayStyle to be used</returns>
         private DisplayStyle GetDisplayStyle()
 		{
-			DisplayStyle initialDisplay = (TestSuiteTreeView.DisplayStyle)
-				UserSettings.GetSetting( "Gui.TestTree.InitialTreeDisplay", DisplayStyle.Auto );
+			DisplayStyle initialDisplay = (DisplayStyle)UserSettings.TestCentric.TestTree.InitialTreeDisplay;
```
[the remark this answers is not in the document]
CharliePoole: I'd be in favor of that. Right now, it's only weakly typed for enums.

src/TestCentric/testcentric.gui/Controls/ResultTabs.cs
```diff
@@ -236,7 +237,9 @@ protected override void OnFontChanged(EventArgs e)
 
         public void InitializeView(ITestModel model, TestCentricPresenter presenter)
         {
-            Settings = model.GetService<ISettings>();
+            Settings = model.Services.UserSettings;
+
+            tabControl.SelectedIndex = Settings.TestCentric.SelectedTab;
```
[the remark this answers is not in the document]
CharliePoole: I got annoyed that it wasn't remembering the last tab I had selected. 😄 The old gui didn't do it either.

src/TestCentric/testcentric.gui/Controls/NotRunTree.cs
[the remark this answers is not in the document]
CharliePoole: Will drop this.

src/TestCentric/testcentric.gui/SettingsPages/SettingsPage.cs
CharliePoole: It's a good point. Will do.

src/TestCentric/testcentric.gui/TestCentricMainForm.cs
CharliePoole: Will check. I was working towards that, but I didn't think I had gotten there yet.

src/TestModel/model/Settings/EngineSettings.cs
CharliePoole: I was trying to think of a good way to do this and even came up with an approach that used an internal setting object with the key specified in one place only. But the code ended up being twice as long as what I have now. Using `nameof` would be a good compromise. Will do it in another PR.

src/TestModel/model/Settings/RecentProjectsSettings.cs
CharliePoole: Right!

src/TestModel/model/TestCentric.Gui.Model.csproj
CharliePoole: Yes, because we are storing and retrieving fonts in the settings. We could avoid it by storing and retrieving strings and doing the conversion in the gui part, but it seemed nicer to just return a Font from the settings property.

CharliePoole:
Thanks @mikkelbu . I will put together another PR with a few changes.

src/TestCentric/testcentric.gui/Controls/ErrorDisplay.cs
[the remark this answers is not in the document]
ManfredLange: Perhaps we want to bite the bullet and reformat the entire code base once? It's still an early stage of the project and we could use a separate commit or even branch for it to reduce the impact (reporting to changes etc.)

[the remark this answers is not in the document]
CharliePoole: A separate branch and merge when we are at some point without outstanding WIP would make sense.

src/TestModel/model/Settings/TestTreeSettings.cs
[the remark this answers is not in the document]
mikkelbu: But I don't think that this `DefaultFixedFont` is used for anything. It is private at not mentioned in the rest of the file (and we don't reflect for this field either AFAICT).

PR #31, opened 2018-03-15
Remove InternalTraceLevel setting; fixes #20

### LINKED ISSUE #20: InternalTrace level not honored
When set in the user settings, this doesn't appear to be effective.

CharliePoole:
This change simply acknowledges the fact that it's not currently possible to use the settings managed by the engine to set the trace level for the engine. We will use only the command-line `--trace` option for this purpose until something changes.

### DIFF
--- a/src/TestCentric/testcentric.gui/AppEntry.cs
+++ b/src/TestCentric/testcentric.gui/AppEntry.cs
@@ -66,10 +66,13 @@ namespace TestCentric.Gui
                 return 2;
             }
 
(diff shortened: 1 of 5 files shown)


CharliePoole:
The original code, from the V2 GUI, assumed that it knew the Default value was Off within the framework. Even though that will probably always be the case, it still seemed like an unwarranted assumption for the runner to make, particularly a runner that aims at supporting multiple frameworks.

Issue #27, opened 2018-03-11
Create CI Build Script


CharliePoole:
Should run on AppVeyor and Travis and create a Chocolatey install.

CharliePoole:
I see that the initial CI build that we merged is building a Debug version. Should be release.

CharliePoole:
Closing this based on #37, which created the AppVeyor build. Will create a separate issue for the Travis build.

PR #37, opened 2018-03-21
Initial CI Build

CharliePoole:
Partial fix to #27 

This is just the AppVeyor build. I have had some issues with the Travis build, so I decided to do a PR that just includes AppVeyor.

The build is failing because one test case repeatedly fails. I've run successfully locally, with no failures and have produced a working packages.

Since the build itself is doing what it is supposed to do by reporting the test failure, I propose we merge this. That will give us a build for other PRs although we'll be forced to examine the reason for any failure until we fix the one we have. I'd like to have more eyes on that anyway.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,13 @@
+branches:
+  except:
+    - /travis-.*/
+
+build_script: 
+  - ps: .\build.ps1 -Target "Appveyor" -Configuration "Release"
+  
+# disable built-in tests.
+test: off
+
+artifacts:
+- path: package\*.zip
+- path: package\*.nupkg
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,7 +3,7 @@ branches:
     - /travis-.*/
 
 build_script: 
-  - ps: .\build.ps1 -Target "Appveyor" -Configuration "Release"
+  - ps: .\build.ps1 -Target "Appveyor" -Configuration "Debug"
   
 # disable built-in tests.
 test: off
--- a/build.cake
+++ b/build.cake
@@ -1,0 +1,197 @@
+#tool nuget:?package=NUnit.ConsoleRunner&version=3.8.0
+//////////////////////////////////////////////////////////////////////
+// ARGUMENTS
+//////////////////////////////////////////////////////////////////////
+
+var target = Argument("target", "Default");
+var configuration = Argument("configuration", "Release");
+
+//////////////////////////////////////////////////////////////////////
+// SET PACKAGE VERSION
+//////////////////////////////////////////////////////////////////////
+
+var version = "0.1";
+var modifier = "";
+var dbgSuffix = configuration == "Debug" ? "-dbg" : "";
+
+//////////////////////////////////////////////////////////////////////
+// DEFINE RUN CONSTANTS
+//////////////////////////////////////////////////////////////////////
+
+// Directories
+var PROJECT_DIR = Context.Environment.WorkingDirectory.FullPath + "/";
+var PACKAGE_DIR = PROJECT_DIR + "package/";
+var CHOCO_DIR = PROJECT_DIR + "choco/";
+var BIN_DIR = PROJECT_DIR + "bin/" + configuration + "/";
+
+// Packaging
+var PACKAGE_NAME = "testcentric-gui";
+var PACKAGE_VERSION = version + modifier + dbgSuffix;
+
+// Solution
+var SOLUTION = "testcentric-gui.sln";
+
+// Test Assembly Pattern
+var ALL_TESTS = BIN_DIR + "*.Tests.dll";
+
+//////////////////////////////////////////////////////////////////////
+// CLEAN
+//////////////////////////////////////////////////////////////////////
+
+Task("Clean")
+    .Does(() =>
+{
+    CleanDirectory(BIN_DIR);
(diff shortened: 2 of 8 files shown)


CharliePoole:
Maybe one of you other guys who has used this on NUnit could take a look as well?

immeraufdemhund:
i'm trying to reproduce the error on my machine. the only difference I see in settings if number of test workers. appveyor has 2 and I have 4. I've tried changing the name of the file being loaded for TestResource which gives a NullReferenceException and not a IndexOutOfBoundsException. I've tried deleting that file then making it 0 bytes which the test passes. My only other guess is file format? I sort of don't believe that would be the problem because the stacktrace shows the problem happening on the second `TestResource`... Unless it can't read the BOM...

immeraufdemhund:
Nope that can't be it either because all 3 files are utf-8 with BOM... so yeah. No clue. sorry.

mikkelbu:
I can take a look at the implementation tonight (I approx. 7 hours). Wrt. the failing test I think we can do the following:
* Try running the CI in debug to get a better stacktrace
* Perhaps try to run the CI in another culture (I cannot see which culture appveyor is using), since we are checking against "\n" vs. '\n' in the code)

But I'm also perplexed about this 😄

CharliePoole:
Since the problem doesn't seem to be with the changes in this PR, I'm still inclined to merge.

CharliePoole:
@mikkelbu I'd like to figure out how to keep up with the best bootstrap scripts from Cake. Unfortunately, they make it a bit hard to keep up, since they maintain copies in multiple places, which are not always the same. Wish they would add a header to the scripts that identified their provenance. 

In any case, we don't make use of bakery, so for the moment let's not worry. This issue isn't closing yet so we may update in the next PR. What would you consider the latest scripts from Cake?

CharliePoole:
I removed build.sh and .travis.yml when I did this partial fix to the issue. I'll concentrate separately on getting Linux to work, which is likely to make us update build.sh (and probably build.ps1 to keep in sync).

immeraufdemhund:
oh.. i use build.sh in git bash, and i'm windows.

CharliePoole:
Interesting. I never tried doing it that way.

mikkelbu:
I did an attempt in #38 to figure out the difference between a local build and a build on appveyor and the only difference is in the test `Any_Formatter_Should_Format_Any_Text` were I get different numbers when I print out `LineCount` and the numbers of `'\n'` in `sourceCode`.

On my machine
```
=> NUnit.UiException.Tests.CodeFormatters.TestGeneralCodeFormatter.Any_Formatter_Should_Format_Any_Text
LineCount: 1
numNewLines: 0
LineCount: 27
numNewLines: 27
LineCount: 59
numNewLines: 58
```
On appveyor
```
=> NUnit.UiException.Tests.CodeFormatters.TestGeneralCodeFormatter.Any_Formatter_Should_Format_Any_Text
LineCount: 1
numNewLines: 0
LineCount: 23
numNewLines: 27
```

So the Line.Count is less on appveyor. I'll diagnose some more.

@CharliePoole Wrt. the newest version I don't think it is written anywhere. I just took the the file mentioned in the reference in the top of `build.ps1` (i.e. https://github.com/cake-build/resources).

CharliePoole:
Shall I leave it at Debug for now?

CharliePoole:
Nice debugging, btw!

CharliePoole:
Answering my own question... I'll leave Debug enabled while we figure stuff out.

mikkelbu:
I think the problem is that git changes the line-endings when I check out `Basic.cs` on a windows machine (and actually only 23 of the newlines are Windows newlines), since on my machine `Basic.cs` has 27 `0D`s and `0A`s (carriage return and line endings). I need to read some more about git and line-endings (e.g. https://help.github.com/articles/dealing-with-line-endings/)

CharliePoole:
I could have messed this up when I moved stuff from V2 to this project. I'm not sure V2 used the same line endings as we have been using for V3. Nice work.

Can you come up with a recommendation as to how we should store line endings? Since we are cross-platform, I think the local value has to vary with the hosting environment.

mikkelbu:
@CharliePoole This solves the issue https://github.com/TestCentric/testcentric-gui/pull/38/commits/a0b0fae0f4a3d7686a426e6117bfab2116ddf891 - now all tests pass, but it fails in "PackageZip", since no CHANGES.txt exists on that branch.
But I'm unsure whether this is the best solution and I've not run into problems like this before (mostly using mercurial + windows at work). I'll try experimenting some more.

CharliePoole:
Is .gitattributes in the project root wrong then?

mikkelbu:
Actually, I think they are ok, but again I'm more versed in mercurial. So I propose that we keep our settings as they are and then make the code more robust wrt. line endings. I think the code - as it is now - will fail on Linux in the same place and also on Windows if we read in a file with both kinds of line endings.

I'll try to supply a PR tonight (in 11 hours time) where I make the line ending handling in `PlainTextCodeFormatter.cs` and `DefaultTextManager.cs` more robust (and at least consistent).

Ps. At some point it could be that we should consider whether a lot of the lexing, parsing and presentation under `GuiException\UiException` could be replaced with e.g. Roslyn, to minimise our codebase and the maintenance burden.

PR #40, opened 2018-03-22
Make the code robust wrt multiple newlines

mikkelbu:
This change makes it possible to successfully read
multiple consecutive (linux) newlines without
throwing an exception.


CharliePoole:
This looks good to me.

PR #41, opened 2018-03-28
Add functionality for buttons

### LINKED ISSUE #28: "Clear All" and "Check Failed" has no effect
The buttons "Clear All" and "Check Failed" have no effect on the selection in the tree (when "Show CheckBoxes" is enabled)

mikkelbu:
The functionality behind the buttons 'Clear All' and
'Check failed' has been enabled. As part of this I've
changed the comparison in `CheckFailedNodesVisitor.Visit`
as otherwise reference equality is used.

I've tried to follow the general NUnit code guidelines
which makes the diff look strange due to mix of spaces
and tabs in the existing code.

Fixes #28

### DIFF
--- a/src/TestCentric/testcentric.gui/Controls/TestSuiteTreeView.cs
+++ b/src/TestCentric/testcentric.gui/Controls/TestSuiteTreeView.cs
@@ -1120,9 +1120,9 @@ namespace TestCentric.Gui.Controls
         {
             public override void Visit(TestSuiteTreeNode node)
             {
-                if (!node.Test.IsSuite && node.HasResult &&
-                       (node.Result.Outcome.Equals(ResultState.Failure) ||
-                        node.Result.Outcome.Equals(ResultState.Error)))
+                if (!node.Test.IsSuite &&
+                    node.HasResult &&
+                    node.Result.Outcome.Status == TestStatus.Failed)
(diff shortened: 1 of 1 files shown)


src/TestCentric/testcentric.gui/Controls/TestSuiteTreeView.cs
```diff
@@ -1102,39 +1102,37 @@ private void LoadAlternateImage(int index, string name, string imageSet)
...
-            if (!node.Test.IsSuite && node.HasResult &&
-                   (node.Result.Outcome == ResultState.Failure ||
-                    node.Result.Outcome == ResultState.Error))
+            public override void Visit(TestSuiteTreeNode node)
             {
-                node.Checked = true;
-                node.EnsureVisible();
+                if (!node.Test.IsSuite && node.HasResult &&
+                       (node.Result.Outcome.Equals(ResultState.Failure) ||
+                        node.Result.Outcome.Equals(ResultState.Error)))
```
CharliePoole: For NUnit 3, I think this works better if you check that `node.Result.Outcome.Status == TestStatus.Failed`. That covers a wider range of failed tests than the equality check on `Outcome`. The initial two conditions have to remain, of course.

PR #43, opened 2018-03-28
Use number of recent files to restrict menu

mikkelbu:
I don't like how I need to "copy" the constant (24) from the settings class.


CharliePoole:
LGTM for now. We may want to revisit what we use as the default at some point.

Note that "24" has two uses here. It's the maximum number that the engine will save for us and it's also the maximum (and default) number we are displaying.

PR #48, opened 2018-06-06
Pass load settings to engine correctly

### LINKED ISSUE #19: Principal Policy not working
This should be tested to be sure, but it doesn't appear to have any effect when set.

### LINKED ISSUE #24: Setting number of agents has no effect
A setting has been added to specify the number of agents but there is currently no backend implementation - the setting is ignored.

CharliePoole:
Fixes #24
Fixes #19

### DIFF
--- a/src/TestCentric/testcentric.gui/SettingsPages/AdvancedLoaderSettingsPage.cs
+++ b/src/TestCentric/testcentric.gui/SettingsPages/AdvancedLoaderSettingsPage.cs
@@ -27,13 +27,13 @@ using System.Windows.Forms;
 
 namespace TestCentric.Gui.SettingsPages
 {
-	public class AdvancedLoaderSettingsPage : SettingsPage
-	{
-		private System.Windows.Forms.Label label3;
-		private System.Windows.Forms.GroupBox groupBox3;
+    public class AdvancedLoaderSettingsPage : SettingsPage
+    {
+        private System.Windows.Forms.Label label3;
+        private System.Windows.Forms.GroupBox groupBox3;
         private System.Windows.Forms.CheckBox disableShadowCopyCheckBox;
-		private System.Windows.Forms.Label label2;
-		private System.Windows.Forms.HelpProvider helpProvider1;
+        private System.Windows.Forms.Label label2;
+        private System.Windows.Forms.HelpProvider helpProvider1;
         private CheckBox principalPolicyCheckBox;
         private Label label7;
         private Label label6;
@@ -42,36 +42,36 @@ namespace TestCentric.Gui.SettingsPages
         private Label label1;
         private System.ComponentModel.IContainer components = null;
 
-		public AdvancedLoaderSettingsPage( string key ) : base( key )
-		{
-			// This call is required by the Windows Form Designer.
-			InitializeComponent();
+        public AdvancedLoaderSettingsPage( string key ) : base( key )
+        {
+            // This call is required by the Windows Form Designer.
+            InitializeComponent();
 
-			// TODO: Add any initialization after the InitializeComponent call
-		}
+            // TODO: Add any initialization after the InitializeComponent call
+        }
 
-		/// <summary>
-		/// Clean up any resources being used.
-		/// </summary>
-		protected override void Dispose( bool disposing )
-		{
-			if( disposing )
-			{
-				if (components != null) 
-				{
-					components.Dispose();
-				}
-			}
-			base.Dispose( disposing );
-		}
+        /// <summary>
+        /// Clean up any resources being used.
+        /// </summary>
+        protected override void Dispose( bool disposing )
+        {
+            if( disposing )
+            {
+                if (components != null) 
+                {
+                    components.Dispose();
+                }
+            }
+            base.Dispose( disposing );
+        }
 
-		#region Designer generated code
-		/// <summary>
-		/// Required method for Designer support - do not modify
-		/// the contents of this method with the code editor.
-		/// </summary>
-		private void InitializeComponent()
-		{
+        #region Designer generated code
+        /// <summary>
+        /// Required method for Designer support - do not modify
+        /// the contents of this method with the code editor.
+        /// </summary>
+        private void InitializeComponent()
+        {
             System.ComponentModel.ComponentResourceManager resources = new System.ComponentModel.ComponentResourceManager(typeof(AdvancedLoaderSettingsPage));
             this.label3 = new System.Windows.Forms.Label();
             this.groupBox3 = new System.Windows.Forms.GroupBox();
@@ -201,11 +201,11 @@ namespace TestCentric.Gui.SettingsPages
             this.ResumeLayout(false);
             this.PerformLayout();
 
-		}
-		#endregion
+        }
+        #endregion
 
-		public override void LoadSettings()
-		{
+        public override void LoadSettings()
+        {
             disableShadowCopyCheckBox.Checked = !Settings.Engine.ShadowCopyFiles;
 
             principalPolicyCheckBox.Checked = principalPolicyListBox.Enabled =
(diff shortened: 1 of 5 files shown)


CharliePoole:
@mikkelbu I've been doing small changes without review, but this one should be looked at. Basically, we were doing nothing to pass settings on to the engine for use, I implemented it for process, domain and agents but we'll have to look at others as well.

CharliePoole:
@TestCentric/gui-team Needs a review from one of you!

src/TestCentric/testcentric.gui/SettingsPages/TestLoaderSettingsPage.cs
```diff
@@ -258,6 +258,8 @@ public override void LoadSettings()
 			bool singleDomain = Settings.Engine.DomainUsage == "Single";
 			multiDomainRadioButton.Checked = !singleDomain;
 			singleDomainRadioButton.Checked = singleDomain;
+
+            multiDomainRadioButton.Enabled = singleDomainRadioButton.Enabled = !multiProcessRadioButton.Checked;
```
immeraufdemhund: mixed tabs and spaces makes this file look like it isn't indented properly

src/TestCentric/testcentric.gui/SettingsPages/TestLoaderSettingsPage.cs
```diff
@@ -303,7 +305,8 @@ public override bool HasChangesRequiringReload
 			{
                 return
                     Settings.Engine.ProcessModel != SelectedProcessModel() ||
-                    Settings.Engine.DomainUsage != SelectedDomainUsage();
+                    Settings.Engine.DomainUsage != SelectedDomainUsage() ||
+                    Settings.Engine.Agents != numberOfAgentsUpDown.Value;
```
immeraufdemhund: i'm assuming that Agents is a number and numberOfAgentsUpDown.Value is a nullable number. the following test works as well.

```
    int a = 1;
    int? b = null;
    
    Console.WriteLine(a==b); // False
    b = 1;
    Console.WriteLine(a == b); // True
    b = 2;
    Console.WriteLine(a == b); // False
```

so you could potentially change it to be `Settings.Engine.Agennts != numberOfAgentsUpDown`

src/TestModel/model/TestModel.cs
```diff
@@ -322,18 +322,31 @@ public void Dispose()
...
+
             //if (Options.InternalTraceLevel != null)
             //    package.AddSetting(EnginePackageSettings.InternalTraceLevel, Options.InternalTraceLevel);
 
             // We use shadow copy so that the user may re-compile while the gui is running.
             package.AddSetting(EnginePackageSettings.ShadowCopyFiles, true);
 
-            // Temp settings for testing
-            package.AddSetting(EnginePackageSettings.ProcessModel, "Single");
+            //// Temp settings for testing
```
immeraufdemhund: double comment

src/TestModel/model/TestModel.cs
```diff
@@ -322,18 +322,31 @@ public void Dispose()
...
+            //package.AddSetting(EnginePackageSettings.ProcessModel, "Single");
```
immeraufdemhund: why are we commenting out code instead of just deleting it? Or atleast writing why we are commenting out code?

immeraufdemhund:
Note that I don't have time to test the changes manually out though, so maybe someone else can do a sanity check on the build.

mikkelbu:
@CharliePoole I can take a loot at this tonight (10 hours) or perhaps in the weekend, as currently I'm abroad.

src/TestCentric/testcentric.gui/SettingsPages/TestLoaderSettingsPage.cs
CharliePoole: It's a spin control, and it's value is decimal - which seems odd BTW.

src/TestModel/model/TestModel.cs
CharliePoole: It was a temporary setting to test in-process before the process setting worked. No longer needed.

CharliePoole:
@immeraufdemhund Thanks. I made some of the changes you suggested.

@mikkelbu I'd appreciate it. Feel free to review the entire approach to how default settings are merged in. 😄

src/TestModel/model/TestModel.cs
```diff
@@ -322,19 +322,29 @@ public void Dispose()
...
         {
             var package = new TestPackage(testFiles);
+            var engineSettings = Services.UserSettings.Engine;
 
             // We use AddSetting rather than just setting the value because
             // it propagates the setting to all subprojects.
 
+            package.AddSetting(EnginePackageSettings.ProcessModel, engineSettings.ProcessModel);
+
+            // Don't set DomainUsage for Multiple to avoid engine error
```
mikkelbu: I don't see similar validation in the console, but if I try `.\nunit3-console.exe --process=Multiple --domain=Multiple .\nunit.engine.tests.dll .\nunit3-console.tests.dll` then it just exists immediately with `Overall result: Unknown` (and I cannot see from the logs what is wrong). Not that I'm against removing possibility for errors 😄, I just want to understand how the settings relate to each other.

Do we have any information about this combination is illegal (I could not find it in the wiki).

src/TestModel/model/TestModel.cs
```diff
@@ -322,19 +322,29 @@ public void Dispose()
...
+            if (engineSettings.ProcessModel != "Multiple")
+                package.AddSetting(EnginePackageSettings.DomainUsage, engineSettings.DomainUsage);
+            // Don't bother checking Agents unless we are running multiple processes
+            else if (engineSettings.Agents > 0)
+                package.AddSetting(EnginePackageSettings.MaxAgents, engineSettings.Agents);
```
mikkelbu: Again, I see no such validation in the console, but I think this makes sense.

mikkelbu:
And additional question. If I set process mode to multiple, like below, and I have a .nunit-file consisting of more than one dll shouldn't the dlls be tested in parallel?
image

src/TestCentric/testcentric.gui/SettingsPages/AdvancedLoaderSettingsPage.cs
```diff
@@ -220,8 +220,8 @@ public override void ApplySettings()
             Settings.Engine.SetPrincipalPolicy = principalPolicyCheckBox.Checked;
 
             Settings.Engine.PrincipalPolicy = principalPolicyCheckBox.Checked
-                ? (PrincipalPolicy)principalPolicyListBox.SelectedIndex
-                : PrincipalPolicy.UnauthenticatedPrincipal;
+                ? (string)principalPolicyListBox.SelectedItem
+                : "UnauthenticatedPrincipal";
```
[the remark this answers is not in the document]
CharliePoole: Probably better. I'm working in this code on a different branch as well, so I'll try to change it in a way that merges easily.

src/TestModel/model/TestModel.cs
```diff
@@ -322,19 +322,29 @@ public void Dispose()
...
+
+            if (engineSettings.SetPrincipalPolicy)
+                package.AddSetting(EnginePackageSettings.PrincipalPolicy, engineSettings.PrincipalPolicy.ToString());
```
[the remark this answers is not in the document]
CharliePoole: Right!

src/TestModel/model/TestModel.cs
```diff
@@ -322,19 +322,29 @@ public void Dispose()
...
             var package = new TestPackage(testFiles);
+            var engineSettings = Services.UserSettings.Engine;
 
             // We use AddSetting rather than just setting the value because
             // it propagates the setting to all subprojects.
 
+            package.AddSetting(EnginePackageSettings.ProcessModel, engineSettings.ProcessModel);
+
+            // Don't set DomainUsage for Multiple to avoid engine error
+            if (engineSettings.ProcessModel != "Multiple")
```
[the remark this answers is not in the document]
CharliePoole: I think it's in the API assembly. I'll check.

src/TestModel/model/TestModel.cs
CharliePoole: It's shouldn't be illegal, but rather meaningless. That is, the domain setting should simply be ignored but isn't. Note that in this project (TestCentric) I'm trying to treat the engine as a black box, so if we find anything that doesn't work we should just not use it. I'm not being 100% successful at it and some of my changes to PrincipalPolicy may need to be backed out because of how the engine works - I made them thinking I could just change the engine, but of course I can't!

CharliePoole:
Regarding nunit project assemblies running in parallel: I think it should work as you describe, but I'm not sure it really does at present.

src/TestModel/model/Settings/EngineSettings.cs
```diff
@@ -84,9 +84,9 @@ public bool SetPrincipalPolicy
             set { SaveSetting(nameof(SetPrincipalPolicy), value); }
         }
 
-        public PrincipalPolicy PrincipalPolicy
+        public string PrincipalPolicy
         {
-            get { return GetSetting(nameof(PrincipalPolicy), PrincipalPolicy.UnauthenticatedPrincipal); }
+            get { return GetSetting(nameof(PrincipalPolicy), "UnauthenticatedPrincipal"); }
```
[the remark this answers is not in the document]
CharliePoole: Actually, since there is a PrincipalPolicy in the class, I had to use System.Security.Principal.PrincipalPolicy!

src/TestModel/model/TestModel.cs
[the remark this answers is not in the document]
CharliePoole: Nope, you're right it's in the engine which we don't want to reference.

CharliePoole:
Of course, whether we have editor.config or not, changing tabs to spaces is going to generate a lot of change. I started doing the changes as I hit each file after you made an earlier comment about mixed tabs and spaces. Maybe it would be better to just do it all in a separate PR.

I have nothing against editor.config. I would rather not maintain multiple files that control the layout if we can avoid it and I also want to stay away from using it take all the decisions out of the hands of the developer, as seems to have happened in the NUnit project. 😞

PR #53, opened 2018-06-09
Remove NRE when changing ShowCheckBoxes

### LINKED ISSUE #52: Show CheckBoxes throws NRE
If one selects `View => Tree => Show CheckBoxes` then a NRE is thrown. The problem is that `TestTree.UserSettings` is never set (and we don't need this field, since we have `TestTree.Model`.

```
See the end of this message for details on invoking 
just-in-time (JIT) debugging instead of this dialog box.

************** Exception Text **************
System.NullReferenceException: Object reference not set to an instance of an object.
   at TestCentric.Gui.Controls.TestTree.checkBoxesMenuItem_Click(Object sender, EventArgs e) in C:\src\NUnit\test-centric-gui\src\TestCentric\testcentric.gui\Controls\TestTree.cs:line 696
   at System.Windows.Forms.MenuItem.OnClick(EventArgs e)
   at System.Windows.Forms.MenuItem.MenuItemData.Execute()
   at System.Windows.Forms.Command.Invoke()
   at System.Windows.Forms.Control.WmCommand(Message& m)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.Form.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)
```

mikkelbu:
Fixes #52

Note, I've decided not to update the setting `TestTreeSettings.ShowCheckBoxes` when changing  `Show CheckBoxes` via the menu or via the context menu. This can be done via `Model`, if we choose so. The reason for this, is it seems that most changes in menus do not update the "default" settings in `Tools => Settings` (except for Full GUI vs Mini GUI).

Also note that we do not change the user interface, if we change the setting `TestTreeSettings.ShowCheckBoxes`. This will first have effect when we restart the program. We could do this by commenting in the method `UserSettings_Changed` (and do some more changes), but it does not seem like we update the UI immediately when changing settings in `Tools => Settings`.

So before making the above mentioned changes I think we should decide upon the connection between the settings in `Tools => Settings` and settings we can set directly in the GUI (in menus and context-menues). I think of them as the "default" vs. the "current" settings, but it can be that this is just my interpretation.

### DIFF
--- a/src/TestCentric/testcentric.gui/Controls/TestTree.cs
+++ b/src/TestCentric/testcentric.gui/Controls/TestTree.cs
@@ -114,8 +114,6 @@ namespace TestCentric.Gui.Controls
 
         private ITestModel Model { get; set; }
 
-        private UserSettings UserSettings { get; set; }
-
 		[Browsable(false)]
 		public bool ShowCheckBoxes
 		{
@@ -692,8 +690,8 @@ namespace TestCentric.Gui.Controls
         //}
 
         private void checkBoxesMenuItem_Click(object sender, System.EventArgs e)
-		{
-            UserSettings.TestCentric.TestTree.ShowCheckBoxes = ShowCheckBoxes = !checkBoxesMenuItem.Checked;
+        {
+            ShowCheckBoxes = !checkBoxesMenuItem.Checked;
         }
 
         private void UpdateCategorySelection()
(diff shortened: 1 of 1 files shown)


CharliePoole:
@mikkelbu I agree with both what you have done and what you suggest we need to look into. I have started writing a sort of spec about how this should work and which settings take precedence.

Since this project is a look-alike of the old NUnit Gui, I often run that in order to see what it does in a particular case.

Anyway, merging this and we can change more in a follow-up PR. I'm working on a more general one to do with settings organization right now.

PR #54, opened 2018-06-14
Reorganize Settings

### LINKED ISSUE #50: Review and reorganize Settings for NUnit 3
I did the first cut of organizing settings pages very mechanically. If something was handled by the nunit.util or nunit.core assemblies in V2, I put the page under the Engine branch. But by design a large number of settings that the core previously queried must now be handled by the runner. ReloadOnRun is a good example. 

The kinds of discrepancies I'm concerned about include:
1. "Settings" that may no longer exist in NUnit 3
2. Settings calling for actions that must now be performed by the runner (like above example).
3. Settings that must be passed in the package rather than read automatically by the engine.
4. Settings that are now passed as strings rather than enums.

Rather than fixing all this one item at a time, I'm setting up this issue and assigning it to myself. I'll make a list of all package settings and global settings in use and try to ensure that they are all handled correctly. Depending on how long it takes, this may involve multiple PRs.

CharliePoole:
Fixes #50

### DIFF
--- a/src/TestModel/tests/SettingsTests.cs
+++ b/src/TestModel/tests/SettingsTests.cs
@@ -1,4 +1,27 @@
-﻿using System;
+﻿// ***********************************************************************
+// Copyright (c) 2018 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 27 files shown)


CharliePoole:
@mikkelbu This will again have white space changes. Once this merges I'll do a big change to detabify everything, but this was already in process.

We still have issues about whether various settings changed in the gui should go to the defaults or not. I'm working on writing that up. I found we get settings from five different places! So that will be a separate issue.

src/TestModel/tests/SettingsTests.cs
```diff
@@ -0,0 +1,228 @@
+﻿using System;
```
mikkelbu: Missing copyright notice.

mikkelbu:
@CharliePoole My weekend is pretty booked, so I'll probably first have time to do a proper review in the beginning of next week.

CharliePoole:
@mikkelbu OK, I'll go with @immeraufdemhund 's review. This is an area we will probably revisit a few times before we get it right. 😺

src/TestModel/tests/SettingsTests.cs
CharliePoole: Added

PR #56, opened 2018-06-18
Restore results correctly after reload

### LINKED ISSUE #14: Tree always cleared of results on reload
The action should follow the setting checkbox that indicates whether to clear results.

CharliePoole:
Fixes #14 

It turns out that the reload code was the same as the load code. It got a completely new runner and started from scratch, rather than reusing the runner. The fix that's in causes us to reload the results when called for after reloading the tests. It does not fix any problems in __how__ the reloading works. We will need to create issues for any specific problems in matching old results to the newly reloaded test. Examples of situations we have yet to deal with include deleting or renaming a test.

### DIFF
--- a/src/TestCentric/testcentric.gui/Controls/TestSuiteTreeView.cs
+++ b/src/TestCentric/testcentric.gui/Controls/TestSuiteTreeView.cs
@@ -646,15 +646,14 @@ namespace TestCentric.Gui.Controls
         /// <param name="test">Test suite to be loaded</param>
         public void Reload(TestNode test)
         {
-            ResultNode result = ((TestSuiteTreeNode)Nodes[0]).Result;
             VisualState visualState = new VisualState(this);
 
             Load(test);
 
             visualState.Restore(this);
 
-            if (result != null && !UserSettings.Gui.ClearResultsOnReload)
-                RestoreResults(result);
+            if (!UserSettings.Gui.ClearResultsOnReload)
+                RestoreResults(test);
         }
 
         /// <summary>
@@ -1044,15 +1043,13 @@ namespace TestCentric.Gui.Controls
             }
         }
 
-        private void RestoreResults(ResultNode resultNode)
+        private void RestoreResults(TestNode testNode)
         {
-            foreach (TestNode child in resultNode.Children)
-            {
-                if (child is ResultNode)
-                    RestoreResults(child as ResultNode);
-            }
 
-            SetTestResult(resultNode);
+            foreach (TestNode child in testNode.Children)
+                RestoreResults(child);
+
+            SetTestResult(Model.GetResultForTest(testNode));
         }
 
         private void LoadAlternateImages()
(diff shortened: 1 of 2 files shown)


CharliePoole:
@TestCentric/gui-team This PR is a good example of how I'd like to keep PRs restricted. There are a ton of other things that need to be fixed, all related to reloading tests. I could have fixed some of them as a part of this PR but I held off. GitHub gives us all the issues and PRs we want for free. :-) Let's try to keep them small and focused, aiming at minimizing the overhead for any review.

I called for somebody to review this one, although I might have just merged it myself. Main reason is that I moved the responsibility for understanding what Result applies to a test node from the tree view to the engine.

src/TestCentric/testcentric.gui/Controls/TestSuiteTreeView.cs
```diff
@@ -1044,15 +1043,13 @@ private void RestoreVisualState()
...
         }
 
-        private void RestoreResults(ResultNode resultNode)
+        private void RestoreResults(TestNode testNode)
         {
-            foreach (TestNode child in resultNode.Children)
-            {
-                if (child is ResultNode)
-                    RestoreResults(child as ResultNode);
-            }
```
immeraufdemhund: extra line. sorry didn't catch this before i hit approve

src/TestCentric/testcentric.gui/Controls/TestSuiteTreeView.cs
```diff
@@ -1044,15 +1043,13 @@ private void RestoreVisualState()
...
 
-            SetTestResult(resultNode);
+            foreach (TestNode child in testNode.Children)
```
immeraufdemhund: TestNode could possibly be replaced with var as testNode.Children is a list of TestNodes

immeraufdemhund:
Sorry about that Charlie I ran it and it worked and so I clicked approve before checking all the files first. Minor changes requested.

src/TestCentric/testcentric.gui/Controls/TestSuiteTreeView.cs
CharliePoole: My theory is that the compiler knows what it is but the person reading the code may not.

CharliePoole:
Thanks for reviewing.

src/TestCentric/testcentric.gui/Controls/TestSuiteTreeView.cs
immeraufdemhund: ok. My bad I should have checked first. I think you've told me that before. FYI I did a search with regex in all files for `^\s*foreach \(\s*var` and got 11 hits. I'm not saying to fix it, but there is 1 in this file on line 602.

CharliePoole: We don't have a coding standard for this project yet but if I were writing it, I'd put use of var in the permissive category, subject to judgement based on ease of understanding. It seems to me that NUnit is moving more toward the establishment of fixed rules, with less room for judgement. It's a trend I'm not fond of. :-)

immeraufdemhund: lolz.. and i just pushed up a remove-var branch ;D (seeing as i forgot to hit comment yesterday when I wrote the above)

PR #57, opened 2018-06-18
Move ClearResultsOnReload setting to Assembly Reload settings page

### LINKED ISSUE #26: Clear Results Setting is on wrong page
The setting for Clearing results when reloading is on the tree settings page but affects more than just the tree. It probably belongs on the reload page.

CharliePoole:
Fixes #26


CharliePoole:
That's weird... they aren't lines I changed manually, so it's something to do with opening in VS2015 for the first time (files are from NUnit V2, VS2012).

PR #69, opened 2018-06-20
removing var from all foreach loops

immeraufdemhund:
i forgot that i told vscode to remove all trailing white spaces so sorry the commit is so large.

### DIFF
--- a/src/TestModel/model/TestModel.cs
+++ b/src/TestModel/model/TestModel.cs
@@ -24,6 +24,7 @@
 using System;
 using System.Collections.Generic;
 using NUnit.Engine;
+using NUnit.Engine.Extensibility;
 
 namespace TestCentric.Gui.Model
 {
@@ -44,7 +45,7 @@ namespace TestCentric.Gui.Model
             TestEngine = testEngine;
             Services = new TestServices(testEngine);
 
-            foreach (var node in Services.ExtensionService.GetExtensionNodes(PROJECT_LOADER_EXTENSION_PATH))
+            foreach (IExtensionNode node in Services.ExtensionService.GetExtensionNodes(PROJECT_LOADER_EXTENSION_PATH))
             {
                 if (node.TypeName == NUNIT_PROJECT_LOADER)
                     NUnitProjectSupport = true;
@@ -174,7 +175,7 @@ namespace TestCentric.Gui.Model
 
             _events.FireTestLoaded(Tests);
 
-            foreach (var subPackage in TestPackage.SubPackages)
+            foreach (TestPackage subPackage in TestPackage.SubPackages)
                 Services.RecentFiles.SetMostRecent(subPackage.FullName);
         }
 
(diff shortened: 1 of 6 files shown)


src/TestModel/model/TestModel.cs
```diff
@@ -44,7 +45,7 @@ public TestModel(ITestEngine testEngine)
             TestEngine = testEngine;
             Services = new TestServices(testEngine);
 
-            foreach (var node in Services.ExtensionService.GetExtensionNodes(PROJECT_LOADER_EXTENSION_PATH))
+            foreach (IExtensionNode node in Services.ExtensionService.GetExtensionNodes(PROJECT_LOADER_EXTENSION_PATH))
```
CharliePoole: var seems clear to me here because the name of the method indicates what we are getting.

src/TestModel/model/TestModel.cs
```diff
@@ -174,7 +175,7 @@ public void LoadTests(IList<string> files)
 
             _events.FireTestLoaded(Tests);
 
-            foreach (var subPackage in TestPackage.SubPackages)
+            foreach (TestPackage subPackage in TestPackage.SubPackages)
```
CharliePoole: ditto

src/TestModel/model/TestModel.cs
```diff
@@ -344,7 +345,7 @@ public TestPackage MakeTestPackage(IList<string> testFiles)
             // We use shadow copy so that the user may re-compile while the gui is running.
             package.AddSetting(EnginePackageSettings.ShadowCopyFiles, true);
 
-            foreach (var entry in PackageSettings)
+            foreach (KeyValuePair<string, object> entry in PackageSettings)
```
CharliePoole: I'd be OK with var here as well, since it's so much shorter and (to me) it's obvious we are dealing with a dictionary because of the following line. Spurious logic? I don't know.

src/TestModel/model/TestModel.cs
```diff
@@ -372,7 +373,7 @@ private List<IRuntimeFramework> GetAvailableRuntimes()
 
                 // Skip duplicates
                 var duplicate = false;
-                foreach (var rt in runtimes)
+                foreach (IRuntimeFramework rt in runtimes)
```
CharliePoole: In both this one and the one above it, the name of the collection tells me it's some kind of runtime. It's not clear in the second that we are getting an interface, while it is in the first one. I'm happy with var here too. 😕

src/TestModel/tests/AvailableRuntimesTest.cs
```diff
@@ -54,7 +54,7 @@ public void DisposeTestModel()
         public void AtLeastOneRuntimeIsAvailable()
         {
             Assert.That(_availableRuntimes.Count, Is.GreaterThan(0));
-            foreach (var runtime in _availableRuntimes)
+            foreach (IRuntimeFramework runtime in _availableRuntimes)
```
CharliePoole: ditto

immeraufdemhund:
ok, so this pull request was before you mentioned user judgment. And since you feel (and i do too) that var is acceptable I think it is ok to close this pull request

CharliePoole:
I'm not a big fan of changes "for purposes of consistency" if that's your intention. OTOH, I'm afraid that our earlier discussion may have led you to do this.

As you look at each individual change, do you feel that they were unclear before?

[Ah! I see you just answered that!!!]

In that case, let's close it and put count this as a learning experience. You have convinced me we need some doc that says - if nothing else - leave this stuff up to the user unless it causes a real problem!

PR #72, opened 2018-06-26
Initial implementation of doubles to replace tests using the engine

CharliePoole:
This is really only a start, with some of the implementations still throwing NYI on unused methods. I think we should commit it and start to enhance it as needed for other tests.

Where I revised some tests, they ran much faster than when the actual engine was used - obviously - and were also much easier to set up than tests that use NSubstitute to mock the engine.

I'll leave this for a day before merging in hopes that someone gets a chance to look it over.


CharliePoole:
@mikkelbu To match the directory it's in. The test utilities project is still named after NUnit because that's where it came from but could probably be changed.

PR #74, opened 2018-06-26
Allow opening multiple files

### LINKED ISSUE #71: Multiple files on command-line or in file open dialog not opened
Only the first file specified on the gui command-line is opened

CharliePoole:
Fixes #71


CharliePoole:
@TestCentric/gui-team Here's an example of a very small diff. Something didn't work. I made very few changes. Now it works. I merged it.

Feel free to do the same as committers. Feel free also to ask for review when you think other minds should consider the change, particularly if it changes how users use the GUI.

PR #78, opened 2018-07-11
Add root node to display as parent to all assemblies

### LINKED ISSUE #73: Multiple loaded files have no root in tree
When multiple test assemblies are opened, they are displayed in the tree as separate roots. That is, there is no common root as in V2. Consequently, to run all the tests using the Run button, the user must enable checkboxes and select each root.

Since this version of the gui is intended to be similar to V2, the easy solution would be to provide an artificial root in the tree, which is selected by default. That's my preference.

One alternative would be to provide a separate "Run All" button.

What do folks think?

CharliePoole:
Fixes #73

### DIFF
--- a/src/TestCentric/testcentric.gui/Controls/TestSuiteTreeView.cs
+++ b/src/TestCentric/testcentric.gui/Controls/TestSuiteTreeView.cs
@@ -599,8 +599,6 @@ namespace TestCentric.Gui.Controls
 
                 try
                 {
-                    //foreach (var child in topLevelNode.Children)
-                    //    AddTreeNodes(Nodes, child, false);
                     AddTreeNodes(Nodes, topLevelNode, false);
 
                     SetInitialExpansion();
@@ -1240,33 +1238,6 @@ namespace TestCentric.Gui.Controls
         {
             foreach (TestSuiteTreeNode node in nodes)
                 FindCheckedNodes(node, topLevel);
-        }
-    }
-
-    #endregion
-
-    #region TestGroup
-
-    /// <summary>
-    /// A TestGroup is essentially a TestSelection with a
-    /// name and image index for use in the tree display.
-    /// Its TreeNode property is externally set and updated.
-    /// It can create a filter for running all the tests
-    /// in the group.
-    /// </summary>
-    public class TestNodeCollection : List<TestNode>, ITestItem
-    {
-        public TestNodeCollection(IEnumerable<TestNode> tests) : base(tests) { }
-
-        public string Name { get { return GetType().Name; } }
-
-        public TestFilter GetTestFilter()
-        {
-            // HACK
-            if (Count == 1 && this[0].FullName == "TestRun")
-                return TestFilter.Empty;
-
-            return Filters.MakeIdFilter(this.ToArray());
         }
     }
 
--- a/src/TestCentric/testcentric.gui/Controls/TestSuiteTreeView.cs
+++ b/src/TestCentric/testcentric.gui/Controls/TestSuiteTreeView.cs
@@ -599,8 +599,9 @@ namespace TestCentric.Gui.Controls
 
                 try
                 {
-                    foreach (var child in topLevelNode.Children)
-                        AddTreeNodes(Nodes, child, false);
+                    //foreach (var child in topLevelNode.Children)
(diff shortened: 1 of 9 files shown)


src/TestCentric/testcentric.gui/Controls/TestSuiteTreeView.cs
```diff
@@ -599,8 +599,9 @@ public void Load(TestNode topLevelNode)
 
                 try
                 {
-                    foreach (var child in topLevelNode.Children)
-                        AddTreeNodes(Nodes, child, false);
+                    //foreach (var child in topLevelNode.Children)
```
immeraufdemhund: remove commented out code

src/TestCentric/testcentric.gui/TestCentricPresenter.cs
```diff
@@ -391,8 +391,7 @@ public void RunFailedTests()
 
         public void RunTests(TestNode test)
         {
-            EnableRunCommand(false);
-            Model.RunTests(test);
+            RunTests(new TestNode[] { test });
```
immeraufdemhund: nice :) I like seeing simplified code

src/TestModel/model/TestNode.cs
```diff
@@ -54,13 +54,40 @@ public class TestNode : ITestItem
...
+                    IsSuite = false;
+                    Name = Xml.GetAttribute("name");
+                    FullName = Xml.GetAttribute("fullname") ?? Name;
+                    Type = "TestCase";
+                    TestCount = 1;
+                    RunState = GetRunState();
+                    break;
+            }
+
+            //IsSuite = Xml.Name == "test-suite" || Xml.Name == "test-run";
```
immeraufdemhund: remove commented out code

src/TestModel/model/TestNode.cs
```diff
@@ -54,13 +54,40 @@ public class TestNode : ITestItem
...
             Xml = xmlNode;
-            IsSuite = Xml.Name == "test-suite" || Xml.Name == "test-run";
             Id = Xml.GetAttribute("id");
-            Name = Xml.GetAttribute("name");
-            FullName = Xml.GetAttribute("fullname");
-            Type = IsSuite ? GetAttribute("type") : "TestCase";
-            TestCount = IsSuite ? GetAttribute("testcasecount", 0) : 1;
-            RunState = GetRunState();
+
+            switch (Xml.Name)
```
immeraufdemhund: took me a second to realize this is much easier to read. Nice

immeraufdemhund: i meant the entire switch. made it very clera what was to be shown.

CharliePoole:
I'm also going to fix the item I noted as `HACK` before merge.

src/TestCentric/testcentric.gui/Controls/TestSuiteTreeView.cs
```diff
@@ -1261,6 +1262,10 @@ public class TestNodeCollection : List<TestNode>, ITestItem
 
         public TestFilter GetTestFilter()
         {
+            // HACK
+            if (Count == 1 && this[0].FullName == "TestRun")
+                return TestFilter.Empty;
+
```
CharliePoole: I really don't like this and I just noticed what may be the root cause of the problem. The TestNodeCollection class is defined in the view, kused by the presenter and really belongs in the model!

CharliePoole:
I removed the comments you noted and also restructured. TestNodeCollection is only used in the model, so I moved it there. I took out the HACK and made the test node itself responsible (if it's a test-run node) to return an empty filter when asked for one.

I also fixed up some incorrect comments and removed code for the somewhat related TestSelection class from the solution, but kept in in the repo. Unused code that called it is compiled out using the NEW_GUI define. It would be cleaner to just delete the code, but I want to figure out how to merge the models for the old and new guis before I do that, so NEW_GUI will at least temporarily hide code in the model that is only there for the new gui.

I'd really like to be done with this and work on the new GUI!!!

immeraufdemhund:
sounds nice... I don't see that it got pushed up yet though.

CharliePoole:
Odd. I'll check.

Issue #95, opened 2018-08-13
Implement MVP for main form and sub-views


CharliePoole:
Issue #93 shows just how fragile it our GUI design is in its current state. The GUI started out as a simple Model-View design, which was standard for Windows apps many years ago, and started evolving slightly toward MVP under NUnit V2. The evolution was never completed and the TestCentric GUI inherited the mixed state.

This causes problems because view elements (forms and controls) have to be given a reference to the model in a second-stage initialization call. Of course, any reference that may be made to the model before this initialization takes place will result in a null reference exception.

The experimental gui uses a purer MVP design for the main form. We should port that approach to this project. The design would work as follows:

1. The main form and each significant portion of it would function as a view, for example, main view, status bar view, tree view, etc.

2. Each view would have an associated presenter.

3. The entry code would initialize all views and presenters before calling Application.Run on the main form. You can see how this is done in the experimental gui.

4. Presenters would have a reference to the model and the view that they control. They would respond to events from each of them and make calls into each of them. They would be tested by mocking the at least the view and possibly also the model.

5. The model and view would have no knowledge of one another or of the presenters.

6. We would __try__ to keep all presenters independent - it remains to be seen if this is possible without injecting too much presentation info into the model. I think this depends on how we partition the views. Right now, there's a bit of mutual interdependence between the tree view and the "TestTree" panel but I think this can be fixed.

7. The experimental gui uses an intermediate layer of "view elements" to abstract controls and give all views a consistent non-Windows interface. Initially, I'd try to do without this layer since it adds complexity, but it's available if we need it.

Implementation of this would involve multiple PRs, dealing with one view at a time, and a final PR to clean up any remnants of the old design in the code.

Note: I only intend this to deal with the main form. Some dialogs deal directly with the model and may as well stay that way, at least for now.

CharliePoole:
This is a (tentative) partitioning of the issue into individual view / presenter pairs that need to be created. It will be updated as new information is discoverd and as the implementation of each one is complete.

 - [ ] Main form (including main menu and any controls not listed separately)

 - [ ] Status bar

 - [ ] Progress bar

 - [ ] Tree display

 - [ ] Category selection

 - [ ] Errors and failures

 - [ ] Tests not run

 - [ ] Text output

CharliePoole:
[...]

PR #96, opened 2018-08-16
Use MVP pattern for StatusBar and ProgressBar

CharliePoole:
Here's a start on issue #95 which will take multiple PRs before it closes.

I picked StatusBar and ProgressBar to start with because they are simple cases but still illustrate the approach fairly completely. I'd appreciate reviews on this PR that cover the general approach and how it's implemented as well as specific code issues.

Note that these two views are output only. Views with buttons and menus will require additions.

I took the general approach from the experimental gui. Each subview has a presenter which is created and then does it's job independently of the others. The experimental gui had an extra layer of "view elements" that served to further isolate the view from the presenter. That has seemed to be more trouble than it's worth. So I didn't repeat it here.

Note that we can't get NREs any longer due to the view or presenter not being initialized. Both of them are fully initialized in their constructors. The presenter gets both the view and the model on construction. Of course, other errors are still possible. 😄


CharliePoole:
@mikkelbu @immeraufdemhund Are either of you able to review this? I'll do some subsequent changes without review but I thought that the structure needed to be looked at by at least one of you before this was merged. I'm now working on further views, chained off this code, but I don't want to go too far until then.

CharliePoole:
Nobody has reviewed this, but I'm going ahead and merging since it's blocking. Any comments on the overall design would still be helpful.

immeraufdemhund:
sorry about that, i didn't do much on the computer this weekend.

PR #102, opened 2018-08-26
Implement MVP for ErrorsAndFailuresView

CharliePoole:
Part of issue #95. This PR deals with the ErrorsAndFailuresView.

In earlier PRs, I used a presenter-first approach, developing the interface and the presenter in parallel and only creating the view itself after I had a fully working presenter with all its tests passing.

This one is more complicated, for a few reasons:
 * The view originates events while others only responded to model events
 * The view contains three different relatively complex controls
 * This is the last of the three tab views so the ResultTabs control can be simultaneously simplified

Consequently, I'm doing it bit by bit with some functionality remaining in the view control itself while other things are moved out to the presenter. IOW the architecture is hybrid for the time being.

@mikkelbu @immeraufdemhund I'd like to develop this PR "under review" with each commit getting a review, provided one or both of you has the time.


CharliePoole:
In this first commit, I have set up the basic MVP structure and implemented handling of all the model events in the presenter. I copied the original ErrorDisplay control to the Views directory and renamed it as ErrorsAndFailuresView. It no longer handles model events directly but still continues to be initialized through the IViewControl interface and does a number of things that will eventually be moved to the presenter.

As a side effect, I simplified the ResultTabs control, since it no longer needs to clear it's three child controls.

CharliePoole:
Second commit removes remaining response to model events from the view and moves it to the presenter. Many more tests. Refactoring of test classes.

Also found an unrelated problem in TextOutputView and fixed it.

The view is still initialized with the model, which it uses to save a reference to UserSettings. That will be the next thing to remove so the view no longer has any ref to the model.

CharliePoole:
Third commit removes remaining references to the model from the view. View is no longer an `IControlView` an all initialization happens on construction of the presenter.

The `TipWindow` feature, which displayed the full result message on hovering over an item, is still disabled. I'll try to get that working and do some cleanup in the final commit, to come.

src/TestCentric/tests/Presenters/ErrorsAndFailuresPresenterTests.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            _settings.Gui.ClearResultsOnReload = false;
+
+            FireTestReloadedEvent(FAKE_TEST_RUN);
+
+            _view.DidNotReceive().Clear();
+        }
+
+        [Test]
+        public void WhenTestIsReloaded_IfClearResultsIfTrue_DisplayIsCleared()
```
mikkelbu: Very minor nit pick `IfClearResultsIfTrue` => `IfClearResultsIsTrue`.

src/TestCentric/tests/Views/ErrorsAndFailuresViewTests.cs
```diff
@@ -26,17 +26,15 @@
 using NUnit.Framework;
 using NUnit.TestUtilities;
 
-namespace TestCentric.Gui.Tests
+namespace TestCentric.Gui.Views
```
mikkelbu: Do we want this namespace for tests? (in general it seems that the namespaces within the `tests` folder are somewhat inconsistent.

src/TestCentric/testcentric.gui/Presenters/ErrorsAndFailuresPresenter.cs
```diff
@@ -67,6 +67,32 @@ private void WireUpEvents()
...
+            {
+                if (e.Result.Status == TestStatus.Failed || e.Result.Status == TestStatus.Warning)
+                    if (e.Result.Site != FailureSite.Parent)
+                        AddResult(e.Result);
+            };
+
+            _model.Events.SuiteFinished += (TestResultEventArgs e) =>
+            {
+                if (e.Result.Status == TestStatus.Failed || e.Result.Status == TestStatus.Warning)
+                    if (e.Result.Site != FailureSite.Parent && e.Result.Site != FailureSite.Child)
```
mikkelbu: Any reason for why you choose `if (e.Result.Site != FailureSite.Parent && e.Result.Site != FailureSite.Child)` instead of `  if (e.Result.Site == FailureSite.SetUp || e.Result.Site == FailureSite.Test || e.Result.Site == FailureSite.TearDown)` (as it was in ErrorsAndFailuresView.cs). I know that they currently are equivalent, but if another value is added to the enum then it will also affect the if-statement.

src/TestCentric/testcentric.gui/Views/ErrorsAndFailuresView.cs
```diff
@@ -424,19 +435,19 @@ public void InitializeView(ITestModel model, TestCentricPresenter presenter)
...
-                if (e.Result.Status == TestStatus.Failed || e.Result.Status == TestStatus.Warning)
-                    if (e.Result.Site != FailureSite.Parent)
-                        InsertTestResultItem(e.Result);
-            };
+            //model.Events.TestFinished += (TestResultEventArgs e) =>
+            //{
+            //    if (e.Result.Status == TestStatus.Failed || e.Result.Status == TestStatus.Warning)
+            //        if (e.Result.Site != FailureSite.Parent)
+            //            InsertTestResultItem(e.Result);
+            //};
```
mikkelbu: Shouldn't this just be deleted?

src/TestCentric/tests/Presenters/ErrorsAndFailuresPresenterTests.cs
CharliePoole: Fixed for my next commit.

src/TestCentric/tests/Views/ErrorsAndFailuresViewTests.cs
CharliePoole: I'm gradually doing away with the `Tests` namespace, which is why it's inconsistent. I've come to feel that it's redundant, since all the test classes have different names from the product classes and most of them include the word "Tests".

src/TestCentric/testcentric.gui/Presenters/ErrorsAndFailuresPresenter.cs
CharliePoole: @mikkelbu Thanks! I'll watch for further reviews.

CharliePoole: We were doing the negative compare in one place and the positive in another. I wanted to be consistent and to use the code that's the easiest to understand. In my mind the negative means "If the failure was caused by the parent setup or if a suite failed because of a child failure, then don't show it" which is exactly how I would specify it in English. Of course, that's a subjective judgement.

Regarding changes, that would mean a pretty basic change to the framework, so it seems unlikely, but if it did happen, I'd prefer the error that shows me too much rather than the one that shows nothing and goes unnoticed.

CharliePoole:
I'm particularly happy that there are 50 tests where we had none before - 199 looking at all of the presenters done so far.

CharliePoole:
Final commit enables use of the TipWindow, displaying the full text of the message when it doesn't fit and you hover over it. There is currently no way to change the setting but I think that's for another issue. We would need to decide whether to change it in the context menu or in a settings dialog.

CharliePoole:
Letting this one sit for a bit in case anyone wants to review further. I'll update status on #95.

CharliePoole:
Going ahead with the merge, but feel free to comment further as this is an ongoing issue with more PRs to come.

PR #108, opened 2018-09-21
Present AboutBox without throwing NRE

mikkelbu:
`Site` was null, so the code failed with a NRE.


mikkelbu:
@CharliePoole I have no problem with merging, but I was waiting on AppVeyor to complete and then I got side-tracked in other issues 😄

CharliePoole:
Weird. I also tried to merge it, but I must have left it hanging when I shut down the computer. Yes, when I saw the short time frame I figured that was what happened, but I did also want to be sure that you're OK with the approach of merging when ready because it's so different from NUnit's

PR #120, opened 2018-10-21
TestRun node won't be loaded if there is only 1 child

### LINKED ISSUE #100: Don't display "Test-Run" as root node in tree with only one assembly
We should try to work as the V2 gui did: A root should be added only when there are multiple assemblies.

drSte:
Fixes #100


mikkelbu:
Hi @drSte 
I've edited your initial message such that it will auto-close the issue when it is merged to master. I'll take a look at the PR now.

mikkelbu:
Hi @drSte 
I don't think that 52a0253 should be part of this PR. You should rather use separate branches for each PR, so that you don't get changes mixed together.

drSte:
Hi @mikkelbu ,

You are totally right, I made a mistake. It is enough to rebase in a separate branch?
BR.

drSte:
Hi @mikkelbu,

I created a new pull request with a separate branch. if that one is ok you can close this one

PR #121, opened 2018-10-22
tiny misspell

zBrianW:
While i didn't have chance to look at the logic for a fix, i did notice this tiny misspell.


mikkelbu:
Thanks for correcting the typo @zBrianW 👍

PR #122, opened 2018-10-22
TestRun node won't be loaded if there is only 1 child

### LINKED ISSUE #100: Don't display "Test-Run" as root node in tree with only one assembly

drSte:
created new PR from a different branch. Fixes #100


mikkelbu:
Thanks for the contribution @drSte 👍

Issue #18, opened 2018-03-11
Reload on Change not working


CharliePoole:
We should set up file watchers and reload the tests if needed. Ideally, only reload the assembly that changed if multiple assemblies are open.

CharliePoole:
__Hacktoberfest contributors...

This is a moderately complex issue. When Reload on Change is set, we should be creating a file watcher for each assembly opened. When a change is detected we should reload the test files.

There is existing code from NUnit V2 that relates to this but it needs to be carefully vetted and updated to work with NUnit 3. There's enough work here that you may want to consider doing it in several PRs.

drSte:
i will try to fix this

drSte:
@CharliePoole it is ok if i use a FileSystemWatcher object?

CharliePoole:
You will want to look at how NUnit V2 does it, since we are basically emulating the V2 GUI for TestCentric 1.0.Basically, there are two classes that watch files for changes:

`FileWatcher` is used to watch for changes in a single project file. I don't think we need this right away since our support of project files is limited right now.

`AssemblyWatcher` looks for changes to the name of any assemblies being tested. It uses a collection of `FileSystemWatchers` to do it. That's the one you need - or something equivalent.

I think I'd do this in two stages....

1. Get AssemblyWatcher ported and working. It has tests, so that will be of some help. I think it belongs in the model.

2. Possibly as a separate PR, get it hooked up and working. If the user wants reload on change, then we have to set up the watchers and when they are triggered, we do our job. You'll find some commented out code where it was used in the original code I ported.

Good luck! Be sure to ask questions if you need to.

CharliePoole:
@drSte Our comments crossed. Take a look at what AssemblyWatcher does. I think you will find that it's a lot more complicated than just a FileSystemWatcher for two reasons...

1. It has to watch a bunch of assemblies, whcih may be in various locations.

2. The assemblies change as you compile, first one, then another, etc. So we have to set up timers in order to avoid reloading before all the changes are complete.

CharliePoole:
BTW, right now, we reload everythng when anything changes. A future enhancement might be to only reload affected assemblies.

drSte:
@CharliePoole ,I am ready with the implementation, Will push it as soon as the previous PR will be accepted

PR #131, opened 2018-10-25
port of AssemblyWatcher from NUnit2

drSte:
related to #18

### DIFF
--- a/src/TestModel/model/AssemblyWatcher.cs
+++ b/src/TestModel/model/AssemblyWatcher.cs
@@ -1,5 +1,5 @@
 ﻿// ***********************************************************************
-// Copyright (c) 2018 Charlie Poole
+// Copyright (c) 2002-2018 Charlie Poole
 //
 // Permission is hereby granted, free of charge, to any person obtaining
 // a copy of this software and associated documentation files (the
@@ -28,7 +28,7 @@ using System.Timers;
 
 namespace TestCentric.Gui.Model
 {
-    public class AssemblyWatcher:IAsemblyWatcher,IDisposable
+    public class AssemblyWatcher:IAsemblyWatcher
     {
         private FileSystemWatcher[] fileWatchers;
         private FileInfo[] files;
--- a/src/TestModel/model/AssemblyWatcher.cs
+++ b/src/TestModel/model/AssemblyWatcher.cs
@@ -1,10 +1,34 @@
-﻿using System;
+﻿// ***********************************************************************
+// Copyright (c) 2018 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+using System.Collections;
 using System.IO;
 using System.Timers;
 
 namespace TestCentric.Gui.Model
 {
-    public class AssemblyWatcher:IAsemblyWatcher
+    public class AssemblyWatcher:IAsemblyWatcher,IDisposable
     {
         private FileSystemWatcher[] fileWatchers;
         private FileInfo[] files;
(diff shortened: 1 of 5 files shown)


src/TestModel/model/AssemblyWatcher.cs
```diff
@@ -0,0 +1,136 @@
+﻿using System;
```
CharliePoole: Each file should have the standard header. Use date range 20xx-2018, where 20xx is whatever the date is for V2.

src/TestModel/model/AssemblyWatcher.cs
```diff
@@ -0,0 +1,136 @@
...
+{
+    public class AssemblyWatcher:IAsemblyWatcher
+    {
+        private FileSystemWatcher[] fileWatchers;
+        private FileInfo[] files;
+
+        protected System.Timers.Timer timer;
+        protected string changedAssemblyPath;
+
+        protected FileInfo GetFileInfo(int index)
```
CharliePoole: These protected items can stay for now, but we should review later after #18 is complete. They seem to be used only for a "mock" we create for testing and we may want to use NSubstitute instead. Leave for now.

src/TestModel/model/IAsemblyWatcher.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <summary>
+        /// Initializes the watcher with assemblies to observe for changes.
+        /// </summary>
+        /// <param name="delayInMs">The delay in ms.</param>
+        /// <param name="assemblies">The assemblies.</param>
+#if CLR_2_0 || CLR_4_0
+        void Setup(int delayInMs, System.Collections.Generic.IList<string> assemblies);
+#else
+        void Setup(int delayInMs, System.Collections.IList assemblies);
+#endif
```
CharliePoole: We don't use these defines in the GUI, so we should just keep the generic and add using statement at top.

src/TestModel/model/IAsemblyWatcher.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// Initializes the watcher with assemblies to observe for changes.
+        /// </summary>
+        /// <param name="delayInMs">The delay in ms.</param>
+        /// <param name="assemblyFileName">Name of the assembly file.</param>
+        void Setup(int delayInMs, string assemblyFileName);
+
+        /// <summary>
+        /// Releases all resources held by the watcher.
+        /// </summary>
+        void FreeResources();
```
CharliePoole: It probably makes sense to drop this specially named method and have the interface inherit from `IDisposable`.

src/TestModel/tests/AssemblyWatcherTests.cs
```diff
@@ -0,0 +1,129 @@
+﻿using System;
+using System.IO;
+using NUnit.Framework;
+
+
+namespace TestCentric.Gui.Model
+{
+    [TestFixture]
+    [Platform(Exclude = "Win95,Win98,WinMe")]
```
CharliePoole: The exclusion of ancient platforms is no longer relevant. :smile:

src/TestModel/tests/AssemblyWatcherTests.cs
```diff
@@ -0,0 +1,129 @@
...
+            FileInfo fileInfo = new FileInfo(fileName);
+            fileInfo.Delete();
+
+            FileInfo temp = new FileInfo(tempFileName);
+            if (temp.Exists) temp.Delete();
+        }
+
+        [Test]
+        // TODO: Exclusion should really only apply to Mono on Windows
+        [Platform(Exclude = "Mono")]
```
CharliePoole: After we merge, I'll test on Linux to see if any of these individual test exclusions can be removed. (I assume you are working on  Windows @drSte)

src/TestModel/model/AssemblyWatcher.cs
drSte: I forgot about copyright :)

src/TestModel/model/IAsemblyWatcher.cs
drSte: removed

src/TestModel/model/IAsemblyWatcher.cs
drSte: done

src/TestModel/tests/AssemblyWatcherTests.cs
drSte: :) done

src/TestModel/tests/AssemblyWatcherTests.cs
drSte: Yes, you are right, I am using VS on Windows. It is also my main use case for the moment ;)

src/TestModel/model/AssemblyWatcher.cs
```diff
@@ -0,0 +1,156 @@
...
+// ***********************************************************************
+
+using System;
+using System.Collections;
+using System.IO;
+using System.Timers;
+
+namespace TestCentric.Gui.Model
+{
+    public class AssemblyWatcher:IAsemblyWatcher,IDisposable
```
CharliePoole: Let's put `IDisposable` on `IAssemblyWatcher` instead. That implies any `IAssemblyWatcher` has to also be `IDisposable`. Doing it this way says that __this__ `IAssemblyWatcher` is disposable, but not necessarily others.

src/TestModel/model/AssemblyWatcher.cs
```diff
@@ -0,0 +1,156 @@
+﻿// ***********************************************************************
+// Copyright (c) 2018 Charlie Poole
```
CharliePoole: 2002-2018

src/TestModel/model/IAsemblyWatcher.cs
```diff
@@ -0,0 +1,68 @@
+﻿// ***********************************************************************
+// Copyright (c) 2018 Charlie Poole
```
CharliePoole: 2010-2018

src/TestModel/model/IAsemblyWatcher.cs
```diff
@@ -0,0 +1,68 @@
...
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System.Collections;
+
+namespace TestCentric.Gui.Model
+{
+    public delegate void AssemblyChangedHandler(string fullPath);
+    public interface IAsemblyWatcher
```
CharliePoole: IAssemblyWatcher : IDisposable

src/TestModel/tests/AssemblyWatcherTests.cs
```diff
@@ -0,0 +1,128 @@
+﻿using System;
```
CharliePoole: Copyright header, 2011-2018

PR #142, opened 2018-11-01
Move visualstate to presenter

CharliePoole:
This moves the saving and restoring of visual state out of the tree view and into the presenter. The presenter now maintains the lookup dictionary from test ids to tree nodes.


CharliePoole:
@mikkelbu @drSte This merge and the ones preceding it make substantial changes to __where__ certain code is located, moving it from the view to the presenter. This is a very difficult kind of change to deal with if conflicts arise so you should update your local masters.

mikkelbu:
@CharliePoole Thanks for the head up. When there are no pending PRs then I think I'll redo my PR and push it (also now I have found a better diff'er, since GitHub cannot handle "ignore whitespace" when the PR is very large).

PR #150, opened 2018-11-11
Update README.md

mikkelbu:
Update version number of the NUnit test engine.


CharliePoole:
I'll rebase so this gets into the release.

mikkelbu:
It is only for the README.md, so it will not end up in the binaries, so only do it, if it is not too much work.

PR #157, opened 2018-11-14
Skip non-test assemblies without error when running a solution

### LINKED ISSUE #154: When opening solution file non-test assemblies cause an error message
In this situation, we should not give an error but simply ignore the file. A solution will normally contain both test and non-test assemblies.

CharliePoole:
Fixes #154 

In the NUnit project, we have talked about having the packages returned by the VS Project Loader include the SkipNonTestAssemblies setting but it hasn't been implemented. In any case, it would be needed for older versions anyway.

This change adds the setting to any top-level solution files to avoid the error. It does require that the GUI know the extension for solution files, but that's a minor issue.

This could be taken a step further by removing the non-test assemblies from the tree but I wanted to get some comments before going that far. It might confuse people if their assemblies just disappear.

### DIFF
--- a/src/TestModel/model/TestModel.cs
+++ b/src/TestModel/model/TestModel.cs
@@ -23,6 +23,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Text;
 using NUnit.Engine;
 
@@ -365,11 +366,16 @@ namespace TestCentric.Gui.Model
 
             if (engineSettings.SetPrincipalPolicy)
                 package.AddSetting(EnginePackageSettings.PrincipalPolicy, engineSettings.PrincipalPolicy);
-
+            
             //if (Options.InternalTraceLevel != null)
             //    package.AddSetting(EnginePackageSettings.InternalTraceLevel, Options.InternalTraceLevel);
 
             package.AddSetting(EnginePackageSettings.ShadowCopyFiles, engineSettings.ShadowCopyFiles);
+
+            //if (VisualStudioSupport)
+                foreach (var subpackage in package.SubPackages)
+                    if (Path.GetExtension(subpackage.Name) == ".sln")
+                        subpackage.AddSetting(EnginePackageSettings.SkipNonTestAssemblies, true);
 
             foreach (var entry in PackageSettings)
                 package.AddSetting(entry.Key, entry.Value);
--- a/src/TestModel/model/TestModel.cs
+++ b/src/TestModel/model/TestModel.cs
@@ -366,16 +366,15 @@ namespace TestCentric.Gui.Model
 
             if (engineSettings.SetPrincipalPolicy)
                 package.AddSetting(EnginePackageSettings.PrincipalPolicy, engineSettings.PrincipalPolicy);
-            
+
             //if (Options.InternalTraceLevel != null)
             //    package.AddSetting(EnginePackageSettings.InternalTraceLevel, Options.InternalTraceLevel);
 
             package.AddSetting(EnginePackageSettings.ShadowCopyFiles, engineSettings.ShadowCopyFiles);
 
-            //if (VisualStudioSupport)
-                foreach (var subpackage in package.SubPackages)
-                    if (Path.GetExtension(subpackage.Name) == ".sln")
-                        subpackage.AddSetting(EnginePackageSettings.SkipNonTestAssemblies, true);
+            foreach (var subpackage in package.SubPackages)
+                if (Path.GetExtension(subpackage.Name) == ".sln")
+                    subpackage.AddSetting(EnginePackageSettings.SkipNonTestAssemblies, true);
 
             foreach (var entry in PackageSettings)
                 package.AddSetting(entry.Key, entry.Value);
--- a/src/TestModel/tests/TestModelPackageTests.cs
+++ b/src/TestModel/tests/TestModelPackageTests.cs
@@ -68,5 +68,25 @@ namespace TestCentric.Gui.Model
         {
             Console.WriteLine($"Current AppDomain has ShadowCopyFiles = {AppDomain.CurrentDomain.SetupInformation.ShadowCopyFiles}");
         }
+
+        [TestCase("my.dll")]
+        [TestCase("my.sln")]
+        [TestCase("my.dll,my.sln")]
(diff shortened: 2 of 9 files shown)


src/TestModel/model/TestModel.cs
```diff
@@ -365,12 +366,17 @@ public TestPackage MakeTestPackage(IList<string> testFiles)
 
             if (engineSettings.SetPrincipalPolicy)
                 package.AddSetting(EnginePackageSettings.PrincipalPolicy, engineSettings.PrincipalPolicy);
-
+
```
mikkelbu: Spaces

src/TestModel/model/TestModel.cs
```diff
@@ -365,12 +366,17 @@ public TestPackage MakeTestPackage(IList<string> testFiles)
...
+            
             //if (Options.InternalTraceLevel != null)
             //    package.AddSetting(EnginePackageSettings.InternalTraceLevel, Options.InternalTraceLevel);
 
             package.AddSetting(EnginePackageSettings.ShadowCopyFiles, engineSettings.ShadowCopyFiles);
 
+            //if (VisualStudioSupport)
+                foreach (var subpackage in package.SubPackages)
+                    if (Path.GetExtension(subpackage.Name) == ".sln")
+                        subpackage.AddSetting(EnginePackageSettings.SkipNonTestAssemblies, true);
```
mikkelbu: Indented too much (since the `if` is commented out).

src/TestModel/tests/TestModelPackageTests.cs
```diff
@@ -68,5 +68,25 @@ public void DisplayShadowCopySettings()
...
+        [TestCase("my.sln")]
+        [TestCase("my.dll,my.sln")]
+        [TestCase("my.sln,my.dll")]
+        [TestCase("my.sln,another.sln")]
+        public void PackageForSolutionFileHasSkipNonTestAssemblies(string files)
+        {
+            TestPackage package = _model.MakeTestPackage(files.Split(','));
+            string skipKey = EnginePackageSettings.SkipNonTestAssemblies;
+
+            foreach (var subpackage in package.SubPackages)
```
mikkelbu: Personally, I would prefer to have brackets around the body of the `foreach` to make the structure more explicit.

src/tests/mock-assembly/MockAssembly.cs
```diff
@@ -24,6 +24,8 @@
 using System;
 using NUnit.Framework;
 
+[assembly: NonTestAssembly]
```
mikkelbu: Is this intentional? (since the project contain tests)

src/tests/test-utilities/FormTester.cs
```diff
@@ -26,6 +26,8 @@
 using System.Windows.Forms;
 using NUnit.Framework;
 
+[assembly: NUnit.Framework.NonTestAssembly]
```
immeraufdemhund: why would we have this here in a test instead of the assembly information? Also 
```suggestion
[assembly: NonTestAssembly]
```
interesting feature

src/TestModel/model/TestModel.cs
immeraufdemhund:
```suggestion

```
extra white space

src/TestModel/model/TestModel.cs
```diff
@@ -365,12 +366,17 @@ public TestPackage MakeTestPackage(IList<string> testFiles)
...
             if (engineSettings.SetPrincipalPolicy)
                 package.AddSetting(EnginePackageSettings.PrincipalPolicy, engineSettings.PrincipalPolicy);
-
+            
             //if (Options.InternalTraceLevel != null)
             //    package.AddSetting(EnginePackageSettings.InternalTraceLevel, Options.InternalTraceLevel);
 
             package.AddSetting(EnginePackageSettings.ShadowCopyFiles, engineSettings.ShadowCopyFiles);
 
+            //if (VisualStudioSupport)
```
immeraufdemhund: where did the visual studio support comment come from? I don't see it in the presenter above

src/TestModel/model/TestModel.cs
CharliePoole: This seems to be something that MonoDevelop does, which VS didn't. I'll clean it up.

I really wish GitHub diffs were friendlier WRT blanks though.

src/TestModel/model/TestModel.cs
CharliePoole: Right... it was initially an experiment. I'll reformat.

In the end I decided we should do this for any solution file even if the support is not installed because (1) it doesn't hurt and (2) it means our tests always work the same way, no matter what is installed.

src/tests/mock-assembly/MockAssembly.cs
CharliePoole: Yes. It's not what the author (me in another role) intended the attribute for but it works. By having it here, the mock isn't loaded when we run the solution but it can still be run independently.

It's a somewhat adhoc solution, however, and it could be argued that we simply shouldn't be running the solution file for NUnit at all. You may remember we have had a lot of discussion about this within the NUnit projectx.

src/tests/test-utilities/FormTester.cs
CharliePoole: I agree with the suggested change but I'll fix it along with other stuff rather than using the Beta feature, which would seem to create lots of extra commits.

This should really be in an AssemblyInfo file, but this project doesn't have any at the moment. I could add them at least to the two projects that have the attribute.

As commented above, I'm more or less misusing the feature so our tests pass when you run the solution file. It's intended to be used in FormTester, which references the framework but is not a test file. It's not really aimed at mock-assembly, but it works there.

src/TestModel/model/TestModel.cs
CharliePoole: Yup... that keeps happening. If anybody has suggestions for avoiding, let me know.

src/TestModel/model/TestModel.cs
CharliePoole: See comment above. I originally thought this should only be done if VS project support was installed, but decided it should work the same in either case. I'll remove the comment.

CharliePoole:
Thanks for the comments. I'll clean up and merge.

PR #177, opened 2018-12-03
font changes are now also applied to child items

### LINKED ISSUE #153: Changing font should affect status bar
It does in the NUnit GUI but at the moment not in TestCentric.

drSte:
fixes #153

### DIFF
--- a/src/TestCentric/testcentric.gui/Views/StatusBarView.cs
+++ b/src/TestCentric/testcentric.gui/Views/StatusBarView.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2016-2018 Charlie Poole
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -20,6 +20,8 @@
 // OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
+
+using System;
 
 namespace TestCentric.Gui.Views
 {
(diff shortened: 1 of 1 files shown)


src/TestCentric/testcentric.gui/Views/StatusBarView.cs
```diff
@@ -106,5 +108,12 @@ public void DisplayTime(double time)
                 timePanel.Visible = true;
             });
         }
+
+        protected override void OnFontChanged(EventArgs e)
+        {
+            StatusLabel.Font = testCountPanel.Font = testsRunPanel.Font = passedPanel.Font =
+                failedPanel.Font = warningsPanel.Font = inconclusivePanel.Font = timePanel.Font = Font;
```
CharliePoole: You're setting the font of the individual panels of the StatusBar control rather than setting the font of the control itself. I'm seeing the overall size of the status bar change, which may be due to this. This is on Linux, of course.

CharliePoole: We should do something about the size. I'd say change the maximum statically first and see if that helps. Just make it big enough for the biggest font we expect. If that doesn't work, then we may need to change the size dynamically when the font changes.

CharliePoole: I'm realizing that as compared to NUnit V2, I am using a status strip and not a status bar control. That might be why there's this new problem. I was a lot more familiar with the old control.

CharliePoole:
Let's get it working in Windows then. If I see a problem in Linux after you merge I'll create a new issue.

CharliePoole:
@drSte Is something holding this up or are you able to merge?

drSte:
@CharliePoole, I am a little bit stuck. mostly because lack of free time. If someone is able to do it faster I can step aside.

CharliePoole:
So what needs to be done? Is it just the vertical size adjustment of the status bar?

drSte:
exactly

CharliePoole:
I'll take a look at it.

CharliePoole:
Merging as is. Will look at the height problem separately since it goes beyond the status bar. That is, there are other controls that have to be resized as well.

PR #187, opened 2018-12-18
Upgrade to Cake 0.31.0 use tool on Travis

devlead:
* Upgrade to Cake 0.31.0
* Adjust Travis to use Cake.Tool 0.31.0

Refers to cake-build/cake#2401

Example Travis build:
https://travis-ci.org/devlead/testcentric-gui/builds/469518326

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,11 @@
 ﻿language: csharp
 sudo: false
+dotnet: 2.1.502
 mono:
   - latest
   - 5.10.1
   - 4.6.2
-script: 
+script:
   - git fetch --unshallow
-  - ./build.sh --target=Travis --configuration=Release
+  - dotnet tool install --version "0.31.0" -g Cake.Tool
+  - ~/.dotnet/tools/dotnet-cake --target=Travis --configuration=Release
--- a/tools/packages.config
+++ b/tools/packages.config
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <packages>
-    <package id="Cake" version="0.30.0" />
+    <package id="Cake" version="0.31.0" />
 </packages>


CharliePoole:
This looks good to me. Before I merge, I'm wondering about one thing: what, if anything, do we lose by not using the bootstrap scripts? I'm happy if we don't have to maintain them but I want to be sure.

devlead:
@CharliePoole currently it means that you pin version in 2 places. Step 2 could be to adopt new bootstrappers. We haven't finished the official ones yet, but I have written ones used by other projects, which could be adopted

https://github.com/devlead/CoreWiki/blob/b062aba3b9b88f60a97a2a997c50907bc14fd0bf/build.ps1
https://github.com/devlead/CoreWiki/blob/b062aba3b9b88f60a97a2a997c50907bc14fd0bf/build.sh

And then you would be using Cake.Tool everywhere.

CharliePoole:
Sounds like a reasonable tradeoff now, with more to come.

CharliePoole:
Hey @devlead ... I'm wondering... now that we use the Cake tool on travis... do we have MsBuild available when building for old versions of Mono, where you normally have to use xbuild?

devlead:
Depends on how old version of Mono guess, as only "newer" versions ship with MSBuild.

You could use the version of MSBuild that ships with .NET Core, set framework overrides to point to mono "GAC" and then run unit tests against mono runtime,  but then could be a small risk something in the toolchain generates some IL old mono don't like. Test hopefully would catch that though.

CharliePoole:
Thanks. Sounds like it would be more work than its worth for the minor problem I'm having. I'll just change my project properties so build can deal with it.

I'm still building for 4.6.2 because Debian still ships it.

PR #188, opened 2018-12-20
Add Project nodes to test tree

### LINKED ISSUE #159: Project name not shown in tree
When opening a project with multiple assemblies, the project name is not shown as the root of the tree. Instead, a "Test Run" entry is constructed and used, with the assemblies falling directly under it. In this case, the actual project should be provided as the root of the tree and the Test Run entry suppressed.

The easiest way to see this is to open the TestCentric GUI solution file.

CharliePoole:
Fixes #159 

This isn't ready to merge but I wanted it to be looked at so I made the PR.

It's a great example of the lengths we have to go to if the tree returned by the standard engine release is used. Nothing is present in the tree to indicate a project. I used the info found in the TestPackage to modify the tree of TestNodes and modified TestNode so that a project node creates a filter that combines its child nodes. IOW, the GUI is doing what the engine was supposed to do in the first place.

@mikkelbu I think you have the most experience to review the code. It's not supported by tests - I only spiked it. I have only tested using the solution, i.e. a single project on the command-line. We need to verify it for multiple projects and projects combined with individual asssemblies.

@OsirisTerje I thought you should look at this because it relates to some separate discussion we have had about the support of the engine.

### DIFF
--- a/src/TestModel/model/TestModel.cs
+++ b/src/TestModel/model/TestModel.cs
@@ -180,8 +180,7 @@ namespace TestCentric.Gui.Model
             TestPackage = MakeTestPackage(files);
 
             Runner = TestEngine.GetRunner(TestPackage);
-
-            Tests = new TestNode(Runner.Explore(TestFilter.Empty));
+            Tests = ExploreTestPackage(TestPackage);
             AvailableCategories = GetAvailableCategories();
 
             Results.Clear();
@@ -225,7 +224,7 @@ namespace TestCentric.Gui.Model
 
             TestPackage = MakeTestPackage(TestFiles);
 
-            Tests = new TestNode(Runner.Explore(TestFilter.Empty));
+            Tests = ExploreTestPackage(TestPackage);
             AvailableCategories = GetAvailableCategories();
 
             if (Services.UserSettings.Gui.ClearResultsOnReload)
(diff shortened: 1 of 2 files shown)


CharliePoole:
@mikkelbu @OsirisTerje @ChrisMaddock *

* Chris, I added you because the problem relates to some of the work you are doing on the engine.

While the code isn't any more complicated than lots of other things I've gone ahead and merged, I really hoped for some review of __whether we should do this at all__.

I dislike having the GUI apply "fixes" to make up for stuff the engine doesn't do, but I also dislike the GUI not working. :smiling_imp: 

The alternatives I see are...

1. Put in a (possibly temporary) fixup, as I have done here.
2. Wait for the engine to be fixed and bundle that version with the GUI.
3. Make a copy of the engine for the GUI and fix it here.

Thoughts anyone?

ChrisMaddock:
> 2. Wait for the engine to be fixed and bundle that version with the GUI.
> 3. Make a copy of the engine for the GUI and fix it here.

What's the reason for not fixing this in the nunit-console repo, sorry? 🤔

CharliePoole:
@ChrisMaddock That was item number 2. I did update one of the issues to indicate what I think ought to be done.

In terms of desirablility, I prefer (2) even though it exposes me to changes in the engine. Both (1) and (3) eliminate that risk. Of those two, (3) is quite a lot of work, while (1) is easy - done in fact. Thinking about it, I suppose (1) is a good interim measure while waiting to see if (2) can happen.

It's a bit of a puzzle!

ChrisMaddock:
@CharliePoole Maybe I'm reading too much in to this, but (2) had a distinctly passive voice. To be more direct about my original question - would you personally be interested in contributing a fix for this issue to the NUnit Engine?

I'm just confused about the motivation to fork a copy of the engine for this particular issue. I appreciate, as you say, that you become "exposed to changes in the engine" - but surely anything that causes a negative effect is poor work in the engine? It seems to me like our priorities are pretty aligned, at this point in time.

------------

And just to actually respond to the question you invited me here for, my opinion would be that the fix belongs in the engine, and if time is going to be spent writing a fix for this issue in any location, I'd personally much prefer that to be a contribution to the NUnit Engine. 😄 In my mind, the TestCentric GUI is a consumer of the engine, and therefore will have it's own feature requests and bug reports to contribute, which are thus-far not exposed in by any other consumer.

CharliePoole:
I think we agree. I might merge in my existing fix temporarily, however, just so I can see the expected tree in the GUI and can select and run a single solution.

Regarding the passive voice, I'll respond offline.

CharliePoole:
As discussed, I'll merge this change for now with the intent to remove it one way or the other when we have a fix in the engine that we are using.

PR #249, opened 2018-12-24
Show tests in categories of their parents

### LINKED ISSUE #238: Categories on test fixtures not considered in test list
I use the `CategoryAttribute` on individual test cases. When all test cases of a test fixture belong to a specific category, I place the `CategoryAttribute` right on the test fixture.
I expect this to be functionally equivalent as if I had marked each test case, when dealing with test cases.

However, in the test list view, when grouping by categories, categories defined by means of the test fixtures are not considered.

The following code:
```C#
[Category("a")]
public class A
{
    [Test]
    public void A1() {
    }

    [Test]
    public void A2()
    {
    }
}

public class B
{
    [Test]
    [Category("a")]
    public void B1()
    {
    }

    [Test]
    [Category("a")]
    public void B2()
    {
    }
}
```
produces this grouping: 
this

I'd expect A1 and A2 to appear under the category `a`.

I should be able to make a PR for this.

CharliePoole:
Fixes #238 

This PR reproduces TestCentric/testcentric-experimental-gui#183 submitted to the experimental gui project by @chm-tm

Since the original PR did not get merged before I moved the experimental gui here, I'm re-creating it so we can decied what to do.

The PR works and does what it says. The question that remains is whether we actually want category groupings to work that way or as originally designed. More discussion here or on the issue.


CharliePoole:
Looking this over, it seems to me that this is useful feature. OTOH, listing by __strict__ categories is also a useful feature. I'm thinking of adding `By Category (strict)` to the list of sorting options. Thoughts?

I'm also wondering whether this dropdown has gotten two complex. If you want to select test list and change the sort, you have to open the menu twice. It might be better to pop up a little dialog that lets you choose both at once. Then the category option could come with a separate checkbox for including / excluding parent categories. Thoughts on this?

immeraufdemhund:
if i want to do two different things i would expect two different accesses to the menu. i'm ok with having to do it twice.

CharliePoole:
Rebasing...

CharliePoole:
I'm going ahead and including this in the alpha3 release so we don't lose it. I added a new menu item "By Category (Extended)" to enable use of the new feature. I don't think this is the final UI, but it keeps it in our faces so we won't forget.

PR #250, opened 2018-12-24
Basic debugging support

### LINKED ISSUE #235: Debugging support
Are you open for a pull request to implement debugging support in the test runner: Starting a test with a debugger attached. Part of running and developing tests is also debugging, and besides Resharper there is not really a suitable graphical test runner available for NUnit 3.

CharliePoole:
Fixes #235 

This PR reproduces TestCentric/testcentric-experimental-gui#185 submitted to the experimental gui project by @Sebazzz

Since the original PR did not get merged before I moved the experimental gui here. With all the file rearranging that has happened, I had to re-create this manually.

@Sebazzz can you check this over to make sure I captured all the changes you had done in the other PR? Once that's done we can talk about any changes and merge it.


Sebazzz:
Sure, let me check. Should have some time between the holidays in 👍

Sebazzz:
LGTM

There is one unrelated test failing, I think it is flaky and not related to this PR.

CharliePoole:
I noticed that test started failing intermittently. It is unrelated however.

So if this looks right to you, we are back where we were when the original PR was being neglected. :frowning_face: Let's do better this time. :smile: 

@immeraufdemhund Made some comments in the original PR TestCentric/testcentric-experimental-gui#185, which I will fix next.

I see two general questions around this for @TestCentric/gui-team to discuss...

1. Do we like this approach - albeit basic - to debugging, at least for now?

2. Is this the UI we want. Specifically, should we have two buttons or one button that lists both the run and debug options? (Note that the button is supposed to remember what it was last used for, but that's broken att he moment.)

CharliePoole:
@Sebazz @mikkelbu @ManfredLange @immeraufdemhund 

What are your thoughts on the interface? Currently, we have a single SplitButton for Run All, Run Selected and RunFailed. (The button is supposed to default to the last selection used, but that currently is not working so it's always Run All unless you use the dropdown.)

Additionaly, for running a single selected item, you can use a right click context menu.

As @Sebazzz has implemented it, there is a new SplitButton, which works just like the Run button but does debug. You can Debug All, Selected or Failed tests.

An alternative would be to use the single button adding a divider line and the three Debug options to the dropdown.

Another alternative would be to have no way to debug from the tool bar and force reliance on the context menu.

What do you think the UI should look like?

CharliePoole:
@mikkelbu The exception when debuging a solution is related to the engine issue I just filed. Both the console and the vs adapter use the engine in a very simple way: load-explore-run-exit. The GUI exercises bits like Reload that have not been used much before.

I'm going ahead and merging this... we can always change the UI if we decide we want it to be different.

src/TestModel/model/TestModel.cs
```diff
@@ -223,7 +223,7 @@ public void ReloadTests()
 
             Runner.Reload();
 
-            _lastRunWasDebugRun = false;
+<           _lastRunWasDebugRun = false;
```
mikkelbu: The merge introduced a problem here

CharliePoole: That was me trying out the web editor. :-)

PR #281, opened 2019-02-24
Include PDB Files in packages

### LINKED ISSUE #278: Include pdb files in packages
This is probably the most straightforward way to enable debugging, even though it increases package size.

CharliePoole:
Fixes #278


CharliePoole:
After several attempts to get the pdbonly setting to work under XBuild I gave up. Mono 4.6.2 is not a platform we use to produce distributed builds, so the built packages now include pdbs only on versions of mono that support MsBuild.

PR #282, opened 2019-02-25
Switch to use of file menu for process and domain settings

### LINKED ISSUE #274: Eliminate Default Setting Page for process model and domain usage?
This is currently a question for the members of the @TestCentric/gui-team @TestCentric/contributors and others.Let's try to answer quickly in order to make this into an actual feature if we are in agreement.

The NUnit V2 GUI had a settings page for the default process model and appdomain. The TestCentric GUI inherited it but I removed it from the Experimental GUI in favor of File menu items for selecting process model, and domain usage.

I think the use of the menu, without saving the setting as a general default, is to be prefered. Reasons:

1. The setting overrides whatever is set within an NUnit project. It seems like it's a bad idea to override the project settings by default, since they are usuallly set up because the particular tests require a certain setting.

2. A hidden default setting, perhaps set by someone else or by the same user and then forgotten, can be very confusing. It seems simpler to load assemblies using the defaults each time and only reload them in a special way if the user requires it for some reason.

3. As now implemented, there are two defaults: the one set on the settings page and the "default default" that is used when nothing is specified.

If I implemented this, I'd make process model and domain usage into File menu items just like the experimental GUI. The number of agents setting could be moved to a different settings page.

I'd appreciate your comments in the next few days so this can be modified if necessary and added to the backlog. Whatever is implemented here, I'd like to do it before dealing with some of the outstanding engine errors that we are experiencing, since this feature would actually make it easier to test those issues.

CharliePoole:
Fixes #274 

The menu is implemented exactly as in the experimental GUI. That is, the setting is persisted before being used to reload the tests. That means the next time the app is re-opened, the last setting will be used.

It's possible we may want to change this in the future, but it should work the same way in both GUIs.


CharliePoole:
@immeraufdemhund You're on top of things today! :smiley:

PR #291, opened 2019-03-14
Re-display poperties when Display hidden checkbox …

### LINKED ISSUE #288: Property dialog not showing assembly process id or appdomain name
The properties exist in the XML but when switching on private properties, they don't appear.

CharliePoole:
Fixes #288


CharliePoole:
This turned out to be very simple... call was commented out.

Issue #241, opened 2017-09-05
Terminate nunit-agent.exe / nunit-agent-x86.exe once they are no longer in use


jerryk414:
The agents used to execute tests via nunit-gui.exe are left around for the duration of the application even when they are no longer in use.

For example, when loading/refreshing a test assembly, a new agent is spawned for testing.  When this happens, the old agent becomes "inactive" - where it will never be used again, but it still sticks around.

This becomes a nuisance when attempting to debug tests via attaching the source to the executing agent - if there are multiple agents, you have to inspect which one is currently using CPU to determine which agent is active.

This also occurs when switching the selected Process Model.

![image](https://user-images.githubusercontent.com/11689280/30074312-1de34afe-923f-11e7-9b7e-0c781cd295a8.png)
![image](https://user-images.githubusercontent.com/11689280/30074319-20d52264-923f-11e7-9d3f-68017483abe8.png)

PR #298, opened 2019-03-25
Disposed Runner when unloading Tests so the Agent process terminates

JeffKryzer:
I'm pretty sure this issue is what is being referred to in issue #241.

### DIFF
--- a/src/TestModel/model/TestModel.cs
+++ b/src/TestModel/model/TestModel.cs
@@ -202,6 +202,7 @@ namespace TestCentric.Gui.Model
             _events.FireTestsUnloading();
 
             Runner.Unload();
+            Runner.Dispose();
             Tests = null;
             AvailableCategories = null;
             TestPackage = null;


CharliePoole:
Were you able to confirm the issue with either the current master or the beta release?

JeffKryzer:
First repro-ed it when downloading the beta release.  Later pulled down the master source code and confirmed it after compiling.

CharliePoole:
We once had some ability to re-use agents, depending on the nature of the change. For example, changing the framework or x86 setting requires a new agent but changing reload-on-run would not.

Somewhere in the depths of time, I think this was lost. I want to check to see if it can be put back in a timely fashion. If not, then I think we should go with your solution for the 1.0 release.

This question may also apply to one or two others of your issues, IIRC.

CharliePoole:
Hey! Thanks for your contribution!

JeffKryzer:
You're welcome, sir.  I'm on a project where we just upgraded from NUnit 2.6 to 3.0 and suddenly we didn't have a GUI to use.  Needed to get it up and running to match functionality from NUnit GUI 2.x, and as you can probably guess, we're running our tests as x86 which immediately failed.  IMO, running the tests inside of Visual Studio just sucks,  So much recompiling just to run a test!

CharliePoole:
The general idea is that the 1.0 release should provide equivalent functionality to the V2 GUI. In a pinch, I may release 1.0 and push anything that's less used to 1.1.

Feel free to write issues where you see missing functionality, if there isn't already one.

Issue #299, opened 2019-03-25
Choosing "Run as X86" from menu doesn't actually load the X86 agent.


JeffKryzer:
1.  Run testcentric.exe.
2.  Load a test assembly.
3.  Open Task Manager and look at processes.  You will notice **nunit-agent.exe** in the list.
4.  Choose File --> Process Model --> Run as X86
5.  You are prompted to reload.  Click Yes.
6.  Look in Task Manager again.  **nunit-agent.exe** is still in the list, rather than nunit-agent-x86.exe.

In addition, running a test assembly which requires 32-bit will fail, confirming that it is still running in a 64-bit process.

CharliePoole:
I'm wondering if we even want the dialog that asks if you want to reload. Why would you say no?

JeffKryzer:
I agree.  But the prompt is in the generic ChangePackageSetting method which is used for changing all settings.  Suggestions?  Pass in a bool like changeRequiresPrompt (forcing the caller to be in charge)?  Or hard code a list of settings which can skip the prompt?  Or just leave it.  You could argue the prompt gives validation that it is being reloaded.  You could remove the Yes/No and just change it to an OK button.

CharliePoole:
It does get complicated! I'll look more closely in the morning.

CharliePoole:
@JeffKryzer Can you confirm that  loading a test assembly that is compiled to target X86 runs the X86 agent even if the menu is not checked? That's what is supposed to happen. If it isn't working, that's an even more serious error.

The X86 option actually means __force__ it to run as X86.

CharliePoole:
Upon review, I see that the ChangePackageSettings method is private and only used for changes to framework, process model, domain usage and runasx86. I think we can reload in all those cases.

Method name could be changed to ChangePackageSettingsAndReload for clarity.

PR #300, opened 2019-03-25
Altered ChangePackageSetting to do an unload/load rather than a reload.

JeffKryzer:
A reload does not use the setting that you just changed, so the unload/load is necessary.  It re-creates the Engine with your new settings.  This fixes issue #299.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
@@ -25,6 +25,7 @@ using System;
 using System.Collections.Generic;
 using System.Drawing;
 using System.IO;
+using System.Linq;
 using System.Text;
 using System.Windows.Forms;
 using NUnit.Engine;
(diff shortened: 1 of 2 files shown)


src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
```diff
@@ -748,7 +749,13 @@ private void ChangePackageSetting(string key, object setting)
             string message = string.Format("New {0} setting will not take effect until you reload.\r\n\r\n\t\tReload Now?", key);
 
             if (_view.MessageDisplay.Ask(message) == DialogResult.Yes)
-                _model.ReloadTests();
+            {
```
CharliePoole: Let's eliminate the message display and test and just do it. We can change the name to something that says what it actually does... like "ChangePackageSettingsAndReload.

JeffKryzer: Sorry for the delay, was on vacation.  Will try to address in next couple days.

JeffKryzer:
Okay, prompt removed in latest commit.

Issue #301, opened 2019-03-25
Tree icons of previous run do not clear when starting second run


JeffKryzer:
If I run my test on a tree with 20 nodes, they will color in (red/green) one by one as the tests run.  If I press Run a second time, the previous results do not clear, so each node is still colored and I can't tell the result of each node like I could the first time.

My expectation is that the nodes would go back to gray immediately when I press the Run button, identical to NUnit 2.0 gui.

CharliePoole:
@JeffKryzer Thanks for all the reports! I'll get to this one as well as the others in the next day or so.

JeffKryzer:
This bug can be solved by adding 1 line of code to TestCentricPresenter.RunTests:

        public void RunTests(TestNode[] tests)
        {
            (_view.TreeView.Tree.TopNode as TestSuiteTreeNode)?.ClearResults();
...

Not sure if you find that to be acceptable.  If so, I'll submit it.

CharliePoole:
That would only catch runs started from the main presenter, not, for example, by right clicking in the tree.

A better way to handle this is in the TreeViewPresenter when a run starting event is detected. That way, it will happen no matter how the run is started, even by some new command we add in the future.

The reason results are not currently cleared, BTW, is that I had been fooling around in the experimental gui with the idea of keeping old (partial) runs but dimming the color each time a new run took place. However, that's for the future and right now clearing makes the most sense.

JeffKryzer:
Indeed you are correct.  I will pursue your suggestion.

CharliePoole:
@JeffKryzer Are you still pursuing this one?

CharliePoole:
FYI: If you put Fixes #301 or Closes #301 in the initial description, then the issue would be closed automatically once the PR is merged.

JeffKryzer:
Oh thanks for the tip.  This is actually the first GitHub project I've worked on, so appreciate the advice.

CharliePoole:
@JeffKryzer 

I merged your fix last year and it has been included in all releases siince 1.0. Strangely, I'm now seeing this bug in the current code. Have you noticed a reversion in any of the releases you have installed?

My working assumption is that something at a lower level that your fix relied on has changed.

PR #307, opened 2019-04-02
Cleared results of previous run before starting another run.

JeffKryzer:
Fixes issue #301.


CharliePoole:
Thanks for doing this!

PR #335, opened 2019-04-26
Make sure to use invoke when accessign treeview control

### LINKED ISSUE #333: InvalidOperationException is thrown from RunStarting event handler
After clicking Run button the following event handler gets invoked:
``` C#
            _model.Events.RunStarting += (e) =>
            {
                (_view.Tree.TopNode as TestSuiteTreeNode)?.ClearResults();
                _view.RunCommand.Enabled = false;
                _view.CheckPropertiesDialog();
            };
```
Accessing property `TreeViewElement.TopNode` throws the following exception:
`System.InvalidOperationException: 'Cross-thread operation not valid: Control 'treeView1' accessed from a thread other than the thread it was created on.'`
This exception is not visible to the end user but prevents the following lines from executing.

CharliePoole:
Fixes #333

Also fixed a few other places that obviously needed the same treatment.

### DIFF
--- a/src/Common/components/Elements/TreeViewElement.cs
+++ b/src/Common/components/Elements/TreeViewElement.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2015 Charlie Poole
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -110,21 +110,51 @@ namespace TestCentric.Gui.Elements
 
         public TreeNode TopNode
         {
-            get { return _treeView.TopNode; }
-            set { _treeView.TopNode = value; }
+            get
+            {
+                TreeNode topNode = null;
+                InvokeIfRequired(() => topNode = _treeView.TopNode);
+                return topNode;
+            }
+            set
+            {
+                InvokeIfRequired(() => _treeView.TopNode = value);
+            }
         }
 
-        public int VisibleCount => _treeView.VisibleCount;
+        public int VisibleCount
+        {
+            get
+            {
+                int visibleCount = 0;
+                InvokeIfRequired(() => visibleCount = _treeView.VisibleCount);
+                return visibleCount;
+            }
+        }
+
 
         public TreeNode SelectedNode
         {
-            get { return _treeView.SelectedNode; }
-            set { _treeView.SelectedNode = value; }
+            get
+            {
+                TreeNode selectedNode = null;
+                InvokeIfRequired(() => selectedNode = _treeView.SelectedNode);
+                return selectedNode;
+            }
+            set
+            {
+                InvokeIfRequired(() => _treeView.SelectedNode = value);
+            }
         }
 
         public TreeNodeCollection Nodes
         {
-            get { return _treeView.Nodes; }
+            get
+            {
+                TreeNodeCollection nodes = null;
+                InvokeIfRequired(() => nodes = _treeView.Nodes);
+                return nodes;
+            }
         }
 
         public IList<TreeNode> CheckedNodes => GetCheckedNodes();


CharliePoole:
@TestCentric/gui-team Wish I could figure out a way to unit test for this. Hard to do a negative test, however. We want... "controls are never accessed using the wrong thread!"

src/Common/components/Elements/TreeViewElement.cs
```diff
@@ -110,21 +110,51 @@ public bool CheckBoxes
...
+                InvokeIfRequired(() => _treeView.TopNode = value);
+            }
+        }
+
+        public int VisibleCount
+        {
+            get
+            {
+                int visibleCount = 0;
+                InvokeIfRequired(() => visibleCount = _treeView.VisibleCount);
```
immeraufdemhund: is this true for get? I thought it was set was the only problem

src/Common/components/Elements/TreeViewElement.cs
```diff
@@ -110,21 +110,51 @@ public bool CheckBoxes
 
         public TreeNode TopNode
         {
-            get { return _treeView.TopNode; }
-            set { _treeView.TopNode = value; }
+            get
+            {
+                TreeNode topNode = null;
+                InvokeIfRequired(() => topNode = _treeView.TopNode);
```
immeraufdemhund: is this true for get? I thought it was set was the only problem

src/Common/components/Elements/TreeViewElement.cs
```diff
@@ -110,21 +110,51 @@ public bool CheckBoxes
...
-        public int VisibleCount => _treeView.VisibleCount;
 
         public TreeNode SelectedNode
         {
-            get { return _treeView.SelectedNode; }
-            set { _treeView.SelectedNode = value; }
+            get
+            {
+                TreeNode selectedNode = null;
+                InvokeIfRequired(() => selectedNode = _treeView.SelectedNode);
```
immeraufdemhund: is this true for get? I thought it was set was the only problem

src/Common/components/Elements/TreeViewElement.cs
```diff
@@ -110,21 +110,51 @@ public bool CheckBoxes
...
+            }
         }
 
         public TreeNodeCollection Nodes
         {
-            get { return _treeView.Nodes; }
+            get
+            {
+                TreeNodeCollection nodes = null;
+                InvokeIfRequired(() => nodes = _treeView.Nodes);
```
immeraufdemhund: is this true for get? I thought it was set was the only problem... This one i'm not positive about because do we ever edit or change the child nodes? or do we just replace the entire tree view?

CharliePoole:
> Not sure if getters needs to be wrapped or not.. If they do then disregard my thoughts

I didn't think so either, which is why they generally haven't been. But the exception was thrown on a getter for `TopNode` and when I added the wrapper code it stopped throwing.

One likely theory is that it depends on just what the getter does. TopNode probably has to do a lot of windows stuff to figure out which nodes are in the window at this time.

But some of my other "fixes" may not be necessary.

src/Common/components/Elements/TreeViewElement.cs
CharliePoole: I have no evidence for this field, but it sounded like another one where Windows may need to do a bunch of work in order to figure out what's visible. On the one hand, I don't want a bunch of unneeded wrapping. On the other, it __should__ work even if it's overkill.

src/Common/components/Elements/TreeViewElement.cs
CharliePoole: This one is definitely confirmed as the source of the exception.

src/Common/components/Elements/TreeViewElement.cs
CharliePoole: Again, not sure. Could be overkill.

src/Common/components/Elements/TreeViewElement.cs
CharliePoole: We definitely modify the property by setting a bunch of nodes into the tree. Does that mean we should wrap on get? Not sure.

CharliePoole:
So the key thing here is whether to fix anything __except__ the one thing that was demonstrated to need it. Problem is that these thread errors are pretty obscure and hard to find. This one had been lurking unfound for a while until we got the issue.

My preference would be not to do anything until proven necessary. But I'd also prefer if we had a good way to scan for such exceptions in the course of an execution. Thoughts?

src/Common/components/Elements/TreeViewElement.cs
immeraufdemhund: well if it is the source of the problem then all the others need it and therefor all my concerns are gone

CharliePoole: I think that's a good observation. If there is some field where we __havent__ needed it up to now that's because we haven't yet used it on a context that requires invoking on the correct thread. For example, maybe we do it before the window is created. But that's not robust. Really, everything should be wrapped. I'll look to see if there is more to do here.

CharliePoole:
There may be more wrapping required, but I'm merging this for now in order to resolve the issue.

PR #339, opened 2019-04-30
propagate InternalTraceLevel to the agent

ideafixxxer:


CharliePoole:
I found that we had already added this to the package settings at the time of loading the test, but doing it upon starting the app makes more sense as it takes effect sooner. Thanks!

PR #354, opened 2019-06-03
Detect and report unhandled exception

### LINKED ISSUE #340: Unable to run selected NUnit V2 tests
Tested with TestCentric Runner for NUnit 1.0.0-Beta - April 13, 2019 installed from Chocolatey.

In a situation where you have mixed test assemblies, that is test assemblies for NUnit V2 and NUnit V3, it appears that running NUnit V2 test assemblies does not work as intended.

The primary Run button as well as the Run options from the Tests menu seems to work as intended for NUnit V3 test assemblies.

**For NUnit V2 test assemblies only "Tests -> Run All" works.**
Running selected or failed tests from the Tests menu does nothing and the Run button also does nothing.
The tests were carried out using the attached test assembly from https://github.com/nunit/nunit/issues/3231

CharliePoole:
This is a partial fix for #340. It reports the error by showing a message for the exception, indicating that an id filter may not be used with NUnit V2.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
@@ -185,14 +185,20 @@ namespace TestCentric.Gui.Presenters
                 //    //log.Warning("Unable to save result to {0}\n{1}", resultPath, ex.ToString());
                 //}
 
-                if (e.Result.Outcome.Status == TestStatus.Failed)
-                    _view.Activate();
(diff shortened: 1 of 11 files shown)


CharliePoole:
@TestCentric/gui-team Request for comment: Is this worth merging as it stands? If we merge this, cases where we silently fail running NUnit V2 tests, will now show a message indicating why they failed. However, it won't actually fix anything.

immeraufdemhund:
it bugs me to no end when I get no exceptions or no messages in visual studio why something didn't run/failed etc etc. If it helps debug something I'm all for it

PR #356, opened 2019-06-24
Add force parameter to TestModel.StopRun

CharliePoole:
I have changed this to do a force stop when the Stop button is pressed. The thing is, it requires a fix to the NUnit framework in order to work correctly. I have submitted a PR with the fix.

So the question is whether to merge this or to try to make it work even with the existing build of the framework. @TestCentric/gui-team What do you think?

### DIFF
--- a/src/Experimental/experimental-gui/Presenters/TreeViewPresenter.cs
+++ b/src/Experimental/experimental-gui/Presenters/TreeViewPresenter.cs
@@ -142,7 +142,7 @@ namespace TestCentric.Gui.Presenters
             _view.RunAllCommand.Execute += () => RunAllTests();
             _view.RunSelectedCommand.Execute += () => RunTests(_selectedTestItem);
             _view.RunFailedCommand.Execute += () => RunAllTests(); // RunFailed NYI
-            _view.StopRunCommand.Execute += () => _model.CancelTestRun();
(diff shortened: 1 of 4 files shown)


CharliePoole:
@immeraufdemhund So to be clear, I created a PR for this in the framework, not the engine. The difference is that the GUI can be upgraded to use a new engine version, but the framework version in use depends on the user. Does that change anything, in your view?

That said, I __believe__ it is possible to do a workaround in either the runner or the engine that would compensate for older framework versions that don't handle forced stoprun correctly. If we had such a workaround, it might be better in the engine than in the GUI, since that would help all runners.

immeraufdemhund:
hmm.. yeah you are right that does change things a bit mentally. because we have many old projects that don't stay up to date with framework. yeah maybe a work around would be best than

CharliePoole:
Now that I have working code in PR nunit/nunit#3281 I can see about doing some sort of workaround.

In that PR, I keep track of any suites that started running but never terminated and I signal them as complete in case of an abort. I think the same thing can be done outside of the framework, by looking at the progress events as they are received.

That could be done in the GUI or in the engine. I'm inclined to test it in our copy of the engine and then propose it to the NUnit team as a fix.

Think I'll merge this and do that as a separate PR.

PR #359, opened 2019-06-29
Handle non-terminating Cancel Run in engine

### LINKED ISSUE #337: Stopping a long running test from TestCentric gui does not stop right away
If you press the Stop button when a long test is running, it doesn't stop the test right away.  It waits until that test finishes.  Only then does the Agent stop and the Run button is re-enabled.

This is different than nunit_gui 2.x which stopped the test immediately.

Note, if you close the TestCentric gui window when a long running test is running, it DOES stop and kill the agent right away as it is closing.  So that's good.

CharliePoole:
This PR closes #337 by handling the case where the version of the NUnit 3 framework in use has not been patched to handle a forced StopRun correctly. It delays five seconds and then, if all outstanding suites have not signaled termination, if does it for them and ends the run. This eliminates the apparent hang in the GUI.

Note that the change has no impact on actually cancelling the run. The error we are compensating for seems to have to do with __notifying__ the caller about the cancellation, which appears to happen correctly.

### DIFF
--- a/src/TestEngine/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
+++ b/src/TestEngine/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
@@ -65,6 +65,9 @@ namespace NUnit.Engine.Runners
         private ITestRunnerFactory _testRunnerFactory;
         private bool _disposed;
 
+        private TestEventDispatcher _eventDispatcher = new TestEventDispatcher();
+        private WorkItemTracker _workItemTracker = new WorkItemTracker();
+
         public MasterTestRunner(IServiceLocator services, TestPackage package)
         {
             if (services == null) throw new ArgumentNullException("services");
@@ -198,6 +201,31 @@ namespace NUnit.Engine.Runners
         public void StopRun(bool force)
         {
             _engineRunner.StopRun(force);
+
+            if (force && !_workItemTracker.AllItemsComplete.WaitOne(5000))
+            {
+                // Some suites never completed, we simulate completion for them
+                int count = _workItemTracker.ItemsInProcess.Count;
+                while (count > 0)
+                {
+                    XmlNode startItem = _workItemTracker.ItemsInProcess[--count];
+                    XmlNode suiteItem = XmlHelper.CreateTopLevelElement("test-suite");
+                    suiteItem.AddAttribute("type", startItem.GetAttribute("type"));
+                    suiteItem.AddAttribute("id", startItem.GetAttribute("id"));
+                    suiteItem.AddAttribute("name", startItem.GetAttribute("name"));
+                    suiteItem.AddAttribute("fullName", startItem.GetAttribute("fullname"));
+                    suiteItem.AddAttribute("result", "Failed");
+                    suiteItem.AddAttribute("label", "Cancelled");
+                    XmlNode failure = suiteItem.AddElement("failure");
+                    XmlNode message = failure.AddElementWithCDataSection("message", "Test run cancelled by user");
+                    _eventDispatcher.OnTestEvent(suiteItem.OuterXml);
+                }
+
+                IsTestRunning = false;
+
+                // Simulate completion of the run
+                _eventDispatcher.OnTestEvent($"<test-run id='{TestPackage.ID}' result='Failed' label='Cancelled' />");
+            }
         }
 
         /// <summary>
@@ -426,12 +454,15 @@ namespace NUnit.Engine.Runners
         /// <returns>A TestEngineResult giving the result of the test execution</returns>
         private TestEngineResult RunTests(ITestEventListener listener, TestFilter filter)
         {
-            var eventDispatcher = new TestEventDispatcher();
+            _workItemTracker.Clear();
+            _eventDispatcher.Listeners.Clear();
+            _eventDispatcher.Listeners.Add(_workItemTracker);
+
             if (listener != null)
-                eventDispatcher.Listeners.Add(listener);
+                _eventDispatcher.Listeners.Add(listener);
 #if !NETSTANDARD1_6
             foreach (var extension in _extensionService.GetExtensions<ITestEventListener>())
-                eventDispatcher.Listeners.Add(extension);
+                _eventDispatcher.Listeners.Add(extension);
 #endif
 
             IsTestRunning = true;
@@ -461,9 +492,9 @@ namespace NUnit.Engine.Runners
                 InsertCommandLineElement(startRunNode);
 #endif
 
-                eventDispatcher.OnTestEvent(startRunNode.OuterXml);
+                _eventDispatcher.OnTestEvent(startRunNode.OuterXml);
 
-                TestEngineResult result = PrepareResult(GetEngineRunner().Run(eventDispatcher, filter)).MakeTestRunResult(TestPackage);
+                TestEngineResult result = PrepareResult(GetEngineRunner().Run(_eventDispatcher, filter)).MakeTestRunResult(TestPackage);
 
                 // These are inserted in reverse order, since each is added as the first child.
                 InsertFilterElement(result.Xml, filter);
@@ -481,7 +512,7 @@ namespace NUnit.Engine.Runners
 
                 IsTestRunning = false;
 
-                eventDispatcher.OnTestEvent(result.Xml.OuterXml);
+                _eventDispatcher.OnTestEvent(result.Xml.OuterXml);
 
                 return result;
             }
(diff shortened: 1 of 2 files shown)


CharliePoole:
@ChrisMaddock Since I plan to submit this as a change to the nunit-console project as well, it would probably make sense for you to review it here, so both versions are kept in sync.

src/TestEngine/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
```diff
@@ -198,6 +201,31 @@ public ITestRun RunAsync(ITestEventListener listener, TestFilter filter)
         public void StopRun(bool force)
         {
             _engineRunner.StopRun(force);
+
+            if (force && !_workItemTracker.AllItemsComplete.WaitOne(5000))
```
ChrisMaddock: You commented elsewhere that this was the right approach, not because it's a specific fix to the NUnit 3 problem (which should perhaps live in the driver), but because it could apply to any other framework which doesn't implement a forced-stop run.

That was a very good point, that I hadn't considered in the implementation of this fix - mind adding a comment here along those lines? I think it could initially be confusing that the MasterTestRunner is terminating suites itself, in certain situations.

src/TestEngine/src/NUnitEngine/nunit.engine/Runners/WorkItemTracker.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System;
```
ChrisMaddock: I'm not sure what approach you're taking to copyright's in this fork, but just to highlight that this would of course want the copyright header in a PR to the NUnit engine. 🙂

src/TestEngine/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
CharliePoole: Will do!

src/TestEngine/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
```diff
@@ -198,6 +201,31 @@ public ITestRun RunAsync(ITestEventListener listener, TestFilter filter)
...
+                // Some suites never completed, we simulate completion for them
+                int count = _workItemTracker.ItemsInProcess.Count;
+                while (count > 0)
+                {
+                    XmlNode startItem = _workItemTracker.ItemsInProcess[--count];
+                    XmlNode suiteItem = XmlHelper.CreateTopLevelElement("test-suite");
+                    suiteItem.AddAttribute("type", startItem.GetAttribute("type"));
+                    suiteItem.AddAttribute("id", startItem.GetAttribute("id"));
+                    suiteItem.AddAttribute("name", startItem.GetAttribute("name"));
+                    suiteItem.AddAttribute("fullName", startItem.GetAttribute("fullname"));
```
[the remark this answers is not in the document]
CharliePoole: An error. We need a test somewhere!

src/TestEngine/src/NUnitEngine/nunit.engine/Runners/WorkItemTracker.cs
CharliePoole: Will do.

src/TestEngine/src/NUnitEngine/nunit.engine/Runners/WorkItemTracker.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        public ManualResetEvent AllItemsComplete = new ManualResetEvent(false);
+        
+        public void Clear()
+        {
+            ItemsInProcess.Clear();
+            AllItemsComplete.Reset();
+        }
+
+        void ITestEventListener.OnTestEvent(string report)
```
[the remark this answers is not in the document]
CharliePoole: That's a good point. What I've done here assumes single threading, which is what we are generally doing in the engine. There are a few problems with that:

1. The NUnit 3 framework ensures that events are serialized and sent on the same thread, but other frameworks might not. This is more of a hypothetical problem.

2. We might be running more than one assembly in parallel, in which case these calls should really be treated as potentially multi-threaded. So this list (or map) of pending items should be protected. Good catch.

src/TestEngine/src/NUnitEngine/nunit.engine/Runners/WorkItemTracker.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        private void RemoveItem(string id)
+        {
+            foreach (XmlNode item in ItemsInProcess)
+            {
+                if (item.GetAttribute("id") == id)
+                {
+                    ItemsInProcess.Remove(item);
+                    return;
+                }
```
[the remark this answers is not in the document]
CharliePoole: Yes, I did the "simplest thing that could possibly work" first, but that's an obvious improvement before merging.

src/TestEngine/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
CharliePoole: BTW, I have no idea whether 5 seconds is the right wait time here, any thoughts?

src/TestEngine/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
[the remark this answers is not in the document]
CharliePoole: Fixed for now.

src/TestEngine/src/NUnitEngine/nunit.engine/Runners/WorkItemTracker.cs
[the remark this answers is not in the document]
CharliePoole: Hmm... It's doable, but a bit more complicated than I like right now. We would need two structures, a map and a list, because we need to maintain the order that the suites started and terminate them in reverse order. That means we would have to add to and delete from both structures. So for now, so long as we only track suites, I'd prefer to leave it alone.

CharliePoole:
The latest commit takes care of most of the issues that @ChrisMaddock raised.

The primary change is that the tracking data for work items in process is now protected by a lock. I refactored in order to make that lock a private member because locking on exposed members is a sure recipe for future deadlocks.

The "database" of pending work items is still just a list. I think that will need to be changed if / when we expand this to include more than test suites. It's entirely private now, so easy enough to change when the time comes.

src/TestEngine/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
ChrisMaddock: Given this is a force stop, and not waiting for tear downs to run, 5 seconds sounds plenty enough to me!

src/TestEngine/src/NUnitEngine/nunit.engine/Runners/WorkItemTracker.cs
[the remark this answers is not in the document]
ChrisMaddock: Is order important? Should a listener not handle events based on their id, rather than order? Again - not an area I'm too familiar with. 🙂

src/TestEngine/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
CharliePoole: I'm tempted to shorten it because it's a little annoying when I run tests but I normally don't cancel runs very often, so it won't bother me in actual usage. Hope that's true for everyone. If it gets annoying, I might look at displaying a message like"Waiting for run to be cancelled."

src/TestEngine/src/NUnitEngine/nunit.engine/Runners/WorkItemTracker.cs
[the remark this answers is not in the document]
CharliePoole: We have always supported a guarantee of correct ordering of events pertaining to the same test. For example, the test can't end before it starts. That also extends to nested tests (suites) so we can't allow the namespace to finish until all contained fixtures finish. Using an ordered list is a simplified way of handling this without explicit references to the structure of the tests.

src/TestEngine/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
ChrisMaddock: Actually yes, I'd agree. 5s is probably excessive - you'd expect a force stop to be a much quicker action! 

One other thought on that...what happens if the force stop is slow, and the framework returns the events at 5.1 seconds. The duplicate events might cause problems, right? Should we disconnect the original listener before we start creating duplicate events in the MasterTestRUnner?

CharliePoole: We'd have to create a new dispatcher and somehow disconnect the old one. The framework has a reference to the old one of course.

CharliePoole: So, thought this over... We could...

1. Give `TestEventDispatcher` an `Enabled` property or `Enable()` method.

2. If our timeout fails, disable the current event dispatcher. If any more events come from the framework, it will get them but ignore them.

3. Create a new TestEventDispatcher. Since we haven't passed this one to any downstream runners, it won't get any events except those we generate.

So, it's doable... but I'm not sure if it's a good idea...

1. We don't yet know that this is a problem.

2. If it is a problem, it means that the test run was not cancelled at all, something we would want to fix in the framework and possibly compensate for in the driver if it occurs.

3. If we __do__ have such problems, using this approach would merely hide them from view.

What do you think?

CharliePoole:
There are still a few things open to discussion, but for purposes of the GUI this is ready to merge. Open questions we can discuss further are:

1. Creating a more performant structure for pending items than the simple list we have.

2. Suppressing further events after a forced StopRun.

One place to do that will be in the PR I create out of this code for the nunit-console project.

PR #379, opened 2019-07-24
TestCentric 1.0 Release

CharliePoole:
**Do Not Merge**

This branch will become the 1.0 production release when complete. I'm doing it "under review" in a PR so that everyone can see what's going into it and hopefully catch any errors or omissions.


CharliePoole:
Mono "latest" has suddenly started to fail the build. Since it doesn't seem to be due to any changes we have made, I changed `.travis.yml` so that particular job is allowed to fail. This has happened before and the mono folks usually correct it in a few days.

CharliePoole:
The CHANGES file now includes recognition of folks who contributed to the release. I decided that was the best place because it is matched on the website by the release notes, which will contain the same information.

The file also includes text that was previously in a separate file, describing how this GUI differs from the NUnit V2 GUI. I'll also be adding some text describing the known problem with selecting individual tests when running V2 tests.

@TestCentric/gui-team I'm pretty well done with the text files at this point, so some review would be helpful.

PR #389, opened 2019-08-12
Fix critical error in engine

### LINKED ISSUE #388: NUnit process with multiple assemblies will not load
This seems to be a regression that went out with the 1.0 release.

CharliePoole:
Fixes #388

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -103,7 +103,8 @@ Setup(context =>
 			branchName = branchName.Substring(8);
 
 		// Default based on GitVersion.yml. This gives us a tag of dev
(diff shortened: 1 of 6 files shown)


CharliePoole:
@ChrisMaddock You may want to take a look at this. I'm not sure if the problem is present in the current NUnit build of the engine. I had added code to my build so that there may be additional project subpackages wrapping the assemblies found in a project. That meant that iterating subpackages was no longer a sufficient way to find assemblies. 

IIRC I did a PR for NUnit as well, with the project level elements, but I don't know if it was merged.

PR #397, opened 2019-09-08
Fix unhandled NullReferenceException thrown when test fails with Assert.Fail()

yuyahagi:
When running the following test case, testcentric.exe v1.0.1 throws an unhandled `System.NullReferenceException`. If it is running without attaching a debugger on a Windows 10 system (with the default CLR configuration), a dialog box of "Unhandled exception has occurred..." gets displayed with the option of continuing or quitting.

```
using NUnit.Framework;

namespace demo
{
    [TestFixture]
    public class TestClass
    {
        [Test]
        public void TestMethod()
        {
            Assert.Fail();
            // Assert.Fail("") also causes the same unhandled exception.
            // Assert.Fail("Some message") does not cause the unhandled exception.
            // There may be other ways of causing the same issue but I'm not sure.
        }
    }
}
```

The test above was written for NUnit v3.12.0 and .Net Framework v4.7.1, targeting "Any CPU" platform and run on a 64-bit Windows 10 system.

The exception message is shown below.

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at TestCentric.Gui.Views.TestResultItem.ToString() in C:\Users\charlie\dev\TestCentric\testcentric-gui\src\TestCentric\testcentric.gui\Views\TestResultItem.cs:line 47
   at System.Windows.Forms.ListBox.NativeInsert(Int32 index, Object item)
   at System.Windows.Forms.ListBox.ObjectCollection.Insert(Int32 index, Object item)
   at TestCentric.Gui.Views.ErrorsAndFailuresView.InsertTestResultItem(TestResultItem item) in C:\Users\charlie\dev\TestCentric\testcentric-gui\src\TestCentric\testcentric.gui\Views\ErrorsAndFailuresView.cs:line 428
   at System.Windows.Forms.Control.InvokeMarshaledCallbackHelper(Object obj)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Windows.Forms.Control.InvokeMarshaledCallback(ThreadMethodEntry tme)
   at System.Windows.Forms.Control.InvokeMarshaledCallbacks()
```

The `NullReferenceException` turned out to be thrown because a `TestCentric.Gui.Views.TestResultItem` was instantiated with a null message and its `ToString()` method later checked for `message.Length` without a null check.

This PR adds a null check in `TestResultItem.ToString()` method along with a few tests.

Another approach might be to modify the model or presenter to ensure a `TestResultItem` always gets instantiated with a non-null message. I think it could possibly be a little cleaner (I don't like null checks, expecially in the view) but I did not take that route because `TestCentric.Gui.Model.ResultNode.GetTrimmedInnerText(XmlNode node)` explicitly returns null string under some conditions and implies that it is acceptable to have a null message.

Would love to hear if this PR aligns with the concepts and style of the project.

### DIFF
--- a/src/TestCentric/tests/Views/TestResultItemTests.cs
+++ b/src/TestCentric/tests/Views/TestResultItemTests.cs
@@ -1,0 +1,17 @@
+using NUnit.Framework;
+
+namespace TestCentric.Gui.Views
+{
+    [TestFixture()]
+    public class TestResultItemTests
+    {
+        [TestCase("message", "stack trace string")]
+        [TestCase("message", null)]
+        [TestCase(null, "stack trace string")]
+        public void ToStringTest(string message, string stackTrace)
+        {
+            var item = new TestResultItem("dummy test name", message, stackTrace);
+            Assert.IsNotNull(item.ToString());
+        }
+    }
+}
--- a/src/TestCentric/tests/Views/TestResultItemTests.cs
+++ b/src/TestCentric/tests/Views/TestResultItemTests.cs
@@ -2,7 +2,7 @@ using NUnit.Framework;
 
 namespace TestCentric.Gui.Views
 {
-    [TestFixture()]
+    [TestFixture]
     public class TestResultItemTests
     {
         [TestCase("message", "stack trace string")]
--- a/src/TestCentric/testcentric.gui/Views/TestResultItem.cs
+++ b/src/TestCentric/testcentric.gui/Views/TestResultItem.cs
@@ -44,7 +44,7 @@ namespace TestCentric.Gui.Views
 
         public override string ToString()
         {
-            if (message.Length > 64000)
+            if (message?.Length > 64000)
                 return string.Format("{0}:{1}{2}", testName, Environment.NewLine, message.Substring(0, 64000));
 
             return GetMessage();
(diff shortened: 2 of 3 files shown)


yuyahagi:
Seeing the failed Travis CI build, I'm having a trouble understanding what its error message implies. I would like to request for help for this. I noticed `[Platform(Exclude = "Linux"...)]` attributes for tests involving the TestCentric view or `AssemblyWatcher` but I'm not sure if a similar attribute should be specified for the new tests in this PR.

CharliePoole:
Thanks for doing this! I'm tied up today, but will review it tomorrow.

mikkelbu:
@yuyahagi The errors did not look related to your change, so I force triggered a build on master and on the PR branch to see if the problem was a general problem on the build server, but both builds were successful, so it must have been a temporary error.

src/TestCentric/tests/Views/TestResultItemTests.cs
```diff
@@ -0,0 +1,17 @@
+using NUnit.Framework;
+
+namespace TestCentric.Gui.Views
+{
+    [TestFixture()]
```
mikkelbu: Nit-pick: Please remove `()`

yuyahagi: Fixed in commit 66ccafd.

CharliePoole: LGTM as well. Thanks for the work @yuyahagi. Thanks for reviewing @mikkelbu. Merge anytime.

mikkelbu:
Thanks for the contribution @yuyahagi :+1:

PR #399, opened 2019-09-11
Small change to show the test assembly name in the main view's title bar.

SteveLeafo:
I was running some tests on a project here and couldn't remember which assembly I was running.

This is a small change to show the assembly name in the main view when there is only one test assembly loaded.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
@@ -121,6 +121,11 @@ namespace TestCentric.Gui.Presenters
                 }
 
                 UpdateViewCommands();
+
+                if (e.Test.Children.Count == 1)
+                {
+                    _view.SetTitleBar(e.Test.Children[0].Name);
+                }
             };
 
             _model.Events.TestsUnloading += (TestEventArgse) =>
--- a/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
@@ -122,9 +122,9 @@ namespace TestCentric.Gui.Presenters
 
                 UpdateViewCommands();
 
-                if (e.Test.Children.Count == 1)
+                if (_model.TestFiles.Count == 1)
                 {
-                    _view.SetTitleBar(e.Test.Children[0].Name);
+                    _view.SetTitleBar(_model.TestFiles.First());
                 }
             };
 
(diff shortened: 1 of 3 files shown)


CharliePoole:
Overall, I'd say this is a good idea. The private method `SetTitleBar` isn't called any longer at all. It's a holdover from the V2 GUI, when the "view" did pretty much everything itself.

I have one uncertainty about where we get the name from (see inline comment) and since I don't have time to look at it too closely for another week or so, I'm asking @mickkelbu to take a look.

src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
```diff
@@ -121,6 +121,11 @@ private void WireUpEvents()
                 }
 
                 UpdateViewCommands();
+
+                if (e.Test.Children.Count == 1)
+                {
+                    _view.SetTitleBar(e.Test.Children[0].Name);
+                }
```
CharliePoole: I think the logic of displaying the filename if there is only one is good. Not sure we should take it from `e.Test.Children`. AFAIK, in the current implementation, the top test element is always a `test-run` and the next level is always a bunch of files - either assemblies or projects. But I'm thinking that may not be the case forever, and it's certainly not guaranteed that any particular test is represented by a file. For example, what if the next element under the top could be a category in the future?

OTOH, the model maintains a list of test files, which we are unlikely to change. Shouldn't we use that list?

SteveLeafo: Yes, using the model's list of test files makes sense.

CharliePoole: It strikes me that the test on number of files is probably two much logic for the view. I think the test should be in the Presenter and the View should just display what it is given.

Sorry I didn't notice this earlier!

SteveLeafo: Hi, I kind of agree, but isn't this how it is currently implemented?

I guess if you wanted to remove logic from the view, we could move the logic in SetTitleBar() that extracts the filename from a path to the presenter.

I didn't make that change initially to keep the changes in this pull request to a minimum.

CharliePoole: In general, lots of logic has been moved from the view, but lots remains. Version 2 (currently experimental) GUI will be much cleaner. I'm OK to leave this as is for now.

CharliePoole:
Thanks for your contribution @SteveLeafo! Let us know if there's anything else you'd like to work on.

Issue #402, opened 2019-09-16
Set Project Configuration to use


CharliePoole:
NUnit V2 has a Project | Configurations submenu listing all configs in the open project. 

This is currently not available for the TestCentric GUI V1.0.

Issue #216 deals with resolving this problem for the experimental GUI. This issue aims at a fix for the standard GUI, for version 1.1.

There are two options for initial implementation.

1. A single setting that applies to all open projects. While not fully satisfactory, it would work for the common case where only one project is available. A suboption is to only allow setting the configuration where there is a single project open.

2. Allowing setting the configuration separately for each project, possibly through the properties panel or a right-click menu item in the tree.

Design work is needed to decide which option to use.

Priority is high because this is an existing feature of the V2 GUI.

CharliePoole:
Actually, after reviewing the code, it seems as if the approach of allowing the config to be set separately per project is the one most easily implemented.

If we try to set it globally, we need to deal with a number of exception cases, like a project not having the selected config or not having any configs. By tailoring the options to each project, we can avoid that, at least for now.

CharliePoole:
OK... here's an issue that took me by surprise. The engine doesn't expose the ProjectService publicly!

Options:

1. Expose it.
2. Expose a pseudo-service that only provides the names of the configs without anything else.
3. Require the user to know the names of the configs and type in the desired config.

I don't much like 3.

immeraufdemhund:
I like 1

CharliePoole:
Simply exposing the interface is non-breaking, in my view. It's currently a public interface in the `nunit.engine` assembly, which users are not supposed to reference directly. We would move it to `nunit.engine.api` making it a part of the published api. I like that because it's really simple.

We could also consider adding a method to the interface that returns the config names. Otherwise, you need to make several calls to get the names.

CharliePoole:
OK, this is getting complicated. Will create separate PRs for some of these:

- [ ] Expose IProjectService in the engine to enable the model to make use of it. Will also add the private `LoadFrom` method to the interface, to allow getting project information without needing to expand a package in the process.

- [ ] Make the engine `TestPackage` property public and allow presenters to set properties directly on a particular subpackage rather than through the current `PackageSettings` property, which affects __all__ subpackages. Provide a way to identify Package level tests (test run, project, assembly) and associate the TestNode with a particular TestPackage.

- [ ] Refactor and clean up current handling of context menus, which differs between the two GUIs.

[...]

PR #409, opened 2019-09-27
Fix to allow keeping the same visual state when changing configs

CharliePoole:
This PR is part of issue #402.

In order to __reload__ a project with changed config, it is necesary to use the same package ID for the new package. Otherwise, when visual state is restored, no matching tests are found.

This commit takes care of that problem in what is basically a hacky manner.

1. TestPackage is given a `SetID` method, so that it's possible to set the ID of a new package back to the same value as the original package.

2. A heuristic is employed to decide when to do this. If the new config for the project contains an assembly with the same file name as the original, then the  ID for the original is substituted. This works well for VS projects and other projects that use the config to represent different builds of the same assembly like Debug and Release. However, projects are not limited to using configs in this way. An NUnit project, for example, may have entirely unrelated assemblies in the various configs. In that case, the newly loaded assemblies will not have a prior visual state and will appear unexpanded.

Since this is an entirely new feature, I plan to merge this and deal with alternative approaches in a future PR.

### DIFF
--- a/src/Experimental/experimental-gui/Presenters/MainPresenter.cs
+++ b/src/Experimental/experimental-gui/Presenters/MainPresenter.cs
@@ -372,8 +372,6 @@ namespace TestCentric.Gui.Presenters
 
         private string CreateOpenFileFilter()
         {
-            const string VS_FILE_TYPES = "*.csproj,*.fsproj,*.vbproj,*.vjsproj,*.vcproj,*.sln";
-
             StringBuilder sb = new StringBuilder();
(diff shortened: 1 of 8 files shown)


CharliePoole:
@mikkelbu @immeraufdemhund This is the fifth PR I've done aganst #402. I merged the others as I went, but feel free to review the whole sequence and/or the overall working of this new feature. 

I generally feel that a lot of improvement is needed but that much of it will need to be done in the engine. I'd rather see the engine split into two pieces, as previously discussed, before doing much more feature work on it, however.

PR #410, opened 2019-09-28
Allow setting project config in experimental GUI

### LINKED ISSUE #216: Experimental GUI: Set project configuration to use
NUnit V2 has a Project | Configurations submenu listing all configs in the open project. 

That's all well and good for V2, since it only allows opening one project at a time.

NUnit 3.0 allows opening a mix of multiple projects and/or assemblies, which makes the implementation a bit more complicated. This is particularly true since the project may not have the same names for their configurations.

Since users need some control over the project configs used we will have to allow them to set the configuration for **each** project that is simultanelously open. This may require a context menu item on the tree or a dynamic menu that lists open files. Design work is needed.

In the initial release, we may be able to leave this out.

CharliePoole:
Fixes #216


CharliePoole:
Getting a strange intermittent error in the zip package test under Travis. Doesn't seem to have anything to do with current changes as it has been happening for a while. I removed that step from the Travis build.

PR #443, opened 2019-11-25
Refactor engine test hierarchy

CharliePoole:
The engine tests were inconsistently organized, some in the namespace of the class under test, some in a contained `Tests` namespace and some under `NUnit.Engine.Tests`. This PR restructures the namespaces so all tests are in the namespace of the class that they test.

In addition, some unnecessary `using` statements were removed.


CharliePoole:
I know how that is! Yeah, I was working on an actual feature in the GUI and couldn't find the right tests in the tree... got tired of that!

When you have time, let me know. I could throw you one of several pieces of work that will lead to running .NET Core under the GUI.

Issue #447, opened 2019-12-02
Engine will not recognize .NET Framework versions beyond 4.5


CharliePoole:
Recognizing 4.6, 4.7 and 4.8 is easy. However, there's an added problem in recognizing 4.6.1, 4.6.2, etc.

The RuntimeFramework class takes a version in it's constructor. When the version has only two components, it is taken to be the framework version. When it has three or four, that is taken to be a CLR version and a different path is followed.

We __could__ implement a hack, whereby small values in the third component indicate a framework version. I'd prefer to fix it so that the constructor always deals with framework version and we use some static method to construct from CLR version. Alternatively, we could just not allow use of a CLR version, which was introduced back when it was common to have pre-releases with a different CLR.

@mikkelbu @immeraufdemhund @ChrisMaddock Which way do you think I should go? I'll replicate this issue in the nunit-console repo as well.

immeraufdemhund:
I honestly have no idea. I've never cared which version I had, but our company often made us upgrade programs to the latest revision because of "security".

ChrisMaddock:
I personally think we need to separate Runtime and Target Frameworks out into two separate classes entirely. 🙂 Anything else seems like a representation which will always be open to ambiguity.

CharliePoole:
Could you clarify what you mean by Runtime vs Target Framework? What would be separated?

ChrisMaddock:
Just the representation of the two.

An example, I have a test assembly that targets .NET Framework 4.5. On my (hypothetical!) machine, I have three runtimes that I could choose to run that on: CLR v2, CLR v4 and Mono. Like wise for a test assembly that targets .NET Standard 2.0 - I could chose to run it on CLR v4, Mono or .NET Core.

I've found it confusing before how we currently use the same `RuntimeFramework` class to represent both concepts. I'd like to separate that out, and make which target frameworks which we support running on which platforms a little clearer. 🙂

CharliePoole:
Yes, I see your point. I have added a compatible implementation of `FrameworkName` to my .NET 2.0 build in order to deal better with target frameworks. One thing that `RuntimeFramework` has, which complicates things a bit, is the ability to specify "Any" framework and/or any version. So, for example, `mono` means any version of Mono and `Any` means any version of any runtime. This was probably a case of over-complicating things.

I'm going to give this some more thought.

[...]

Issue #445, opened 2019-11-25
Better error message when attempting to load .NET Core assemblies


CharliePoole:
Currently, if we point the GUI at a .NET Core or .NET Standard assembly, we get a message saying the file couldn't be found at best and in the worst case the tree node for the assembly has a blank or empty name.

Actually running .NET Core tests is where we want to go, but that's a big job. For the moment, I'd like us to detect that the assembly is non-runnable by us and say so. Ideally, we should be able to identify the actual platform requested, e.g. .NET Core 2.1.

PR #448, opened 2019-12-02
Refactoring of RuntimeFramework and RuntimeFrameworkService

CharliePoole:
This is preliminary refactoring in preparation for working on issue #445. I intended to fix the issue in one PR, but it's looking more complicated than I thought and I want to merge this now.

First commit removes all knowledge of available frameworks from `RuntimeFramework`, moving it to `RuntimeFrameworkService`

Second commit creates a compatible implementation of the `FrameworkName` class in the .NET 2.0 build of `nunit.engine.core`.

Third commit makes use of `FrameworkName` by including it as a property of `RuntimeFramework` but no further use is made of the property as yet.

@immeraufdemhund @mikkelbu I'd appreciate a quick look at this, if you can find the time. I'm going to base my fix for #447 off of this PR, which is why I need to merge it.


mikkelbu:
I can take a look at it now, but I think it would also be good to have import from @ChrisMaddock given that he has proposed something similar (and also worked with the code, see e.g. nunit/nunit-console#702 and related issues).

CharliePoole:
Thanks @mikkelbu!

@ChrisMaddock You unfortunately don't come up as a suggested name on this project so I sometimes forget you! Love to have you take a look as well.

src/TestEngine/nunit.engine.tests/Services/RuntimeFrameworkServiceTests.cs
```diff
@@ -95,24 +113,34 @@ public void EngineOptionPreferredOverImageTarget(string framework, int majorVers
...
 
             Assert.Multiple(() =>
             {
-                Assert.That(net20Package.Settings[EnginePackageSettings.RuntimeFramework], Is.EqualTo("net-2.0"));
-                Assert.That(net40Package.Settings[EnginePackageSettings.RuntimeFramework], Is.EqualTo("net-4.0"));
-                Assert.That(topLevelPackage.Settings[EnginePackageSettings.RuntimeFramework], Is.EqualTo("net-4.0"));
+                // HACK: this test will pass on a windows system with .NET 2.0 and .NET 4.0 installed or on a 
+                // linux system with a newer version of Mono with no 2.0 profile.
+                // TODO: Test should not depend on the availability of specific runtimes
+                if (platform == PlatformID.Win32NT)
```
mikkelbu: Could we decorate the test with an `PlatformAttribute` instead?

src/TestEngine/nunit.engine.tests/RuntimeFrameworkTests.cs
```diff
@@ -257,30 +230,30 @@ public override string ToString()
 
 #pragma warning disable 414
         static FrameworkData[] frameworkData = new FrameworkData[] {
-            new FrameworkData(RuntimeType.Net, new Version(1,0), new Version(1,0,3705), "net-1.0", ".NET 1.0"),
+            new FrameworkData(RuntimeType.Net, new Version(1,0), new Version(1,0,3705), "net-1.0", ".NET 1.0", ".Net Framework,Version=v1.0"),
```
mikkelbu: Should ".Net" be ".NET" and should we have a space between `,` and `Version` ? (same comment for all the elements below

src/TestEngine/nunit.engine.core/Services/ExtensionService.cs
```diff
@@ -417,22 +417,22 @@ internal void FindExtensionsInAssembly(ExtensionAssembly assembly)
...
-                        throw new NUnitEngineException($"Extension {assembly.FilePath} targets {assemblyTargetFramework.DisplayName}, which is not available.");
-                    }
-                    else
-                    {
-                        log.Info($"Assembly {assembly.FilePath} targets {assemblyTargetFramework.DisplayName}, which is not available. Assembly found via wildcard.");
-                        return;
-                    }
-                }
-#endif
+//#if !NETSTANDARD2_0
```
mikkelbu: Should we mark this with an TODO or similar?

src/TestEngine/nunit.engine.tests/Compatibility/FrameworkNameTests.cs
```diff
@@ -0,0 +1,187 @@
+// ***********************************************************************
+// Copyright (c) 2015 Charlie Poole, Rob Prouse
```
mikkelbu: Year

src/TestEngine/nunit.engine.tests/Compatibility/FrameworkNameTests.cs
```diff
@@ -0,0 +1,187 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+#if NET35
```
mikkelbu: Should we also include these tests when `NET20`?

src/TestEngine/nunit.engine.tests/Compatibility/FrameworkNameTests.cs
```diff
@@ -0,0 +1,187 @@
...
+        [TestCase("A,Profile=C")]
+        [TestCase("A,======")]
+        [TestCase("A,    =B=")]
+        [TestCase("A,1  =2=3")]
+        public static void ConstructFromStringThrowsArgumentException(string arg)
+        {
+            Assert.Throws<ArgumentException>(() => new FrameworkName(arg));
+        }
+
+        public static void ConstructFromStringThrowsArgumentNullException()
```
mikkelbu: Missing `TestAttribute`

src/TestEngine/nunit.engine/Services/RuntimeFrameworkService.cs
```diff
@@ -264,6 +299,202 @@ private static void ApplyImageData(TestPackage package)
...
+        {
+            return ((int)key.GetValue("Install", 0) == 1);
+        }
+
+        private void FindDefaultMonoFramework()
+        {
+            if (RuntimeFramework.CurrentFramework.Runtime == RuntimeType.Mono)
+                UseCurrentMonoFramework();
+            else
+            if (Environment.OSVersion.Platform == PlatformID.Win32NT)
```
mikkelbu: I prefer writing the `else if` on the same line

src/TestEngine/nunit-agent/Program.cs
```diff
@@ -92,9 +92,9 @@ public static void Main(string[] args)
             LocateAgencyProcess(agencyPid);
 
             log.Info("Agent process {0} starting", pid);
-            log.Info("Running under version {0}, {1}",
-                Environment.Version,
-                RuntimeFramework.CurrentFramework.DisplayName);
+            //log.Info("Running under version {0}, {1}",
+            //    Environment.Version,
+            //    FrameworkFinder.CurrentFramework.DisplayName);
```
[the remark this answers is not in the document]
CharliePoole: Good catch! It was a complicated refactoring, whch I made easier by creating a separate class, `FrameworkFinder` for the stuff I was removing/moving. Eventually I removed it, leaving behind this commented out code, which should now  revert back to how it was originally.

src/TestEngine/nunit.engine.tests/Compatibility/FrameworkNameTests.cs
CharliePoole: The tests target 2.0. only, I believe. I'll want to change that eventually but it depends on whether we want to continue to run in-process (which I don't).

src/TestEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -279,6 +281,10 @@ public void StopService()
 
         public void StartService()
         {
+            _runtimeService = ServiceContext.GetService<IRuntimeFrameworkService>();
```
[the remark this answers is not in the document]
CharliePoole: Our general pattern is for any service to look up other services it depends on in StartService, giving an error if that service is not available. `ServiceManager` then aborts the run if the dependency was not found. See `RecentFilesService` as another exampe of a dependency.

BTW I always had the idea that we could make service initialization order-independent by detecting such dependencies first. However, for now, we have to add services and initialize in an order that takes the dependencies into account.

src/TestEngine/nunit.engine/nunit.engine.csproj
```diff
@@ -18,6 +18,9 @@
     <PackageReference Include="System.Diagnostics.Process" Version="4.3.0" />
     <PackageReference Include="System.Xml.XPath.XmlDocument" Version="4.3.0" />
   </ItemGroup>
+  <ItemGroup>
+    <Compile Remove="Services\FrameworkFinder.cs" />
```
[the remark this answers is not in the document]
CharliePoole: That's an odd effect. It's because I first Removed framework finder using the IDE, made sure everything compiled and then deleted it. I'll edit.

src/TestEngine/nunit-agent/Program.cs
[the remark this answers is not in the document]
CharliePoole: Fixed!

src/TestEngine/nunit.engine.core/Internal/TargetFrameworkHelper.cs
```diff
@@ -23,6 +23,7 @@
 
 
 using System;
+using System.Runtime.Versioning;
```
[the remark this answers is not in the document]
CharliePoole: No longer needed since we are not using `FrameworkName` here. Could come back, however. :smiley:

src/TestEngine/nunit.engine.core/Services/ExtensionService.cs
CharliePoole: Good catch! It was temporarily commented out so I put it back substituting RuntimeFramework for the temporary FrameworkFinder class.

src/TestEngine/nunit.engine.tests/Compatibility/FrameworkNameTests.cs
CharliePoole: For now, I added NET20

CharliePoole:
Thanks to both of you. I made all the changes except for the comment about how one Service should initialize another Service on which it depends. If we want to change that, I think it has to be a global change to all Services at once. Leaving the comment unresolved in case we want to discuss more.

PR #450, opened 2019-12-08
Recognize .NET Framework versions 4.5 and higher

### LINKED ISSUE #447: Engine will not recognize .NET Framework versions beyond 4.5

CharliePoole:
Fixes #447

This PR introduces an heuristic to recognize whether an arbitrary version is in the proper form to represent a framework. The version must be of one of two forms:
 * MAJOR.MINOR
 * MAJOR.MINOR.BUILD where BUILD < 3
Using the registry, it determines what 4.x versions of the .NET Framework are available. They are listed and returned by the `IAvalableRuntimes`interface. The GUI is able to display them as available.

This PR does nothing to ensure that the proper flavor of .NET Framework 4.x is actually made available. In particular, this is not tested anywhere. That's a subject for another issue, however.

### DIFF
--- a/src/TestEngine/nunit.engine.core/RuntimeFramework.cs
+++ b/src/TestEngine/nunit.engine.core/RuntimeFramework.cs
@@ -61,10 +61,11 @@ namespace NUnit.Engine
         }
 
         /// <summary>
-        /// Construct from a runtime type, version and profile. If the version has
-        /// two parts, it is taken as a framework version. If it has three
-        /// or more, it is taken as a CLR version. In either case, the other
-        /// version is deduced based on the runtime type and provided version.
+        /// Construct from a runtime type, version and profile. The version
+        /// may be either a framework version or a CLR version. If a CLR
+        /// versionis provided, we try to deduce the framework version but
+        /// this may not always be successful, in which case a version of
+        /// 0.0 is used.
         /// </summary>
         /// <param name="runtime">The runtime type of the framework.</param>
         /// <param name="version">The version of the framework.</param>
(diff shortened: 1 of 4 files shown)


src/TestEngine/nunit.engine.core/RuntimeFramework.cs
```diff
@@ -61,22 +61,28 @@ public RuntimeFramework(RuntimeType runtime, Version version)
         }
 
         /// <summary>
-        /// Construct from a runtime type, version and profile. If the version has
-        /// two parts, it is taken as a framework version. If it has three
-        /// or more, it is taken as a CLR version. In either case, the other
-        /// version is deduced based on the runtime type and provided version.
+        /// Construct from a runtime type, version and profile. The version
+        /// may be either a framework version or a CLR version. If a CLR
+        /// versionis provided, we try to deduce the framework version but
```
immeraufdemhund: > versionis

```suggestion
        /// version is provided, we try to deduce the framework version but
```
missing space

PR #459, opened 2019-12-12
Give error message when trying to run .NET Core tests

### LINKED ISSUE #445: Better error message when attempting to load .NET Core assemblies

CharliePoole:
Fixes #445 by throwing a specific exception when attempt is made to run a test targeting .NET Core.


CharliePoole:
Travis currently failing to install dotnet SDK for several issues.

CharliePoole:
@TestCentric/gui-team I'd appreciate one of you reviewing this. I think it's an improvement but may require some tweeking later.

Currently...
 `RunTimeType` is an enum that indicates which Runtime is in use. It is used to create a `Runtime`.
 `Runtime` represents a particular Runtime, like .NET Framework or .NET Core.
 `RuntimeFramework` represents a particular version of a runtime, like .NET Core 2.1.

These names may not be the best and we can change them when we have something better.

As @ChrisMaddock has pointed out, we use `RuntimeFramework` to represent

 * An actual installed framework, including the one we are using
 * The target framework for an assembly
 * The runtime requested by the user

I think the first is different from the other two, which are actually filters that help select a runtime. I'd like to make that distinction in a future PR, in which case we would no longer need `RuntimeType.Any` or `Runtime.Any`. Instead, the caller could specify a possibly empty list of acceptable runtimes in some sort of `RuntimeFilter` class. That would eliminate a lot of complexity.

CharliePoole:
Looks like nobody has had time to review this... I'll go ahead and merge but I'll still look at any comments you may later make and fix anything that needs it.

src/TestEngine/testcentric.engine.core/Runtime.cs
```diff
@@ -0,0 +1,188 @@
...
+            // Properties are completely determined by the RuntimeType
+            switch(runtimeType)
+            {
+                case RuntimeType.Any:
+                    DisplayName = "Any";
+                    // FrameworkIdentifier is left as null
+                    break;
+                case RuntimeType.Net:
+                    DisplayName = ".NET";
+                    FrameworkIdentifier = ".NETFramework";
```
mikkelbu: Shouldn't we use `FrameworkIdentifiers.Net` here? (a similar comment to the other cases).

src/TestEngine/testcentric.engine.core/Runtime.cs
```diff
@@ -0,0 +1,188 @@
...
+        /// </summary>
+        /// <param name="name">Name of the Runtime to check</param>
+        /// <returns>True if name represents a known runtime, otherwise false</returns>
+        public static bool IsKnownRuntime(string name)
+        {
+            foreach (string item in KnownRuntimes)
+                if (item.ToLower() == name.ToLower())
+                    return true;
+
+            return false;
```
mikkelbu: This could also be written as `return KnownRuntimes.Any(r => String.Equals(r, name, StringComparison.OrdinalIgnoreCase))`

src/TestEngine/testcentric.engine.core/Runtime.cs
```diff
@@ -0,0 +1,188 @@
...
+            switch(s)
+            {
+                case ".NETFramework":
+                    return Runtime.Net;
+                case ".NETCoreApp":
+                    return Runtime.NetCore;
+                default:
+                    return null;
+            }
+        }
```
mikkelbu: A bit inconsistent that `Parse(string s)` will fail on wrong input, but `FromFrameworkIdentifier(string s)` will return `null`.

And for `FromFrameworkIdentifier(string s)` we should use the constants.

src/TestEngine/testcentric.engine.core/Runtime.cs
```diff
@@ -0,0 +1,188 @@
...
+            return left._runtimeType == right._runtimeType;
+        }
+
+        public static bool operator !=(Runtime left, Runtime right)
+        {
+            return left._runtimeType != right._runtimeType;
+        }
+    }
+
+    public static class FrameworkIdentifiers
```
mikkelbu: Currently, not used as far as I can tell.

src/TestEngine/testcentric.engine.core/RuntimeFramework.cs
```diff
@@ -112,48 +103,51 @@ private bool IsFrameworkVersion(Version v)
 
         private Version GetClrVersionForFramework(Version frameworkVersion)
         {
-            switch (Runtime)
+            if (Runtime == Runtime.Net || Runtime == Runtime.Any)
```
mikkelbu: I can see that we are just keeping the existing behaviour, but why does `Runtime.Any` set version in the same manner as `Runtime.Net`?

src/TestEngine/testcentric.engine.tests/RuntimeFrameworkTests.cs
```diff
@@ -228,33 +228,38 @@ public override string ToString()
...
-            new FrameworkData(RuntimeType.Mono, new Version(3,5), new Version(2,0,50727), "mono-3.5", "Mono 3.5", ".Net Framework,Version=3.5"),
-            new FrameworkData(RuntimeType.Mono, new Version(4,0), new Version(4,0,30319), "mono-4.0", "Mono 4.0", ".Net Framework,Version=4.0"),
-            new FrameworkData(RuntimeType.Mono, RuntimeFramework.DefaultVersion, RuntimeFramework.DefaultVersion, "mono", "Mono", null),
-            new FrameworkData(RuntimeType.Any, new Version(1,1), new Version(1,1,4322), "v1.1", "v1.1", null),
-            new FrameworkData(RuntimeType.Any, new Version(2,0), new Version(2,0,50727), "v2.0", "v2.0", null),
-            new FrameworkData(RuntimeType.Any, new Version(3,5), new Version(2,0,50727), "v3.5", "v3.5", null),
-            new FrameworkData(RuntimeType.Any, new Version(4,0), new Version(4,0,30319), "v4.0", "v4.0", null),
-            new FrameworkData(RuntimeType.Any, RuntimeFramework.DefaultVersion, RuntimeFramework.DefaultVersion, "any", "Any", null)
+            new FrameworkData(Runtime.Net, new Version(1,0), new Version(1,0,3705), "net-1.0", ".NET 1.0", ".NETFramework,Version=v1.0"),
```
mikkelbu: I think usually there is a space between `,` and `Version`

src/TestModel/tests/TestModelAssemblyTests.cs
```diff
@@ -67,6 +67,7 @@ public void CheckStateAfterLoading()
             Assert.False(_model.HasResults, "HasResults");
 
             var testRun = _model.Tests;
+            TestContext.Error.WriteLine(testRun.Xml.OuterXml);
```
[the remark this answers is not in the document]
CharliePoole: It was... already removed it in another PR but I'll have to make sure the merge doesn't put it back.

src/TestEngine/testcentric.engine.core/Runtime.cs
CharliePoole: I'm in a bit of a quandary about where to retain this info centrally. I first used `FrameworkIdentifiers` and then stopped using it, but forgot to delete the class. Seems like there needs to be a list of known framework identifiers someplace, so I guess I'll go back to using it.

src/TestEngine/testcentric.engine.core/Runtime.cs
CharliePoole: I'll go one way or the other.

src/TestEngine/testcentric.engine.core/RuntimeFramework.cs
CharliePoole: Because in the past, we have consistently assumed that "Any" would use the current runtime, which has always been .NET. Consequently, this keeps the code working but doesn't make sense. I definitely want to do more work on this.

src/TestEngine/testcentric.engine.tests/RuntimeFrameworkTests.cs
CharliePoole: It's a source of great confusion. AFAICS Microsoft's `FrameworkName` is generous on what it accepts in the constructor or to parse, but will always display `FullName`, without the space. I believe my net20 implementation does the same. Since I use this data field both as an argument and tan expected value for `FullName` I went without the space.

I'll take another look through the code (both mine and Microsoft's) to be sure it's OK. Of course, once we start using this to detect the framework to run we may discover more.

CharliePoole:
@mikkelbu Thanks for the review. I'll do a patch before releasing.

src/TestEngine/testcentric.engine.core/Runtime.cs
CharliePoole: Unfortunately, not available for the net20 build. I used conditional compilation. I don't want to maintain our own System.Linq just for this.

src/TestEngine/testcentric.engine.core/Runtime.cs
CharliePoole: Only caller was throwing on null, I removed that and let this method throw instead.

src/TestEngine/testcentric.engine.tests/RuntimeFrameworkTests.cs
CharliePoole: I verified the statement above. Spaces are ignored at either end of the string or around the comma or equal sign. See `FrameworkNameTests`. The `FullName` property never uses spaces, so that's what I'd call the canonical name.

PR #468, opened 2019-12-16
Deprecate InProcess Execution

### LINKED ISSUE #454: Deprecate in-process execution
See #444

CharliePoole:
Fixes #454

### DIFF
--- a/src/Common/testcentric.common/Options/CommandLineOptions.cs
+++ b/src/Common/testcentric.common/Options/CommandLineOptions.cs
@@ -50,7 +50,7 @@ namespace TestCentric.Gui
             this.Add("run", "Automatically run the loaded tests.",
                 v => RunAllTests = v != null);
 
-            this.Add("process=", "{PROCESS} isolation for test assemblies. Values: InProcess, Separate, Multiple. If not specified, defaults to Separate for a single assembly or Multiple for more than one.",
+            this.Add("process=", "{PROCESS} isolation for test assemblies. Values: InProcess, Separate, Multiple. If not specified, defaults to Separate for a single assembly or Multiple for more than one. Option Single is permitted but DEPRECATED and will be removed in a future release.",
                 v =>
                 {
                     if (CheckRequiredValue(v, "--process", "InProcess", "Separate", "Multiple"))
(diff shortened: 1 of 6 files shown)


src/Common/testcentric.common/Options/CommandLineOptions.cs
```diff
@@ -50,7 +50,7 @@ public CommandLineOptions(params string[] args)
             this.Add("run", "Automatically run the loaded tests.",
                 v => RunAllTests = v != null);
 
-            this.Add("process=", "{PROCESS} isolation for test assemblies. Values: InProcess, Separate, Multiple. If not specified, defaults to Separate for a single assembly or Multiple for more than one.",
+            this.Add("process=", "{PROCESS} isolation for test assemblies. Values: InProcess, Separate, Multiple. If not specified, defaults to Separate for a single assembly or Multiple for more than one. Option Single is permitted but DEPRECATED and will be removed in a future release.",
```
mikkelbu: This is confusing for a user. I would prefer "Option InProcess is permitted but DEPRECATED and will be removed in a future release."

src/Common/testcentric.common/Options/CommandLineOptions.cs
```diff
@@ -67,7 +67,7 @@ public CommandLineOptions(params string[] args)
                         MaxAgents = val;
                 });
 
-            this.Add("inprocess", "Synonym for --process.InProcess.",
+            this.Add("inprocess", "Synonym for --process.InProcess. DEPRECATED: This option will be removed in a future release.",
```
mikkelbu: Perhaps
`"Synonym for --process=InProcess...`
instead of
`"Synonym for --process.InProcess...`

src/Common/testcentric.common/Options/CommandLineOptions.cs
```diff
@@ -67,7 +67,7 @@ public CommandLineOptions(params string[] args)
...
                 v => ProcessModel = "Single");
```
mikkelbu: `v => ProcessModel = "Single");`
should be 
` v => ProcessModel = "InProcess");`

src/Common/testcentric.common/Options/CommandLineOptions.cs
CharliePoole: You're right! I forgot that we deprecated `Single` in the engine and I didn't use it at all in the GUI. I think I'll do this as a separate little PR and also remove `Single` entirely from our engine, since the enumeration is not publicly exposed.

CharliePoole:
Thanks, as noted, I'll do a separate PR and merge. I'll add an issue as well so it gets into the readme.

Issue #476, opened 2019-12-17
Eliminate use of Mono.Cecil from engine core assembly


CharliePoole:
Much of the usage of Mono.Cecil in the engine core came about simply because the package was available already. Removing it's use would be good because:

1. The core needs to be built for all execution targets and Mono.Cecil may not be available for all.
2. It would reduce the size of the core.
3. It could eliminate the problem we have with using multiple versions of Mono.Cecil by eliminating version 0.9, the last one that supported .NET 2.0.

Much of what we do with cecil in the core can be done easily via reflection. Other info can be passed on from the upper level of the engine in the TestPackage, as we now do for many things. 

We should bear in mind that we started using Mono.Cecil for examining assemblies when we might not be able to load them successfully. However, once we are done in the engine core, we have already decided to load the assembly and we may as well do so and then reflect as necessary.

I believe this change would also fix issue #475

CharliePoole:
Currently experimenting with a much broader option on the issue-476 branch. I'm reviving our old `AssemblyReader` class, refactored and renamed as `AssemblyInspector` and working to make it do the (very few) things that we currently get from `Mono.Cecil`

Here's the list of functionality we need, which is really quite small in comparison to what `Mono.Cecil` can do:

 - [ ] Examine an assembly even if we are not currently running under the runtime it targets.
 - [ ] Make sure it's a valid managed (.NET) assembly.
 - [ ] See if 32-bit execution is required.
 - [ ] Get the runtime version (CLR version) used for the assembly
 - [ ] Look at assembly references to find what test framework it uses.
 - [ ] Examine assembly attributes, in particular the TargetFrameworkAttribute.

CharliePoole:
So, in order to get assembly references, I started pulling in some files from `Mono.Cecil`, because interpreting all the metadata tables is too complex to invent when code already exists!

I ended up with my own class plus a dozen or so `Mono.Cecil' classes. But it looks like I can just add a few more from Cecil and drop most of my own. The Cecil code is already well-tested, so that's what I'm about to do. 

I'm noting this here because the existing code will disappear as soon as I do a forced push.

CharliePoole:
I've moved the code for `AssemblyView` and it's dependencies to a separate assembly. It was very confusing to have a reference to `Mono.Cecil` in the same project I'm working in to replace it!

CharliePoole:
@immeraufdemhund @ManfredLange @ChrisMaddock

[...]

PR #503, opened 2020-01-15
Create PackageSettingsService

### LINKED ISSUE #502: Create new Service to manage PackageSettings
In nunit/nunit-console#702, @ChrisMaddock proposed creation of a new `AssemblyTargetingService`, to handle the inspection and annotation of packages to indicate the target framework and to inspect potential extension assemblies for the same purpose.

I initially planned to implement this service for the TestCentric engine, but further analysis led me to broaden it's function somewhat, adding annotations for other purposes than targeting. I'd like to centralize all annotation of packages to reflect

1. Assembly Targeting
2. Policy decisions for higher level packages, e.g. what runtime to use if multiple assemblies target different runtimes.
3. Expansion of project packages

If possible, the same service should handle targeting of extension assemblies.

This will allow us to put all annotation of packages into one place within the engine, making it easier to move forward with some other issues, particularly #476

CharliePoole:
Fixes #502

This new service performs all the annotations of package settings in the testcentric.engine assembly as well as being the sole direct user of Mono.Cecil. The testcentric.engine.core assembly still uses Mono.Cecil to handle extension assemblies, which will be changed as a result of #476.


CharliePoole:
@ChrisMaddock I'd appreciate your taking a look at  this. Do you think it would be useful for nunit-console?

PR #517, opened 2020-02-03
Net45 core

CharliePoole:
This PR adds a .NET 4.5 build of the following assemblies:

 * `testcentric.engine.core.dll`
 * `testcentric.engine.metadata.dll`
 * `testcentric.engine.api.dll`
 * `testcentric.engine.core.tests.dll`

We already had a .NET 4.5 build of `testcentric.engine.dll` and its tests, so there is now a full build of the engine for .NET 4.5.

We'll add a .NET 4.5 agent eventually. At the moment, this addition is simply to make it easier to add other changes to the .NET 4.5 engine itself by building all its dependencies for .NET 4.5, which were previously only built for .NET 2.0.


CharliePoole:
Actually implemented as a NET40 Build

PR #527, opened 2020-02-20
Update packaging

### LINKED ISSUE #526: Get rid of iconUrl in our nuget package
It's deprecated in favor of icon.

A general review of the package would be good at the same time.

Looking at the same for chocolatey would also be useful.

CharliePoole:
Fixes #526


CharliePoole:
Changes to the nuget package:
 * Changed license element from a file to an SPDX license expression
 * Replaced iconUrl with icon
 * Removed deprecated summary element
 * Updated copyright date

Issue #534, opened 2020-02-25
Automate packaging and deployment


CharliePoole:
Currently, packaging and publication of new releases is done manually on a local developer machine. The goal of this issue is to automate the entire process from creation through publication and release for both pre-release (dev) builds and final releases.

Tasks:

 - [x] Reorganize build.cake into two or three scripts using #load. Generally review structure of the script. Clean up and refactor as needed.(PR #536)
 - [x] Verify package content prior to publication. (PR #538)
 - [x] Run tests of packages prior to publication. (PR #543)
 - [x] Update assemblyInfo automatically (see also #428) (PR #546)
 - [x] Push dev builds to myget when an issue is merged successfully into master. (PR #548)
 - [x] Push rc builds to myget on merge as well. (PR #549)
 - [ ] Push packages to NuGet and chocolatey for final releases.
 - [ ] Create final releases automatically on GitHub

CharliePoole:
@TestCentric/gui-team I'd appreciate some input on the second task listed above. What should we check about a package before releasing.

Possibilities... unzip, check (some or all) content, run a test.

Problem... Not sure how to run the gui itself in CI, since no display is available, so I'm currently only running batch tests.

Essentially, I'm asking what is needed beyond unit tests and package creation in order to release the packages automatically.

immeraufdemhund:
hmmm.. content can change, things can break... my first assumption is to just get FileVersionInfo of the gui and check that it's the same version... but other than that I don't know of anything that I'd do cause then it would feel like i'm testing someone else's tool

CharliePoole:
The kinds of errors I have run into are failing to include all the right files in the right directories and failing verification on chocolatey due to not meeting some requirement of theirs. So I'm not trying to test the pack command but that I gave it all the right files in the first place. OTOH, once that's correct, I suppose it's not likely to change.

Maybe I should create a small placeholder test and then add to it if the need arises. At the moment, I'm re-running a substantial portion of the unit tests, which is probably not serving any purpose except proving (indirectly) that all the right files are there.

CharliePoole:
Probably automating running the GUI __before__ we create a package would probably be more useful to us. Right now it's possible for all unit tests to pass but the GUI crashes when you start it. I check that manually, of course, but it would be nice to have it done for me even if just on the local machine.

CharliePoole:
[...]

PR #536, opened 2020-03-01
Restructuring the build.cake script

CharliePoole:
Part of issue #534. Covers the first listed task.

@mikkelbu I think this general restructuring of the script is important enough that you should review it before I merge. But if you can't get to it for a while, that's OK too - I can make any changes in the next PR just as easily.


build/parameters.cake
```diff
@@ -0,0 +1,137 @@
...
+
+		ZipPackage = new FilePath($"{PackageDirectory}{PACKAGE_NAME}-{PackageVersion}.zip");
+		NuGetPackage = new FilePath($"{PackageDirectory}{NUGET_PACKAGE_NAME}.{PackageVersion}.nupkg");
+		ChocolateyPackage = new FilePath($"{PackageDirectory}{PACKAGE_NAME}.{PackageVersion}.nupkg");
+
+		UsingXBuild = context.EnvironmentVariable("USE_XBUILD") != null;
+
+		MSBuildSettings = new MSBuildSettings {
+			Verbosity = Verbosity.Minimal,
+			ToolVersion = MSBuildToolVersion.Default,//The highest available MSBuild tool version//VS2017
```
mikkelbu: I would say that this is `VS2019`

build/parameters.cake
```diff
@@ -0,0 +1,137 @@
...
+//////////////////////////////////////////////////////////////////////
+
+public class BuildParameters
+{
+	public BuildParameters(ICakeContext context)
+	{
+		Configuration = context.Argument("configuration", "Release");
+		PackageVersion = GetPackageVersion(context);
+
+		var baseDir = context.Environment.WorkingDirectory.FullPath + "/";
```
mikkelbu: Just curious. Can Linux etc. handle paths that are using `"/"` as separator?

build/parameters.cake
```diff
@@ -0,0 +1,137 @@
...
+
+		XBuildSettings = new XBuildSettings {
+			Verbosity = Verbosity.Minimal,
+			ToolVersion = XBuildToolVersion.Default,//The highest available XBuild tool version//NET40
+			Configuration = Configuration,
+		};
+
+		RestoreSettings = new NuGetRestoreSettings();
+		// Older Mono version was not picking up the testcentric source
+		if (UsingXBuild)
```
mikkelbu: I prefer brackets when the body of an if-statement spans multiple lines

build.cake
```diff
@@ -1,167 +1,46 @@
 #tool nuget:?package=NUnit.ConsoleRunner&version=3.10.0
 #tool nuget:?package=GitVersion.CommandLine&version=5.0.0
-#addin nuget:?package=Cake.Incubator&version=5.0.1
+#tool "nuget:https://api.nuget.org/v3/index.json?package=nuget.commandline&version=5.3.1"
```
mikkelbu: What is this used for ?

build.cake
```diff
@@ -1,167 +1,46 @@
...
+	var parameters = new BuildParameters(context);
 
-			// Nuget limits "special version part" to 20 chars. Add one for the hyphen.
-			if (suffix.Length > 21)
-				suffix = suffix.Substring(0, 21);
 
-			packageVersion = gitVersion.MajorMinorPatch + suffix;
-		}
+	if (BuildSystem.IsRunningOnAppVeyor)
+			AppVeyor.UpdateBuildVersion(parameters.PackageVersion + "-" + AppVeyor.Environment.Build.Number);
```
mikkelbu:
```suggestion
		AppVeyor.UpdateBuildVersion(parameters.PackageVersion + "-" + AppVeyor.Environment.Build.Number);
```

build.cake
```diff
@@ -1,167 +1,46 @@
...
 
-		if (BuildSystem.IsRunningOnAppVeyor)
-			AppVeyor.UpdateBuildVersion(packageVersion + "-" + AppVeyor.Environment.Build.Number);
-	}
+    Information("Building {0} version {1} of TestCentric GUI.", parameters.Configuration, parameters.PackageVersion);
```
mikkelbu:
```suggestion
	Information("Building {0} version {1} of TestCentric GUI.", parameters.Configuration, parameters.PackageVersion);
```

build/parameters.cake
CharliePoole: Depends on what you use, of course. It was previously set to VS2017, which is commented out but should be removed for clarity.

build/parameters.cake
CharliePoole: The "/" __is__ the default Linux directory separator. It handles "\" as an alternate, just the opposite of Windows. I have always used "/" when doing a build script because historically many build systems (e.g. NAnt) used it as their native separator.

CharliePoole: Also 'cause you can use it in a C# string without escaping it or using '@'.

build/parameters.cake
CharliePoole: A style thing, of course, but I agree in this case. I just copied this from another location.

build.cake
CharliePoole: This forces the Cake NuGet commands to use NuGet 5.3.1 in every environment, downloading it if necessary.

CharliePoole:
It seems like the space/tabs thing happens when I edit in Visual Studio as opposed to VS Code. There's probably a setting for that.

build/parameters.cake
mikkelbu: 😄 I shouldn't review code late at night

build.cake
```diff
@@ -488,16 +254,16 @@ Task("PackageChocolatey")
...
-		CleanDirectory(ZIP_TEST_DIR);
+		CleanDirectory(parameters.ZipTestDirectory);
 
-		Unzip(File(ZipPackage), ZIP_TEST_DIR);
-		CopyTestFiles(BIN_DIR, ZIP_TEST_DIR + "bin/");
+		Unzip(parameters.ZipPackage, parameters.ZipTestDirectory);
+		CopyTestFiles(parameters.OutputDirectory, parameters.ZipTestDirectory + "bin/");
 
-		NUnit3(ZIP_TEST_DIR + "bin/" + ALL_TESTS);
+		NUnit3(parameters.ZipTestDirectory + MODEL_TESTS);
```
mikkelbu: Prior this was `ALL_TESTS`. Is this intended? 
(similar comment below)

CharliePoole: Yes. I felt I was running too many tests just to verify each package, so I picked the model tests because they are fairly limited. May want to revisit this in a later PR.

CharliePoole:
Yes, I can see that. Ideally, I suppose the movement should be done all by itself and commited before anything else happens. I'm switching to VSCode to do my editing, so let's see if that helps.

CharliePoole:
I think that `build.cake` is still pretty unwieldy but I'm not sure how to split it just yet so I'll merge this and move on to the next task.

Issue #537, opened 2020-03-05
Missing files in Release 1.3.0


CharliePoole:
The 1.3.0 release had agents for net20 and net40. Although the build itself created and tested both agents, the released packages did not include the net40 agents. As a result, the GUI fails to open assemblies compiled for the .NET 4.0 Framework or higher.

CharliePoole:
Fixed by #538

Issue #535, opened 2020-02-26
Cannot load test assembly after 1.3 update


MartinZec:
Hello,

I downloaded "testcentric-gui-1.3.0.zip" and tried the new version.
With 1.2.0 I could load the test assembly and execute tests.
With 1.3.0 I see at the Assembly Properties a 
"Test Count: 1 Should Run? No" and "NotRunnable",
as attached.

I do not see any log or other output why testcentric 1.3.0 cannot load my tests.

Do I need to change anything on the assembly?

![Testcentric_1_3_0](https://user-images.githubusercontent.com/44475413/75369048-f5441180-58c2-11ea-877b-bf9464e6bf9a.png)

CharliePoole:
What runtime are you targeting with your test assembly? Are you running with any non-default settings, like running in process for example? Can you load the same test with the experimental GUI (tc-next.exe)? If so, what does the XML tab show?

Strangely, I saw the same problem right before releasing. But I did a full rebuild and it disappeared, so I went ahead and pushed the release.

CharliePoole:
@immeraufdemhund This is the sort of thing I was concerned about in our discussion on #534. Right now, the only testing of packages we have before release is manual. Something automated, right before release, which actually exercised both GUIs in some way might be useful. In fact, a simple smoke test would have caught this one...

I'm thinking I should probably set up a bare VM and install the packages on it for manual testing until we come up with something better. That runs against my desire to make releasing really easy, so we can release more often, but getting it right has priority.

@MartinZec Have you tried either the nuget or the chocolatey packages?

MartinZec:
I used .NET 4.7.1.
tc-next.exe shows 
"Method not found: 'System.Runtime.Versioning.FrameworkName TestCentric.Engine.RuntimeFramework.get_FrameworkName()"

I used the zipped fiels, https://github.com/TestCentric/testcentric-gui/releases/download/1.3/testcentric-gui-1.3.0.zip

CharliePoole:
It appears that this is the same as issue I identified as #537. All three packages are missing the agent for running .NET 4.0 and higher tests. This will be corrected in the 1.3.1 release, which will be available some time tomorrow.

You may be able to run the tests in-process until then, but I'm not sure.

CharliePoole:
Fixed by #538

CharliePoole:
[...]

PR #538, opened 2020-03-06
Add package checking tasks to build

CharliePoole:
This is part of #534. **Ready to Merge**

~I'm ready to merge this even though the CI builds are failing, becaues they are failing in the way they are supposed to fail and pointing out a problem we already had! :smiley_cat:~
CORRECTION: The missing agents are covered by a separate issue, but it's probably better to keep master clean and fix the problem right in this same PR.

That is, the checks show that none of the three packages contain the net40 agents, which is the point of critical bug #537. I want to fix that tomorrow and do an immediate release.

Note that the way I implemented the check tasks is that they fail the build but they do __not__ stop subsequent tasks that run tests against the packages. That's because the package tests really only test the engine, not the gui. The tests themselves are run using NUnit Console. I anticipate changing this in the future, at which point the tests would depend on the checks.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -3,6 +3,7 @@
 #tool "nuget:https://api.nuget.org/v3/index.json?package=nuget.commandline&version=5.3.1"
 
 #load "./build/parameters.cake"
+#load "./build/package-checks.cake"
 #load "./build/helpers.cake"
 
 using System.Xml;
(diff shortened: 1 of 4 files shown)


CharliePoole:
This PR now fixes #537 as well. Turned out to be a trivial fix.

CharliePoole:
I believe this also fixes #535.

CharliePoole:
I'm going ahead without review since I promised the new release today. There will be subsequent PRs on the issue where we can make further changes.

PR #546, opened 2020-03-12
Automatically update AssemblyInfo files

### LINKED ISSUE #428: Update assemblyinfo automatically
Currently, the cake build uses GitVersion to determine the version for the packages but the updating of CommonAssemblyInfo.cs must be done manually. The build should update the file automatically.

CharliePoole:
Fixes #428 as well as being part of #534

### DIFF
--- a/build/parameters.cake
+++ b/build/parameters.cake
@@ -6,8 +6,22 @@ public class BuildParameters
 {
 	public BuildParameters(ICakeContext context)
 	{
-		Configuration = context.Argument("configuration", "Release");
-		PackageVersion = GetPackageVersion(context);
+		Configuration = context.Argument("configuration", DEFAULT_CONFIGURATION);
+		PackageVersion = context.Argument("packageVersion", DEFAULT_VERSION);
+		bool versionProvided = context.HasArgument("packageVersion");
+
+		var dash = PackageVersion.IndexOf('-');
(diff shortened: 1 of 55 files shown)


build/parameters.cake
```diff
@@ -6,8 +6,22 @@ public class BuildParameters
...
+		var version = dash > 0
+			? PackageVersion.Substring(0, dash)
+			: PackageVersion;
+		
+		AssemblyVersion = version + ".0";
+		AssemblyFileVersion =  version;
+		AssemblyInformationalVersion = PackageVersion;
+
+		// TODO: Make GitVersion work on Linux
+		if (!versionProvided && context.IsRunningOnWindows())
```
CharliePoole: The new test on `versionProvided` fixes a problem where any packageVersion entered at the command-line was used on Linux but ignored in favor of GitVersion on windows. If we accept an argument we should honor it. In addition to being useful in some situations, providing the argument is useful when testing the script.

build/versioning.cake
```diff
@@ -0,0 +1,45 @@
+using System.Text.RegularExpressions;
+
```
CharliePoole: After merging this, I'll probably combine some packaging-related stuff in helpers.cake into this script. Didn't do it right now so as to make review easier.

PR #548, opened 2020-03-15
Move publishing to MyGet from AppVeyor.yml to build.cake

CharliePoole:
Part of issue #534


CharliePoole:
Unfortunately, the only way to test what happens on AppVeyor after a merge is to merge! So here goes.

CharliePoole:
And it worked! Upon merge, packages were published to MyGet with the `dev` suffix.

Issue #453, opened 2019-12-09
Execute tests under .NET Core


CharliePoole:
This is the master issue for running tests under .NET Core in the 1.3 (or later) GUI. It essentially maps to the entire "Run Tests under .NET Core" project. Most of the tasks items listed will have separate issues created for them.

What we need to do...

- [x] Recognize .NET Core Test assemblies - already done under #459 

- [x] Decide which versions of .NET Core we will initially support

- [x] Determine approach to use for communication from the runner to the agents

- [x] Determine how to handle 32- vs 64- bit execution (if relevant)

- [x] .NET Core 2.1

    - [x] Build `testcentric.core.engine`

    - [x] Create agent process executable

    - [x] Create transport layer

    - [x] Modify engine to initiate correct agent

- [x] .NET Core 1.1

    - [x] Build `testcentric.core.engine`

    - [x] ~~Create agent process executable~~ Using 2.1 or 3.1 agent

    - [x] ~~Create transport layer~~ using 2.1 or 3.1 transport layer

    - [x] Modify engine to initiate correct agent

- [x] .NET Core 3.1

    - [x] ~~Build `testcentric.core.engine`~~ Using NetStandard 2.0 core

    - [x] Create agent process executable

    - [x] Create transport layer

    - [x] Modify engine to initiate correct agent

CharliePoole:
@TestCentric/gui-team @ChrisMaddock

FYI the above task list has been updated to break out each version of .NET Core I want to support and show progress.

CharliePoole:
@TestCentric/gui-team @ChrisMaddock

Going to bed now after finally managing to run .NET Core App 2.1 tests from the GUI, running under .NET Framework 4.8. There are still some bugs, but the general approach is working.

For this experiment, I used Named Pipes, but the same solution will work (possibly better) using TCP or a memory stream. Basically, it's just a binary serialization over a stream.

CharliePoole:
While I'm currently able to run .NET Core 2.1 tests under Windows, it doesn't work under Linux because of less support for named pipes. I'm pulling this issue and it's sub-issues out of the 1.3 milestone in order to get the release complete.

It looks like a TCP transport will be needed for the broadest support. However, what has been done is not wasted because the message protocol that was developed here will be used with TCP.

Since this branch has a lot of changes already, I'll be going through to see if anything ought to be merged in a separate PR before releasing.

CharliePoole:
All tasks completed!

PR #559, opened 2020-04-06
Implement execution of .NET Core 2.1 Tests

### LINKED ISSUE #508: Create agent process executable for .NET Core 2.1.
Initially, we will support running .NET Core 1.1 and 2.1 tests in addition to .NET Framework. This issue aims at creating the 2.1 agent exe, which will depend on existing dlls targeting .NET Standard 2.0.

### LINKED ISSUE #512: Create transport protocol for use with .NET Core agents.
The .NET Framework agents use remoting to communicate with the engine. This is not available under .NET Core. Instead, we'll use a TCP-based protocol, replacing existing calls and returns, as well as progress notifications, by messages.

Options under consideration...

 * Simple TCP sockets
 * Use of an open-source TCP messaging package
 * Use of a web protocol
 * ~~Named Pipes (only if we upgrade engine to run under .NET 3.5 or higher)~~

It's possible we will implement multiple approaches eventually, starting with the simplest right now.

CharliePoole:
Fixes #512
Fixes #508 

A big step toward #453 

This PR resolves multiple issues in order to implement execution of .NET Core 2.1 tests under an agent process communicating with the GUI via TCP. Some of this work was done previously in the context of an experimental implementation, which is why it is all combined here in one PR.


CharliePoole:
This is a pretty big step so I'd like to get a few reviews of  the code. It's unfortunately a fairly big change to review. I'd also love it if someone would download one of the packages from AppVeyor and exercise it. You can find the artifacts here: https://ci.appveyor.com/project/CharliePoole/testcentric-gui/builds/31974065/artifacts.

Note that only version 2.1 is supported at this point. Adding 3.1 will be pretty easy but 1.1 is harder since I won't be able to use Binary Serialization.

CharliePoole:
@TestCentric/gui-team This is ready for review. I'm confident of it because there are quite a lot of tests, but I'd like at least one review before merging.

Note that this PR does what it does. :smile: That is, I wouldn't go to a 1.4 release with it. But it's a major step ahead and I'd like to see some folks using the GUI to run .NET Core 2.1 tests and giving feedback.

What's missing for a production release:
  * .NET Core 3.0 or 3.1 (pretty easy using the same protocol)
  * .NET Core 1.1 (needs a different protocol unless we run under a higher version)
  * Reflecting options in the GUI menus (Current menu assumes you can change the runtime without checking the target runtime of every assembly)

CharliePoole:
@ChrisMaddock @jnm2 GitHub won't let me ask for your reviews through the Reviewers button, but if you have time for even a cursory examination of the structure, it would help. In particular, would such an overall approach (See Communication namespace) work for NUnit?

src/TestEngine/testcentric.engine.core.tests/Communication/Protocols/BinarySerializationProtocolTests.cs
```diff
@@ -0,0 +1,164 @@
...
+using TestCentric.Engine.Communication.Messages;
+
+namespace TestCentric.Engine.Communication.Protocols
+{
+    public class BinarySerializationProtocolTests
+    {
+        private BinarySerializationProtocol wireProtocol = new BinarySerializationProtocol();
+
+        [Test]
+        public void WriteAndReadBytes()
```
mikkelbu, line 21: It seems that this is not this testing any of our code?

src/TestEngine/testcentric.engine.core.tests/Communication/Protocols/BinarySerializationProtocolTests.cs
```diff
@@ -0,0 +1,164 @@
...
+            Assert.That(copy, Is.EqualTo(bytes));
+        }
+
+        [Test]
+        public void SerializeMessage()
+        {
+            var originalPackage = new TestPackage(new string[] { "mock-assembly.dll", "notest-assembly.dll" });
+
+            var bytes = wireProtocol.SerializeMessage(originalPackage);
+            Console.WriteLine($"Serialized {bytes.Length} bytes.");
```
mikkelbu, line 46: Should we keep the Console.WriteLine in the tests

src/TestEngine/testcentric.engine.core.tests/Communication/Protocols/BinarySerializationProtocolTests.cs
```diff
@@ -0,0 +1,164 @@
...
+            }
+
+            var received = new List<TestEngineMessage>(wireProtocol.Decode(stream.ToArray()));
+            Assert.That(received.Count, Is.EqualTo(commands.Length));
+
+            for (int i = 0; i < commands.Length; i++)
+                Assert.That(((CommandMessage)received[i]).CommandName, Is.EqualTo(commands[i]));
+        }
+
+        //[Test]
```
mikkelbu, line 129: Is this on purpose?

src/TestEngine/testcentric.engine/Services/TestAgency.cs
```diff
@@ -97,7 +107,18 @@ private ITestAgent CreateRemoteAgent(TestPackage package, int waitTime)
                 if (_agentStore.IsReady(agentId, out var agent))
                 {
                     log.Debug($"Returning new agent {agentId:B}");
-                    return new RemoteTestAgentProxy(agent, agentId);
+
+                    switch (targetRuntime.Runtime.FrameworkIdentifier)
+                    {
+                        case ".NETFramework":
```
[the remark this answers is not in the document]
CharliePoole: I started out with strings, but I agree it makes sense to collect them somewhere.

src/TestEngine/testcentric.engine.tests/testcentric.engine.tests.csproj
```diff
@@ -7,6 +7,7 @@
     <AssemblyOriginatorKeyFile>..\..\nunit.snk</AssemblyOriginatorKeyFile>
     <DebugType>Full</DebugType>
     <OutputPath>..\..\..\bin\$(Configuration)\engine-tests\</OutputPath>
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
```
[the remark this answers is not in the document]
CharliePoole: Yes, without this, the dependencies of the tests don't get copied into the directory. .NET Core can figure out how to load them but we can't load (for example) the correct nunit.framework.

This is a fundamental problem in .NET Core, which both Chris and Joseph have also been looking at.. While I don't like this approach, it gives us some forward progress for now.

We need to come back to this and maybe a separate issue is a good idea. I'd like to see what xunit does about this.

src/TestEngine/testcentric.engine.core/Communication/Messages/CommandMessage.cs
```diff
@@ -0,0 +1,141 @@
...
+        {
+            CommandName = commandName;
+            Arguments = arguments;
+        }
+
+        public string CommandName { get; }
+
+        public object[] Arguments { get; }
+    }
+
```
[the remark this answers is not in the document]
CharliePoole: Yup. An earlier implementation, which I simplified.

src/TestEngine/testcentric.engine.tests/Services/AgentProcessTests.cs
```diff
@@ -39,12 +39,10 @@ public void SetUp()
...
         [TestCase("net-2.0", false, "agents/net20/testcentric-agent.exe")]
         [TestCase("net-2.0", true, "agents/net20/testcentric-agent-x86.exe")]
-#if NETCORE_SUPPORT
         [TestCase("netcore-2.1", false, "agents/netcoreapp2.1/testcentric-agent.dll")]
         [TestCase("netcore-2.1", true, "agents/netcoreapp2.1/testcentric-agent-x86.dll")]
-        [TestCase("netcore-1.1", false, "agents/netcoreapp1.1/testcentric-agent.dll")]
-        [TestCase("netcore-1.1", true, "agents/netcoreapp1.1/testcentric-agent-x86.dll")]
-#endif
+        //[TestCase("netcore-1.1", false, "agents/netcoreapp1.1/testcentric-agent.dll")]
+        //[TestCase("netcore-1.1", true, "agents/netcoreapp1.1/testcentric-agent-x86.dll")]
```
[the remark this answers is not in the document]
CharliePoole: Leaving them for use when I get to 1.1 - if I get there that is!

src/TestEngine/testcentric.engine.core.tests/Communication/Protocols/BinarySerializationProtocolTests.cs
CharliePoole: Correct... it was an initial test to help me figure it out but can go now.

src/TestEngine/testcentric.engine.core.tests/Communication/Protocols/BinarySerializationProtocolTests.cs
CharliePoole: Nope... gone!

src/TestEngine/testcentric.engine.core.tests/Communication/Protocols/BinarySerializationProtocolTests.cs
CharliePoole: No longer needed! Deleted!

src/TestEngine/testcentric.engine.core/Communication/Protocols/BinarySerializationProtocol.cs
```diff
@@ -0,0 +1,252 @@
...
+using TestCentric.Engine.Communication.Messages;
+
+namespace TestCentric.Engine.Communication.Protocols
+{
+    /// <summary>
+    /// BinarySerializationProtocol serializes messages in the following format:
+    ///
+    ///     [Message Length (4 bytes)][Serialized Message Content]
+    ///     
+    /// The message content is in binary form as produced by the .NET BinaryFormatter.
```
mikkelbu, line 21: I was wondering whether `BinaryFormatter` was the best option. It is unsure if we ever allow users to create messages - https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2300?view=vs-2019 - also it has a couple of other drawbacks - https://stackoverflow.com/questions/703073/what-are-the-deficiencies-of-the-built-in-binaryformatter-based-net-serializati. But as long as the producer and consumer are in lockstep (meaning same platform, same .NET version ...) then we should be safe

src/TestEngine/testcentric.engine.core/Communication/Protocols/BinarySerializationProtocol.cs
```diff
@@ -0,0 +1,252 @@
...
+            while (ReadSingleMessage(messages)) { }
+
+            return messages;
+        }
+
+        public void Reset()
+        {
+            if (_receiveMemoryStream.Length > 0)
+            {
+                _receiveMemoryStream = new MemoryStream();
```
mikkelbu, line 89: This is a general comment about this file. All the `new MemoryStream` and `Array.Copy` in this class seem inefficient?

src/TestEngine/testcentric.engine.core/Communication/Transports/Tcp/TestAgentTcpTransport.cs
```diff
@@ -0,0 +1,139 @@
...
+                        break;
+                    case "CountTestCases":
+                        filter = (TestFilter)command.Arguments[0];
+                        SendResult(_runner.CountTestCases(filter));
+                        break;
+                    case "Run":
+                        filter = (TestFilter)command.Arguments[0];
+                        SendResult(_runner.Run(this, filter));
+                        break;
+#if !NETSTANDARD1_6
```
mikkelbu, line 109: The entire file is covered by `!NETSTANDARD1_6`, so this is not necessary

src/TestEngine/testcentric.engine.tests/Transports/Tcp/TcpServerTests.cs
```diff
@@ -0,0 +1,89 @@
...
+            using (TcpClient client = new TcpClient())
+            {
+                client.Connect(_server.EndPoint);
+                client.Client.Send(new Guid().ToByteArray());
+
+                Thread.Sleep(1); // Allow the connection event to run
+                Assert.That(_serverConnections.Count, Is.EqualTo(1), "Should have received 1 connection event");
+                Assert.That(_serverConnections[0].Connected, "Server is not connected to client");
+
+                Assert.True(client.Connected, "Client is not connectedToServer");
```
mikkelbu, line 48: connectedToServer => connected to server

src/TestEngine/testcentric.engine.tests/Transports/Tcp/TcpServerTests.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            using (TcpClient client = new TcpClient())
+            {
+                byte[] buffer = new byte[256];
+                new Random().NextBytes(buffer);
+
+                client.Connect(_server.EndPoint);
+                client.GetStream().Write(buffer, 0, 256);
+
+                //byte[] bytes = new byte[256];
```
mikkelbu, line 63: Comment. Also there are no asserts in this test?

src/TestEngine/testcentric.engine.tests/Transports/Tcp/TcpServerTests.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        [Test]
+        public void SingleClientConnection()
+        {
+            using (TcpClient client = new TcpClient())
+            {
+                client.Connect(_server.EndPoint);
+                client.Client.Send(new Guid().ToByteArray());
+
+                Thread.Sleep(1); // Allow the connection event to run
```
mikkelbu, line 44: Instead of Sleep we could also notify from the event and then block here until the notification.

src/TestEngine/testcentric.engine.core/Communication/Protocols/BinarySerializationProtocol.cs
```diff
@@ -0,0 +1,252 @@
...
+        /// Returns a boolean value indicates that if there is a need to re-call this method.
+        /// </returns>
+        /// <exception cref="CommunicationException">Throws CommunicationException if message is bigger than maximum allowed message length.</exception>
+        private bool ReadSingleMessage(ICollection<TestEngineMessage> messages)
+        {
+            //Go to the begining of the stream
+            _receiveMemoryStream.Position = 0;
+
+            //If stream has less than 4 bytes, that means we can not even read length of the message
+            //So, return false to wait more bytes from remore application.
```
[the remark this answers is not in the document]
CharliePoole: Made the change in my current working branch to avoid a rebase later.

src/TestEngine/testcentric.engine.core/Communication/Protocols/BinarySerializationProtocol.cs
CharliePoole: Yes, I read both of these. In the end I decided it was perfect for this (but only this) situation, where TestAgency has full control over what code runs at both ends of the comm channel. I would want to find something else for a situation between different machines, where we don't control what code is running on the other end of the line.

This is the sort of thing that led me to the conclusion that we need to be able to use various protocols and transports, not merely replace remoting with something else.

src/TestEngine/testcentric.engine.core/Communication/Protocols/BinarySerializationProtocol.cs
CharliePoole: They do at first glance, but I didn't see a lot of degradation and so kept it. I reviewed code from 3 or 4 communication packages and all the best-written ones seemed to do it this way, with the exception of those that merely transferred strings or some sort of fixed format record.

That said, I'd be all for tuning it as we see any performance issues. We could also think about some kind of benchmark using our current remoting versus tcp to do the same tests.

src/TestEngine/testcentric.engine.core/Communication/Transports/Tcp/TestAgentTcpTransport.cs
CharliePoole: Right! Left-overs.

src/TestEngine/testcentric.engine.tests/Transports/Tcp/TcpServerTests.cs
CharliePoole: Right! Removed in my current branch.

CharliePoole:
Thanks @mikkelbu I'll go through them and incorporate in the next PR as appropriate.

src/TestEngine/testcentric.engine.tests/Transports/Tcp/TcpServerTests.cs
CharliePoole: You mean something like this?

```c#
_server.ClientConnected += (c, g) =>
{
    _serverConnections.Add(c);
   // signal some event
};
```

and the block on it in the test?

It seemed like a bit much for just the test code and we have used `Sleep(1)` fairly regularly for things like this where we just want the dispatcher to give threads a chance to run. Do you think I should change this one?

mikkelbu: I think it is fine as it is. I've just worked in a place, where we had some tests that tried to do something similar to `Thread.Sleep(1)` and they failed once in while, so we did a lot of work to make them more deterministic. But this was also tests that were run more than 100 times a day in CI.

jnm2:
> In particular, would such an overall approach (See Communication namespace) work for NUnit?

Yes, and I wish I had seen this (my fault) before duplicating the work in https://github.com/nunit/nunit-console/pull/774. It looks good.

CharliePoole:
@jnm2 FWIW I've copied or @mentioned you pretty much every step of the way. :smile:

PR #633, opened 2020-12-17
Ensure BuildWebsite is called before DeployWebsite

### LINKED ISSUE #631: Website not being rebuilt for production release
The last production release (1.5.2) completed all release activities except for updating the website. The release was deployed successfully, but the call to DeployWebsite failed because BuildWebsite had never been called.

CharliePoole:
Fixes #631

### DIFF
--- a/build/website.cake
+++ b/build/website.cake
@@ -35,6 +35,7 @@ Task("UpdateWebsite")
 	{
 		if (parameters.IsProductionRelease)
 		{
+			BuildWebsite(parameters.WebDirectory);
 			DeployWebsite(parameters);
 		}
 		else


immeraufdemhund (approved):
seems simple enough. I was wondering if we could use `.DependentOn` or `.Criteria` but looks like it is more difficult. Soo... looks good to me

CharliePoole:
`IsDependentOn` doesn't work in several cases in this site. See also CreateDraftRelease and CreateProductionRelesae. That's because the thing I want to depend on is part of the BuildParameters, which are not initialized early enough. So I'd have to move IsProducionRelease into a global variable or method. But that depends on other things, which I'd also have to move. :disappointed: 

At any rate, this approach allows me to give a message about why it's being skipped.

PR #646, opened 2020-12-24
Add minimal tests to ensure dialogs don't throw when created

### LINKED ISSUE #640: Review tests of Dialogs
Issue #638 showed some problems with dialogs failing to display - throwing an exception when they were activated by the user. At a minimum, we need some tests that display each dialog without an exception. Additional testing within each dialog is highly desired as well but this may require some restructuring of the app, since dialogs don't participate in our MVP architecture when displayed. Maybe something like the Humble Dialog pattern will work for this.

CharliePoole:
fixes #640

### DIFF
--- a/src/TestCentric/tests/Dialogs/DialogCreationTests.cs
+++ b/src/TestCentric/tests/Dialogs/DialogCreationTests.cs
@@ -1,0 +1,28 @@
+using System;
+using NUnit.Framework;
+using NSubstitute;
+using TestCentric.Gui.Model;
+
+namespace TestCentric.Gui.Dialogs
+{
+    public class DialogCreationTests
+    {
+        /// <summary>
+        /// Minimal creation test for each dialog, in order to ensure
+        /// that no exception is thrown.
+        /// </summary>
+        /// <param name="dlgType"></param>
+        // TODO: We can't test TreeBasedSettingsDialog easily because
+        // it references the main presenter. This needs to be changed.
+        [TestCase(typeof(AboutBox))]
+        [TestCase(typeof(TestPropertiesDialog))]
+        [TestCase(typeof(SettingsDialogBase))]
+        [TestCase(typeof(ExtensionDialog), null)]
+        [TestCase(typeof(ParameterDialog))]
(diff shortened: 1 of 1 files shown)


src/TestCentric/tests/Dialogs/DialogCreationTests.cs
```diff
@@ -0,0 +1,28 @@
...
+        // it references the main presenter. This needs to be changed.
+        [TestCase(typeof(AboutBox))]
+        [TestCase(typeof(TestPropertiesDialog))]
+        [TestCase(typeof(SettingsDialogBase))]
+        [TestCase(typeof(ExtensionDialog), null)]
+        [TestCase(typeof(ParameterDialog))]
+        [TestCase(typeof(TestParametersDialog))]
+        public void CreateDialog(Type dlgType, params object[] args)
+        {
+            Activator.CreateInstance(dlgType, args);
```
immeraufdemhund, line 25: I read a long time ago that having test code show that no errors are thrown is not a test. I'm just wondering if a slightly more verbose test wouldn't be in order. Like that no error is thrown or that the object isn't null?

CharliePoole: I agree with that and it's definitely the received wisdom. I've even preached it myself. That's why there's a comment about it's being minimal. :smile: 

I could make it more formally a test by surrounding the creation with `Assert.DoesNotThrow` but I tried it both ways and I think the output from this one is easier to understand.

I could test for non-null, but I believe that's impossible unless you instantiate a `Nullable<T>` with `Activator`.

So bottom line... the errors you found would have been caught by these tests, so I have to say they __are__ real tests!

Future direction: make a separate test for each dialog, do the create in a OneTimeSetUp method and verify other stuff about the created dialog.

Further future: split each into two class using the Humble Dialog pattern. That will allow separate testing of the controlling class.

CharliePoole: @immeraufdemhund But feel free to improve this after I merge!!! :smile_cat:

Issue #656, opened 2021-01-11
Pluggable Agents


CharliePoole:
Currently, the GUI engine includes the following agents:
* .NET Framework 2.0 (default for tests targeting .NET 2.0 through 3.5)
* .NET Framework 4.0 (default for tests targeting .NET 4.0 through 4.8)
* .NET Core 2.1 (default for tests targeting .NET Core 1.x or 2.x)
* .NET Core 3.1 (default for tests targeting .NET Core 3.x)
* .NET 5.0 (default for tests targeting .NET 5.0)

In addition to the defaults, the File menu may be used  to run tests under any compatible agent. For example, you may use the .NET 5.0 engine to run tests targeting any .NET Core version.

All of these agents are built-in. That is, they are included as part of the GUI package and the code for deciding which one to use is built into the TestAgency service.

A pluggable structure would simplify adding (and removing) agents and would allow users to create special-purpose agents of there own.

The general idea is to make each agent an extension, at least in form. A few agents would still be built-in, but would work like extensions internally. This is what we already do, for example, with the NUnit 3 framework driver. It's form is that of an extension, but all the code resides in the core assembly.

A pluggable agent would be an extension implementing a new interface. The interface would need to be published but might not have to be part of the nunit.engine.api assembly in order to avoid compatibility issues.

The interface would allow an agent to be passed a test package and decide whether it was able to run that package. All the current centralized code to decide what agent should be used would be replaced by a loop querying the available agents.

The above is merely the outline of a design... more work will be needed to firm it up.

This is a breaking change, so I'm assigning it to the next major release.

CharliePoole:
@ChrisMaddock This is similar, I think, to what we have talked about for the NUnit engine.

CharliePoole:
NOTE: I marked this breaking on the assumption that we would stop bundling some agents as a result of implementing this feature. If we continue to bundle all agents, then it's non-breaking.

CharliePoole:
With the merging of PR #668, the code has now been reorganized so that each agent is independent and implements the `IAgentLauncher` interface. One result is that the agent launching code has a lot of repetition, which __could__ be factored out to a base class. However, I haven't done that because at least some of these will eventually be packaged separately and having them inherit implementation from the engine itself would make that more difficult.

[...]

Issue #664, opened 2021-01-16
Replace AppDomain and Process with "Agent" from user perspective


CharliePoole:
We have a File menu item which allows the user to control how AppDomains are used and another that controls how Processes are used. The first is being removed and the second is to be simplified by dropping InProcess execution.

This issue is a proposal to take it a step farther, replacing the ProcessModel menu with some sort of an Agent selection menu, listing available agents. Depending on the implementation, we might use a separate selection dialog rather than simple sub-menus.

In the current implementation, once in-process execution is dropped, agents will obviously still be processes. However, the concept of an "Agent" is general enough to cover execution on a server or on a connected device such as a phone. Agents would have some set of defined properties but would also include a text description to help the user make a choice. Some agents might be launched by the GUI, just as Agent processes are now launched. But others might be running separately and registered to accept requests from the GUI.

This idea is related to issue #656 (Pluggable Agents) but either could be done first.

If done, this would be a breaking change, so it's in the 2.0 milestone.

CharliePoole:
Moving ahead with this as a feature... Currently, it's the last feature needed for release 2.0 although more stuff may come up as I work on it.

Current flow: `MasterTestRunner` makes use of various services to expand the `TestPackage` and get a real, internal runner. Currently that all flows down to a `ProcessRunner`, which then uses `TestAgency` to launch an agent as a separate process.

New flow: `MasterTestRunner` should use the same info to get an agent directly. The agent will then launch a process and return a runner proxy for `MasterTestRunner` to use. Much simpler.

Note: The term "agent" is currently used with two meanings and will continue to be so. One is the actual agent process, usually running `testcentric-agent.exe`. The other is the local agent proxy, which stands in for the agent process and runner and is referenced by the `MasterTestRunner`. Usually the context makes it clear which one we mean but when there's doubt, we'll refer to the agent process versus the agent proxy.

The first step in implementing this change will be to modify `MasterTestRunner` to always use an agent, without knowing that the agent will create a process. This is important because future agents may work in different ways, such as by connecting to a server running on the network.

CharliePoole:
This is done to the extent needed for 2.0. That is, like the Runtime Selection it replaces, it operates as a global override for all assemblies included in the top-level TestPackage. For that reason, if the package contains incompatible assemblies (i.e. both .NET Framework and .NET Core) then the menu is not enabled, In addition, it isn't enabled unless there is a choice to make... i.e. more than one agent is available.

[...]

PR #668, opened 2021-01-27
Separate launchers in preparation for creating extensions

CharliePoole:
Partial fix to #656 

This PR isolates the launcher for each agent into a separate file. This is a cleaner organization, albeit leading to some duplication, and a preliminary step to making some or all of the launchers into extensions.


CharliePoole:
@ChrisMaddock FYI This is the path I imagine for the NUnit Engine to move to pluggable agents as well.

src/TestEngine/testcentric.engine/Services/AgentLaunchers/IAgentLauncher.cs
```diff
@@ -0,0 +1,19 @@
...
+using System;
+using System.Diagnostics;
+using NUnit.Engine;
+
+namespace TestCentric.Engine.Services
+{
+    public interface IAgentLauncher
+    {
+        bool CanCreateProcess(TestPackage package);
+        Process CreateProcess(Guid agentId, TestAgency agency, TestPackage package);
```
ChrisMaddock, line 16: Suggestion: how about CreateAgent/CanCreateAgent, and an abstraction over Process?

Is it a requirement that an agent is in a separate process, or could I create an in-process-agent if I wanted to? (Could that potentially also be the answer to one of our other issues, as well?) Also thinking of possible future platforms where multiple processes may not be available?

I don't think this matters particularly for this PR, but would be a consideration for me if we were going to make this a public extensibility interface in the NUnit Engine. Up to you if think it's worth considering, or can be left to a future concern!

CharliePoole: We're thinking alike. See #664 :smile: 

Both that issue and this one are in my 2.0 milestone.

PR #837, opened 2021-10-25
chore: Use Mono.Options rather than copied source code

### LINKED ISSUE #836: Convert command-line processing to latest version of Mono.Options
We are currently using a slightly modified copy of the source code for an older version. Better to use the available package and modify our own code to work with it as needed.

mikkelbu:
Fixes #836


CharliePoole:
@mikkelbu I'll take a shot at using delayed signing here

CharliePoole:
The assembly wasn't built for delayed signing as I thought, so that option is out. What's the downside of eliminating strong naming for the gui itself as opposed to the reusable assemblies?

mikkelbu:
> The assembly wasn't built for delayed signing as I thought, so that option is out. What's the downside of eliminating strong naming for the gui itself as opposed to the reusable assemblies?

I cannot answer this without researching more about strong naming, as I only know the high-level concepts.

CharliePoole:
I've done some reading and thinking. I think the issue is that even if strong names are no longer useful outside of the .NET Framework, as some have claimed, we still support netfx and our users may want to use some things from their own stongly named assemblies. In fact, that's how we ended up using strongly named assemblies ourselves... users wanted to reference them that way even back in NUnit V2 days.

So... here's what I'm thinking... 

I've observed that the error doesn't come up until some method is actually called rather than when we restore or build. Maybe  we could ensure that all the calls to mono.options happen in the exe. In fact, that may already be the case. I think it's fine for the exe to not be strong named.

If that doesn't work, we could consider making the gui assembly itself non-strong-named while leaving the model, which is intended to be shared, strong-named.

This is slightly more work than I imagined at first, but probably not a lot. Do you feel like giving it a shot? Simplest first step would be to just stop signing the exe and see what happens.

CharliePoole:
Replaced by PR #845

PR #910, opened 2022-07-14
Patch 1

MudassirNoor:
Minor updates to README.md and VERSION.md for better readability :).

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -1,16 +1,16 @@
 # TestCentric Runner Versioning
 
-**TestCentric** packages use a version number in the form MAJOR.MINOR.PATCH. We change the
+**TestCentric** packages use a version numbering system in the form of `MAJOR.MINOR.PATCH`. We change the:
 
-- MAJOR version when we have made incompatible API changes
-- MINOR version when adding functionality in a backwards-compatible manner
-- PATCH version when you making backwards-compatible bug fixes
+- `MAJOR` version when making incompatible API changes.
+- `MINOR` version when adding functionality in a backwards-compatible manner.
+- `PATCH` version when making backwards-compatible bug fixes.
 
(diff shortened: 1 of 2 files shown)


VERSIONING.md
```diff
@@ -63,4 +63,4 @@ The command-line API is defined by the documentation for the GUI, which includes
 
 Even in the case of non-breaking changes, as defined above, every effort will be made to avoid negative impact on users.
 
-The included **Experimental GUI** is not subject to the above considerations. It is currently versioned separately from the standard GUI, with a MAJOR component of 0. The **Experimental GUI** may gain and lose features at any time until it becomes (as we eventually intend) the second major version of the GUI.
```
CharliePoole, line 66: The Experimental GUI no longer exists, so let's drop this paragraph.

VERSIONING.md
```diff
@@ -1,16 +1,16 @@
...
-- PATCH version when you making backwards-compatible bug fixes
+- `MAJOR` version when making incompatible API changes.
+- `MINOR` version when adding functionality in a backwards-compatible manner.
+- `PATCH` version when making backwards-compatible bug fixes.
 
 What remains is to define the API or APIs we are protecting from incompatible changes and the degree of change that is needed before something is considered a breaking change.
 
 ## TestCentric APIs
 
-First, let's note that the **TestCentric GUI** is not designed to be used as a library, so there is no ABI or library-like API to worry about. So what other APIs do we need to support?
```
CharliePoole, line 13: ABI is correct... Application Binary Interface as opposed to a (source-level) API. That's why both are mentioned.

CharliePoole (changes requested):
Thanks for doing this. I think it generally improves the text. See two comments for changes I'd like to see.

VERSIONING.md
MudassirNoor: My bad, I will revert that change.

CharliePoole (approved):
Looks good. Thanks for your work.

PR #1089, opened 2024-08-16
Improve robustness of ProgressBar for "no test cases are selected" (Solves Issue 1087)

### LINKED ISSUE #1087: Toolbar: "Run selected tests" => System.DivideByZeroException
**Description:**
I observe an exception while starting a test run using the toolbar button 'Run selected tests'. The error message states that there's a DivideByZeroException.
I'm using **TestCentric-Gui 2.0.0-beta5** and the test assembly is targeting **.Net Framwork 4.8.**

This error shows up reliable if I execute this operation as the first test action after loading a project. It also shows up when running all tests first (successfully) and then start "Run selected tests" afterwards. Moreover it doesn't seems to make a difference if I select one test or the top node in the test tree view.

A close look to the screenshot and the error message might indicate that the ProgressBar fails in this scenario.
grafik

**Steps to reproduce the behavior:**
1. Start TestCentric GUI
2. Create or load project containing some tests
3. Start 'Run selected tests' using the toolbar
=> error occurs

**Environment:**
- OS: Windows 11

rowo360:
Fixes #1087 

This PR closes issue #1087 by improving the robustness of the ProgressBar class and thereby avoiding the DivideByZeroException. So overall it's considered as a simple and low risk solution to fix this issue without changing any user workflow.

The user output in this case is: 
Please note the Test count:0 in the screenshot and the empty ProgressBar

Alternativate solutions:
- Don't start 'Run selected tests' workflow, if no tests are selected (maybe with some message box)
- Disable toolbar button 'Run selected tests', if no tests are selected

The additional observation mentioned in this issue about the Settings -> Tree Display -> Display a checkbox is not handled by this PR, but I propose it should handled by a separate issue which defines the target state.

### DIFF
--- a/src/TestCentric/tests/Controls/TestCentricProgressBarTests.cs
+++ b/src/TestCentric/tests/Controls/TestCentricProgressBarTests.cs
@@ -1,0 +1,70 @@
+// ***********************************************************************
+// Copyright (c) Charlie Poole and TestCentric contributors.
+// Licensed under the MIT License. See LICENSE file in root directory.
+// ***********************************************************************
+
+using System.Drawing;
+using NUnit.Framework;
+
+namespace TestCentric.Gui.Controls
+{
+    [TestFixture]
+    public class TestCentricProgressBarTests
+    {
+        [Test]
+        [TestCase(0, 0)]
+        [TestCase(50, 100)]
+        [TestCase(100, 200)]
+        public void GetProgressWidth_MaximumValueIsDefined_ReturnsExpectedValue(int progress, int expectedWidth)
+        {
+            TestCentricProgressBar progressBar = new TestCentricProgressBar();
+            progressBar.Maximum = 100;
+            progressBar.Value = progress;
+
+            Rectangle rect = new Rectangle(0, 0, 200, 50);
+            int width = progressBar.GetProgressWidth(rect);
+
+            Assert.That(width, Is.EqualTo(expectedWidth));
+        }
+
+        [Test]
+        public void GetProgressWidth_MaximumValueIsZero_ReturnsZero()
+        {
+            TestCentricProgressBar progressBar = new TestCentricProgressBar();
+            progressBar.Maximum = 0;
+            progressBar.Value = 0;
+
+            Rectangle rect = new Rectangle(0, 0, 200, 50);
+            int width = progressBar.GetProgressWidth(rect);
+
+            Assert.That(width, Is.EqualTo(0));
+        }
+
+        [Test]
+        [TestCase(0, 0)]
+        [TestCase(50, 50)]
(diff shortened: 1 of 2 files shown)


CharliePoole (approved):
Seems like quite a simple change (best kind) so I'm ready to merge it.

I do think it would be better to have a unit test that fails without this change and passes with it. It's obvious that the fix will work but somebody could take it out in the future and I'd like some test to fail if that happens. Can you take a look at the tests and see if that's easily done? It's probably not worth more than a few hours effort, if that.

FWIW, when this code was originally written (NUnit V2 GUI), I tried to ensure that OnPaint would not be called in this situation. But platforms have changed. :-)

CharliePoole:
@rowo360 Let me know if you want me to merge as is or if you plan to update this further.

rowo360:
I also think that a few tests are helpful here. Therefore I added a new commit to this PR. 
However I need to change my original solution to test something at all. The challenge was that it is a Windows Forms Control and the changes were located in the protected OnPaint() method. It's possible to execute this method from a test, but it's difficult to assert something meaningful - except that there's no exception. The method uses concrete classes from the PaintEventArgs, but there are no interfaces around, which can be easily assert for received calls.

So overall I adapted my solution slighty by extracting two methods which take care about the Maximum=0 use case and which are testable. From customer point of view there's no difference - here's the screenshot:
TestOutput

src/TestCentric/tests/Controls/TestCentricProgressBarTests.cs
```diff
@@ -0,0 +1,70 @@
...
+
+using System.Drawing;
+using NUnit.Framework;
+
+namespace TestCentric.Gui.Controls
+{
+    [TestFixture]
+    public class TestCentricProgressBarTests
+    {
+        [Test]
```
CharliePoole: Removing [Test] as a matter of style. NUnit ignores it when [TestCase] is used.

src/TestCentric/tests/Controls/TestCentricProgressBarTests.cs
```diff
@@ -0,0 +1,70 @@
...
+            progressBar.Maximum = 0;
+            progressBar.Value = 0;
+
+            Rectangle rect = new Rectangle(0, 0, 200, 50);
+            int width = progressBar.GetProgressWidth(rect);
+
+            Assert.That(width, Is.EqualTo(0));
+        }
+
+        [Test]
```
CharliePoole: As above...

CharliePoole (approved):
That looks great. I made a change directly eliminating the two redundant [Test] annotations. I think the auxiliary methods should probably be internal but I'll do that separately since we don't have `<InternalsVisibleTo>` currently in the project file.

PR #1091, opened 2024-08-19
Improve "Show checkboxes in tree" functionality

### LINKED ISSUE #1090: Show checkboxes in tree functionality not 100% consistent
**Description:**
At least for me the 'Show checkboxes in tree' functionality is sometimes confusing or is not 100% consistent in all cases. Therefore I raise this issue as I've the impression that it doesn't work as intended. But please correct me, if I miss anything.
I'm using **TestCentric-Gui 2.0.0-beta5**.

**Steps to reproduce:**
Problem A:
1. Create a new project
2. Open settings and activate "Show checkboxes" (Tools -> Settings -> Tree Display)
    => No checkboxes are shown
4. Create another project => still no checkboxes are shown
5. Restart the application and create another project, the checkboxes are shown suddenly.
(Note: it's important to use really new test dlls here - which doesn't already have a VisualState file)

Problem B:
1. Create two projects 
2. Activate "Show checkboxes" via tree context menu in one project and keep it deactivated in the other one
3. Load both projects one after the other
4. Open context menu
   => The state of the context menu entry "Show checkboxes", doesn't match to the actual state of the tree.

ContextMenuIsssue

**Show checkbox functionality in general (my interpretation):**
There are two options for the customer to activate "Show checkboxes in tree". Although these options sounds similar their intention is slightly different: 
- In the settings: this option is about to set the default visibility of the checkboxes for newly created projects. It's stored in the application settings. Customers uses this option, if they want to show checkboxes in the tree in general.
- In the context menu of the tree: this option is about to set the visibilty of the checkboxes in the current project. It's stored individually in the VisualState file for each test dll. Customers uses this option, if they want to show checkboxes for individual projects only.

**Expected behavior:**
- If "Show checkboxes" is changed in the settings, it's applied to the test tree immediately.
- If a test is loaded, the checkboxes are shown according to the VisualState file (if present) or the default value from the setting is applied.
  (Moreover the state of the context menu entry 'Show checkboxes" is consistent to the tree)

**Environment:**
- OS: Windows 11

rowo360:
This PR closes #1090 by applying the "Show checkboxes" settings in these use cases:

- If "Show checkboxes" is changed in the settings, it's applied to the test tree immediately.
- If a test is loaded, the checkboxes are shown according to the VisualState file (if present) or the default value from the setting is applied.
  (Moreover the state of the context menu entry 'Show checkboxes" is consistent to the tree)

These changes are intended to provide a more consistent user experience when working with "Show checkboxes" functionality.
But please check first, if you agree with this proposed solution from customer point of view.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/TreeViewPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/TreeViewPresenter.cs
@@ -55,9 +55,8 @@ namespace TestCentric.Gui.Presenters
             {
                 EnsureNonRunnableFilesAreVisible(ea.Test);
 
-                VisualState visualState;
-
-                if (TryLoadVisualState(out visualState))
+                bool visualStateLoaded = TryLoadVisualState(out VisualState visualState);
+                if (visualStateLoaded)
                 {
                     switch (visualState.DisplayStrategy)
                     {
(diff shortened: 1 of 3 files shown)


CharliePoole (approved):
Looks good! Thanks!

PR #1095, opened 2024-08-21
Hide ResultSummary view when closing/reloading tests; Fix cross thread access for closing XML/TestProperties view

### LINKED ISSUE #1092: Close project => Result summary view remains open.
**Description:**
When closing a project all test related views gets closed (for example TestPropertyView). But the ResultSummaryView remains open, still showing the results from the last test run. The user cannot hide the view himself because the button 'show/hide result' is disabled as no test is loaded.
This behavior is observed by using both view modes: **Full-GUI and Mini-GUI.**

grafik

The same behavior can also be observed when creating a new project or loading another project (instead of closing) - also in these cases the result summary view remains open showing the results from a different project.

**Reproduction:**
1. Load project and run tests
2. Show result summary view
3. Close project or load another project or new project
=> Result summary view remains open

**Expected behavior:**
I assume that the ResultSummary View should be hidden in these cases: close project, load project and create project.
Otherwise, I think it is confusing for the user when test results for previous tests are displayed.

**Environment:**
- OS: Windows 11

### LINKED ISSUE #1093: Internal error message box: Mini GUI => ShowProperties => Run tests of selected node
**Description:**
An internal error message box pops up in the described use case when working with the Mini-GUI and showing the property view.

**Reproduction:**
1. Switch to Mini-GUI view mode
2. Load project and select one test node
3. Use context menu to show test properties view
4. Click toolbar button 'Run all tests' or 'Run selected tests'
 => Error message box is shown

The same behavior can be observed with the 'View as XML" dialog.

**Additional information:**
Ultimately, this message box is displayed because there is a cross-thread operation resulting in an InvalidOperationException.

**Environment:**
- OS: Windows 11

rowo360:
This PR closes #1092 and closes #1093.
I hope it's OK to combine both fixes in one PR as these changes are small.

**Issue 'ResultSummary view':**
The proposed changes take care that the ResultSummary view gets hidden, when tests are unloading and when tests are reloading. I was wondering whether event Unloading/Reloading or event Unloaded/Reloaded are better for this task, but came to the conclusion that for this use case it's not a essential difference. So I stick with the event Unloading/Reloading.

**Issue cross-thread access:**
The solution uses the existing extension method InvokeIfRequired to solve the cross-thread access.
I have given some thought to whether unit tests can be written here, but I failed to find any solution. I assume that some UI thread must be running during test execution to reproduce this issue in a test run, that seems challenging on first sight. If there is already a sample solution somewhere for this use case, then I can also apply it here.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
@@ -130,6 +130,7 @@ namespace TestCentric.Gui.Presenters
             {
                 UpdateViewCommands();
 
+                _view.RunSummaryDisplay.Hide();
                 BeginLongRunningOperation("Unloading...");
             };
 
@@ -144,6 +145,7 @@ namespace TestCentric.Gui.Presenters
             {
                 UpdateViewCommands();
 
+                _view.RunSummaryDisplay.Hide();
                 BeginLongRunningOperation("Reloading...");
             };
 
--- a/src/TestCentric/testcentric.gui/Presenters/TreeViewPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/TreeViewPresenter.cs
@@ -16,6 +16,7 @@ namespace TestCentric.Gui.Presenters
     using System.Xml;
     using System.Drawing;
     using System.IO;
+    using TestCentric.Gui.Controls;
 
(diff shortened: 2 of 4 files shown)


CharliePoole (commented):
Thanks... very thorough... I agree there's no need for a test to avoid cross-thread access.

PR #1099, opened 2024-08-26
Close property window when switching from Mini GUI to full GUI view mode

### LINKED ISSUE #1098: Switch from Mini to Full GUI mode => Properties Dialog remains open
**Describe the bug**
The 'Properties' dialog is intended for the Mini-GUI view mode to display all details of a test. However this dialog remains open, if the user switches from Mini-GUI to Full-GUI mode. Especially if this dialog is pinned, it might remain open for a while.
Fortunately, I didn't notice any problems of this dialog in the Full-GUI mode, but actually it's not the intention to show the 'Properties' in a separate dialog because the properties are already displayed in the right panel.

**To Reproduce**
1. Switch to Mini-GUI view mode
2. Open context menu on one tree node => select "Properties..."
    => Properties dialog will be displayed
3. Switch to Full-GUI view mode
  => Properties dialog is still open

grafik

**Expected behavior**
I propose that the Properties dialog is getting automatically closed in this use case. Just to simplify the workflows and to avoid any unexpected use cases in the future.

**Environment:**

- TestCentric Version 2.0.0-beta5 and main developer branch
- OS: Windows 11

rowo360:
This PR closes #1098 by closing the Properties dialog when switching from Mini GUI to Full GUI mode.

This fix doesn't solve any critical existing problems, but is instead intended to simplify the use cases and to avoid any unexpected handling in the future. Of course, we can discuss whether the dialog is also helpful in Full GUI mode. But since it cannot actually be opened using the Full GUI mode, I would not support it in this way either.

This PR includes also a second improvement:
All toolbar buttons provide a tooltip - all buttons except one! This missing tooltip was added now.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/TreeViewPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/TreeViewPresenter.cs
@@ -133,6 +133,10 @@ namespace TestCentric.Gui.Presenters
                     case "TestCentric.Gui.TestTree.ShowCheckBoxes":
                         _view.ShowCheckBoxes.Checked = _treeSettings.ShowCheckBoxes;
                         break;
+                    case "TestCentric.Gui.GuiLayout":
+                        if (_model.Settings.Gui.GuiLayout == "Full")
+                            ClosePropertiesDisplay();
+                        break;
                 }
             };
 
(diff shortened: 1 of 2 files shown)


CharliePoole (approved):
Looks good to me. Thank you.

PR #1100, opened 2024-08-28
Reset all tree node images to the initial state before starting a new test run

### LINKED ISSUE #1094: Run selected tests => Test tree icon state: reset or remain?
**Description:**
I notice that the tree icons are handled differently in **TestCentric 2 beta** than in **TestCentric 1.6.4** when executing a set of selected tests one after the other. It's only observable when executing 'Run selected tests' not when executing 'Run all tests'.
I cannot judge exactly whether this is better or worse. But at least a conscious decision should be made, if to keep the current behavior or switch back to the old one.

TestCentric 1.6.4: All tree icons are reset to the default state before running any tests
TestCentric 2: The tree icons are not reset and therefore the icons represent an accumulate state from all previous test runs

**Reproduction:**
1. Load test project
2. Execute 'Run selected tests' for one test fixture
3. Execute 'Run selected tests' for a different test fixture
    => Observe resulting tree icons

TreeIconState

**Additional note:**
The Visual Studio test runner doesn't reset the icon state in this use case. Instead, it displays the icons slightly faded out in this case (see screenshot). That's also a nice solution! It might be an idea for some future improvement.

VisualStudioTestTree

**Environment:**
- TestCentric Version 2.0.0-beta5 and main developer branch
- OS: Windows 11

rowo360:
This PR closes #1094.

All tree node images are reset to the default state before starting a new test run. The behavior is now identical to the previous TestCentric 1.6.4 version.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,7 +2,7 @@ image: Visual Studio 2022
 
 build_script: 
   - cmd: dotnet --info
-  - ps: .\build.ps1 --Target=AppVeyor --Configuration=Release
+  - ps: .\build.ps1 --Target=ContinuousIntegration --Configuration=Release
   
 # disable built-in tests.
 test: false
--- a/build.cake
+++ b/build.cake
@@ -4,7 +4,7 @@
 const string REF_ENGINE_VERSION = "2.0.0-dev00023";
 
 // Load the recipe
-#load nuget:?package=TestCentric.Cake.Recipe&version=1.2.1-dev00010
+#load nuget:?package=TestCentric.Cake.Recipe&version=1.3.1
 // Comment out above line and uncomment below for local tests of recipe changes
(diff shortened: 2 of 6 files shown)


rowo360:
Yes, that's true. So far I didn't run that script locally. 
But I'll get familiar with this one now.

rowo360:
I actually wanted to report that I was successful, but now I noticed that the build is failing again. :-(
I think I've actually managed it quite well (getting familiar with cake build): running the script locally results into the same error message and when updating to TestCentric.Cake.Recipe version 1.3.1 it was successfully. So far, so good :-)

However I noticed now, that running the script with target (--Target=AppVeyor) results into the same error message as on the server ("Error: Reflexive edges in graph are not allowed."). Do you have an idea what's missing here?

PR #1103, opened 2024-09-05
Refactor TreeViewPresenter - introduce interface ITreeDisplayStrategy

rowo360:
This PR is about a refactoring task we discussed about in #1102: 
I introduced a new interface _ITreeDisplayStrategy_ along with a new interface _ITreeDisplayStrategyFactory_. The goal of this task is mainly to improve the testability of the TreeViewPresenter class.

The new interface _ITreeDisplayStrategy_ provides exactly that set of functions, which were currently used by the TreeViewPresenter class. Moreover I thought it's benefical to move also the creation of the concrete classes out of the TreeViewPresenter class , so I introduced the new factory class _TreeDisplayStrategyFactory_. I decided to pass in the factory class via the constructor - but I can change it to some other pattern if required.

Regarding tests:
Some existing tests require a concrete TreeDisplayStrategy instead of a Substitute (because the concrete Strategy fires an event), so I adapted those tests. The test class WhenTestRunBegins is a good example of the improvement - some lines of code could be deleted here.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/ITreeDisplayStrategyFactory.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/ITreeDisplayStrategyFactory.cs
@@ -1,0 +1,25 @@
+// ***********************************************************************
+// Copyright (c) Charlie Poole and TestCentric contributors.
+// Licensed under the MIT License. See LICENSE file in root directory.
+// ***********************************************************************
+
+using TestCentric.Gui.Model;
+using TestCentric.Gui.Views;
+
+namespace TestCentric.Gui.Presenters
+{
+    /// <summary>
+    /// This factory interface supports the creation of TreeDisplayStrategy classes
+    /// </summary>
+    public interface ITreeDisplayStrategyFactory
+    {
+        /// <summary>
+        /// Creates a concrete class implementing interface ITreeDisplayStrategy
+        /// Supported displayStrategy:
+        /// - "NUNIT_TREE" (also default case) 
+        /// - "FIXTURE_LIST"
+        /// - "TEST_LIST"
+        /// </summary>
+        ITreeDisplayStrategy Create(string displayStrategy, ITestTreeView treeView, ITestModel testModel);
+    }
+}
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-image: Visual Studio 2022
+image: Previous Visual Studio 2022
 
 build_script: 
   - cmd: dotnet --info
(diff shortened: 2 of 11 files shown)


src/TestCentric/testcentric.gui/Presenters/ITreeDisplayStrategyFactory.cs
```diff
@@ -0,0 +1,25 @@
...
+        /// <summary>
+        /// Creates a concrete class implementing interface ITreeDisplayStrategy
+        /// Supported displayStrategy:
+        /// - "NUNIT_TREE" (also default case) 
+        /// - "FIXTURE_LIST"
+        /// - "TEST_LIST"
+        /// </summary>
+        ITreeDisplayStrategy Create(string displayStrategy, ITestTreeView treeView, ITestModel testModel);
+    }
+}
```
CharliePoole, line 25: I see why you are doing this. I'll add a general comment about the approach, which I'm not sure is what we want over the long term.

rowo360:
I also think it's best from a testing point of view to test the TreeViewPresenter and the concrete DisplayStrategy separately (event generation vs handling). This was not possible in the past, but with the new interface ITreeDisplayStrategy this can now be implemented and I think we shoud go into this direction! For example, there are still no tests for the TestListDisplayStrategy/FixtureDisplayStrategy. But I would prefer to do this later.

Having said this, now a statement in the other direction: from my experience it's also beneficial to have some tests which are testing a set of classes instead of tests testing only one single class. Ultimately, the question is what is the unit? And I believe both kind of 'units' provide value. Overall, I would therefore not refactor **all** existing test classes so that they do some nice separate testing - no these test classes provide some value and I would keep them unchanged. But instead I would add new test classes to fill the gaps - these new classes might test single classes or a set of classes. Although I definitely think that the number of testing single classes should far exceed the other case :-)
You should just always be aware of whether you are doing one or the other - and maybe make it somehow visisble (Class name; namespace...)

Still the question remain about the best process to restructure the tests? I propose to create a separate issue for this task and describe our vision within that task.

CharliePoole:
Sounds as if we are thinking the same way. In my coaching, I generally try to get teams to test "units" of from 1 to 3  collaborator classes, and I don't count use of low-level common-use helpers in that number, so long as the helpers are already well tested. In this case, I'd tend to lean more heavily on tests that isolate the presenters from the strategies but keep a few around that test how they work together. I'm also all for higher-level functional tests and integration tests but generally run as a separate step.

I'm going to merge this now. :-)

Issue #1104, opened 2024-09-13
Engine Extensions dialog stopped working


FlorianAsal:
**Describe the bug**
Can't see the Extensions Assembly Path on the Extensions dialog, as the dialog crashes

**To Reproduce**
1. Go to 'Tools'
2. Go to 'Extensions'
2. Click on '/TestCentric/Engine/TypeExtensions/IAgentLauncher'
3. Installed Extensions ListView updates Items/flashing all the time, may caught in infinite loop -> no response from tool at all -> crash of Gui

**Expected behavior**
See the entries

**Screenshots**
![grafik](https://github.com/user-attachments/assets/b296115d-026e-4c62-9a71-24c7b4ad811d)

**Environment (please complete the following information):**
 - TestCentric Version 2.0.0-beta5(Release)
 - Installed via choco in default location
 - OS: Windows 10
 - Target Runtime for your test assembly: no assembly loaded

Wanted to debug the behavior but don't know where to provide the extensions. Starting the TestCentric.Gui.Exe from VisualStudio, no agent will be found and the extensions view is empty. Some help very much appreciated, as I want to have a closer look on other problems: Facing null exceptions on the gui and the agents when loading my net7.dll compiled as x86

CharliePoole:
Can you try this with the latest dev build in our myget feed? Use this command

```
choco install testcentric-gui --version 2.0.0-dev00444 --source https://www.myget.org/F/testcentric/api/v2
```

rowo360:
I can confirm the described behavior using **TestCentric Version 2.0.0-beta5(Release)**.
The behavior can also be reproduced with the latest dev version (using your choco command). But surprisingly the error does not show up with my own dev build. I didn't figure out the difference.

rowo360:
@Galgemaennle:
I am a newbie myself regarding the development of TestCentricGUI, so I can explain how I managed to debug TestCentric.GUI.Exe from VisualStudio. But please don't see it as some official recommendation - it's just my personal experience :-)
Currently TestCentric complains about missing agents, so I copied them to the correct location. The folder for the agents are named:
TestCentric.Extension.Net462PluggableAgent.2.5.0
TestCentric.Extension.Net60PluggableAgent.2.5.0
TestCentric.Extension.Net70PluggableAgent.2.5.0
TestCentric.Extension.Net80PluggableAgent.2.5.0

[...]

PR #1107, opened 2024-09-16
Break resize loop in extension dialog

FlorianAsal:
This fix breaks the resizing infinity loop of issue #1104
As the listbox appears to be very small and even gets covered by a horizontal scrollbar, it's configured to increase its size with the dialog size.


rowo360 (approved):

rowo360:
@Galgemaennle:
Super, that looks great. Not only did you fix this annoying bug, you also made sure that the height of the extension dialog can now be changed. I think that was indeed also urgently needed, as the list of extensions was otherwise tiny (especially when the system starts to add scrollbars to this ListView sometimes).
Thanks a lot for this improvement!

I think we can merge this PR right away. 
The only thing I'm wondering now is whether we want to increase the default height of this dialog slightly (now, that the Dialog is resizable in height :-))
That's not necessary, but just an observation/idea. What do you think?

PR #1109, opened 2024-09-18
Fix crash when opening XmlDisplay for a TestGroup node

### LINKED ISSUE #1105: Crash "View as XML..." executed on tree category node
**Describe the bug**
This bug is a continuation of the work to support the different tree view display formats (#1101).
There's a crash when executing "View as XML..." on a tree category node.

**To Reproduce**
1. Open a test project
2. Switch tree view mode to 'Fixture list' 
3. Switch tree grouping to 'Category'
4. Open context menu on some category node and select "View as XML..."
5. Crash

**Important**
This crash is not happening only for category nodes, but instead for all kind of group nodes. For example also for "Outcome" grouping - nodes 'passed' and 'not executed' or "Duration" grouping - nodes 'fast' and 'slow'.

 - TestCentric Version - latest developer build 2.0.0-dev00451
 - OS: Windows 11

rowo360:
This PR closes #1105 by disabling the context menu item 'View as XML..' for all kind of GroupNodes.

Here's a screenshot:

In addition, the use case must also be taken into account if the dialog is 'pinned' and the user selects some GroupNode, now.
The proposed solution keeps the dialog open, but deletes the entire content and only displays the name of the tree node in it's caption. Here's a screenshot:

Actually, the context menu no longer needs to be disabled, because the Dialog can handle GroupNodes somehow now. But from my point of view it's OK to keep it disabled, as no real XML content is displayed.

### DIFF
--- a/src/TestCentric/testcentric.gui/Dialogs/XmlDisplay.cs
+++ b/src/TestCentric/testcentric.gui/Dialogs/XmlDisplay.cs
@@ -17,7 +17,7 @@ namespace TestCentric.Gui.Dialogs
         private ITestModel _model;
 
         private TreeNode _treeNode;
-        private TestNode _testNode;
+        private ITestItem _testItem;
         
         public XmlDisplay(ITestModel model)
         {
@@ -47,13 +47,17 @@ namespace TestCentric.Gui.Dialogs
                 throw new ArgumentNullException(nameof(treeNode));
 
             _treeNode = treeNode;
-            _testNode = treeNode.Tag as TestNode;
+            _testItem = treeNode.Tag as ITestItem;
+
+            if (_testItem == null)
+                throw new ArgumentException("Unknown object associated to treenode");
 
             SuspendLayout();
-            TestName = (treeNode.Tag as ITestItem)?.Name;
+            TestName = _testItem.Name;
 
             // Display empty XML content for TestGroup nodes (for example category nodes)
-            XmlNode fullXml = (_testNode != null) ? GetFullXml(_testNode) : null;
+            TestNode testNode = _testItem as TestNode;
+            XmlNode fullXml = (testNode != null) ? GetFullXml(testNode) : null;
             xmlTextBox.Rtf = new Xml2RtfConverter(2).Convert(fullXml);
 
             ResumeLayout();
@@ -63,7 +67,8 @@ namespace TestCentric.Gui.Dialogs
 
         public void OnTestFinished(ResultNode result)
         {
-            if (result.Id == _testNode.Id)
+            TestNode testNode = _testItem as TestNode;
+            if (testNode != null && result.Id == testNode.Id)
                 Invoke(new Action(() => Display(_treeNode)));
         }
 
--- a/src/TestCentric/testcentric.gui/Dialogs/XmlDisplay.cs
+++ b/src/TestCentric/testcentric.gui/Dialogs/XmlDisplay.cs
@@ -50,8 +50,10 @@ namespace TestCentric.Gui.Dialogs
             _testNode = treeNode.Tag as TestNode;
 
             SuspendLayout();
-            TestName = _testNode.Name;
-            var fullXml = GetFullXml(_testNode);
+            TestName = (treeNode.Tag as ITestItem)?.Name;
+
+            // Display empty XML content for TestGroup nodes (for example category nodes)
+            XmlNode fullXml = (_testNode != null) ? GetFullXml(_testNode) : null;
             xmlTextBox.Rtf = new Xml2RtfConverter(2).Convert(fullXml);
 
             ResumeLayout();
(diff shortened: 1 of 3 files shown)


src/TestCentric/testcentric.gui/Dialogs/XmlDisplay.cs
```diff
@@ -50,8 +50,10 @@ public void Display(TreeNode treeNode)
             _testNode = treeNode.Tag as TestNode;
 
             SuspendLayout();
-            TestName = _testNode.Name;
-            var fullXml = GetFullXml(_testNode);
+            TestName = (treeNode.Tag as ITestItem)?.Name;
+
+            // Display empty XML content for TestGroup nodes (for example category nodes)
+            XmlNode fullXml = (_testNode != null) ? GetFullXml(_testNode) : null;
```
CharliePoole, line 56: I'm OK to merge this, but first read my general comment about the overall use of ITestItem vs TestNode.

src/TestCentric/testcentric.gui/Presenters/TreeViewPresenter.cs
```diff
@@ -450,6 +450,9 @@ private void InitializeContextMenu()
 
             var layout = _model.Settings.Gui.GuiLayout;
             _view.TestPropertiesCommand.Visible = layout == "Mini";
+
+            // For example test category nodes or test result nodes (passed/failed)
+            _view.ViewAsXmlCommand.Enabled = !(_view?.ContextNode?.Tag is TestGroup);
```
CharliePoole, line 455: I agree with your comment that this is no longer needed so long as we display something for categories, etc.

CharliePoole (changes requested):
I'd like you to look slightly deeper into this display.

You were initially confused looking at the original code, and I can see why. It's written as if everything should be a testnode. That's because this is is older code from before we allowed display of XML under the FixtureList and TestList strategies. In the NUnit strategy, every `TreeNode` tag is a `TestNode` but in general it's now an `ITestItem`. 

I'm glad you drew my attention to this. I'm thinking we should...
1. Have a `_testItem` field in place of `_testNode`.
2. Throw in `Display` if the tag is not an `ITestItem`.
3. Eliminate any other casts of the tag in the class and use the `_testItem` instead.
4. In the case of `OnTestFinished`, you should note that there can't be a ResultNode unless there was a test executed, so a cast of `_testItem` to `TestNode` must alwas succeed.

Would you first review my thinking and tell me if it makes sense? If so, I'd prefer to do it now, and get this class working once and for all. But if you'd rather merge this and come back later, feel free to argue. :-)

rowo360:
That's fine, I can continue to work on this and improve the code.
I have two questions regarding your suggestions:
- the new field `_testItem` will be of type ITestItem, right?
- Yes, I think we'll not have any casts for the Tag anymore. But I expect that we still need a cast of the `_testItem` to TestNode. At least before calling method GetFullXml() (or inside that method) and in the OnTestFinished() method. But that's ok, isn't it?

rowo360:
I forgot to comment on point no 4: 
Can the _testItem here not be a TestNode after all? Imagine I 'pinned' the dialog and select a TestGroup. Now the _testItem is not a TestNode, but I can start a test run and the OnTestFinished() method gets invoked.

CharliePoole:
Yes, `private ITestItem _testItem` and the arg to Display would be ITestItem, because that's the most we can guarantee.

Yes, the Tag is an `object` so I think we should cast it to a known type (`ITestItem`) only once. Currently, there are two methods needing a Testode, `Display` and `OnTestFinished` they would have to cast _testItem. I think `GetFullXml` should continue to get TestNode because that's what it needs and it should only be called if we have one. Type safety makes things cleaner.

You're right about point #4. You need a cast. But there's nothing to do if it's not a `TestNode`.

src/TestCentric/testcentric.gui/Presenters/TreeViewPresenter.cs
CharliePoole: Weren't you going to remove this?

CharliePoole (approved):
Thanks for reworking this. Only thing I noticed is that you left in the code in TreeViewPresenter to disable the menu item for non-testnodes. You had commented that we no longer needed to disable it and I agree. But I'm OK either way. If you make that change, just merge if GitHub lets you. Otherwise, I'll re-approve. (I always see a checkbox, which allows me to override the approval rules, but that may be because I'm the owner.)

rowo360:
That's right, I haven't done that yet! But I've made up for it now!
So I rollbacked that changes regarding enabling/disabling the context menu item, so that the 'View as XML..:' is always enabled again.
BTW: A NullCheck has been cheated into this commit, which was intended for later. But I hope that this kind of check is not critical.
I will do the merge, now.

CharliePoole:
Great! In addition, this verifies you have all the github superpowers you need. :-)

PR #1110, opened 2024-09-18
Fix merge error: Use _treeDisplayStrategyFactory for creation of display strategies

rowo360:
Unfortunately, I am only now realizing that my merge for PR #1102 was not 100% accurate or not 100% as it was intended. I'm a bit surprised and I'm not sure when I made the mistake: it was not optimal that I did conflicting changes in two separate branches/PRs in parallel, maybe I did not solve the merge conflicts as indented (I need some better tooling support) or when I did the rollback of that optimization I did not work carefully....?

But nonetheless, I add this treeDisplayStrategyFactory to the TreeViewPresenter with the intention that it is used in all places. But unfortunately there was one code location remaining, which creates the Tree Strategies on its own. This is fixed now.

One general question:
Do you prefer that I write an issue for this PR?

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/TreeViewPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/TreeViewPresenter.cs
@@ -116,7 +116,7 @@ namespace TestCentric.Gui.Presenters
                         break;
                     case "TestCentric.Gui.TestTree.DisplayFormat":
                         {
-                            Strategy = CreateDisplayStrategy(_treeSettings.DisplayFormat, _view, _model);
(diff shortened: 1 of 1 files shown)


CharliePoole (approved):
Don't sweat it! Having multiple PRs open can create difficulties like this as has happened to me as well. One trick if working on multiple things together is to wait until one PR closes before starting another. After the first PR is merged, rebase  your other open branches on top of main and fix any conflicts before continuing.

I'd ask for a new issue if we had released toe fix already but it's fine that this will be represented in the release notes as a single  fix.

Issue #1118, opened 2024-09-21
Test result icon in tree is not applied for group nodes


rowo360:
**Describe the bug**
When I run a test, the test results are displayed in the TreeNode with a matching icon.
This applies for all test nodes except group nodes.

Here are some screenshots:

1. 'Fixture list' view and group by 'Category' => run all tests
<img src="https://github.com/user-attachments/assets/244f3d2a-22d3-4530-8c23-a84f58d9bc31" width="250">

=> All category nodes keep their 'initial' state.

2. 'Fixture list' view and group by 'Category' => run only one single test case 'TestB'
<img src="https://github.com/user-attachments/assets/d877afb3-fd1f-4eaa-bbda-5e5de3ec5b98" width="250">

=> State of test fixture 'Feature1Tests' is updated; but Category node not

3. 'Test list' view and group by 'Outcome' => run tests of group 'Failed'
<img src="https://github.com/user-attachments/assets/2ed381d3-fae1-4488-8aaf-3ee94ea5b89b" width="250">

=> 'Failed' node is not getting updated

**Expected behavior**
I expect that the test result state is applied to all kind of tree node.
Even if only a subset of the tests is executed, its test results should be passed upwards. Just as I know it from the NUnitTreeDisplay format.

**Environment (please complete the following information):**
- TestCentric Version - latest developer build 2.0.0-dev00451
- OS: Windows 11

rowo360:
I believe that the current system behavior can be explained by looking at the XML structure. Whenever the execution of a test case is completed, we receive an event and perform actions to update the icon. 

That's fine, however group nodes (like categories) are no represented in this XML structure, so we'll receive no events for them.
This means that we have to determine the correct test result for groups ourselves and then set the icon.

**Good news:**
I think that exactly this behavior is already implemented in the code base. But unfortunately it's not invoked anymore.
This code is located here: `CategoryGrouping.OnTestFinished()`

<img src="https://github.com/user-attachments/assets/62a2d791-84d6-472e-b03b-32492ace1996" width="400">

So my idea is to simply reactivate this code. If that works, then I can check why this is actually implemented in the CategoryGrouping class and how the two DuractionGrouping and OutcomeGrouping behave.

CharliePoole:
Cool! Of course, it may be disabled because it caused some problem or it may be unfinished work, but let's find out.

rowo360:
Issue is solved by PR #1121

Issue #1106, opened 2024-09-16
Tree format 'Test list' and grouping 'By Category' => no categories are displayed


rowo360:
**Describe the bug**
This bug is a continuation of the work to support the different tree view display formats (https://github.com/TestCentric/testcentric-gui/issues/1101).

Maybe this issue is no bug at all, but instead the expected system behavior. But I was irritated for a long time and thought something wasn't working properly. Now that I have analyzed, debugged and understood it, I can explain the behavior. But maybe other users feel the same way and we can support them better.

**To Reproduce**
1. Open test project which contains some test categories
2. Switch tree format to 'test list'
3. Switch grouping to ' By Category'
=> Categories in test project are not found!

<img src="https://github.com/user-attachments/assets/7b5a454e-ccb6-4217-a68c-7af209314b20" width="250">

However if I switch to grouping 'By Category (extended)' the categories are suddenly displayed:
<img src="https://github.com/user-attachments/assets/ddc0f451-9335-457f-838f-6030de927f51" width="250">

**Expected behavior**
I expected that the categories are already shown when selecting 'By category' grouping.

**Environment (please complete the following information):**
- TestCentric Version - latest developer build 2.0.0-dev00451
- OS: Windows 11

rowo360:
The actual root cause of this behavior is the location at which I defined the test category attribute.
In my case I defined them on the **TestFixture** level, only.

I was not aware of that it make a difference!
But while debugging this use case, I noticed that when working in 'test list' mode, the TestGrouping only receives the XML nodes for single test cases. And it only checks, if this test case has the category set - which is not the case. Therefore it won't find the catgories.
Having understood this, I added a test category to test cases directly and yes indeed, now the 'By category' grouping work as expected.
<img src="https://github.com/user-attachments/assets/075376ca-3012-4634-9386-b6f4786cf5b6" width="250">

For a long time I didn't understand the difference between 'By category' and 'By category (extended)'. Now, as I'm aware of the XML and debugged it, I know that the extended mode will search also in the anchestors for a matching category.
Therefore the categories were found in my initial experiement when I used 'By category (extended)' - because the category at the test fixture were found.

When I started this research, I thought that it didn't matter where I defined the attribute (Fixture or test case level). In both cases I expected that my test case will have the category.

Overall, I wonder if users will understand the difference between 'By category' and 'By category (extended)'? Maybe a different naming will help? Or we can decide that we only support the extended mode?

[...]

PR #1121, opened 2024-09-22
Test result icon for test groups

rowo360:
I propose a solution in this PR for issue #1118. The current state is still not the final one, because all the tests are missing yet (and I expect that these scenarios are good testable). But let's discuss the proposed solution first.

I split the solution into three separate commits, hoping that it's easier to keep an overview.
1. commit: Reactivate existing solution to update tree icons for category groups
2. commit: Support update of tree icons for outcome and duraction groups
3. commit: Enhance update of tree icons for category nodes

I'll add some details in the next comments.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/CategoryGrouping.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/CategoryGrouping.cs
@@ -43,31 +43,6 @@ namespace TestCentric.Gui.Presenters
                     group.ImageIndex = _displayStrategy.CalcImageIndexForGroup(group);
         }
 
-        public override void OnTestFinished(ResultNode result)
-        {
-            var imageIndex = DisplayStrategy.CalcImageIndex(result.Outcome);
-            if (imageIndex >= TestTreeView.SuccessIndex)
-            {
-                var treeNodes = _displayStrategy.GetTreeNodesForTest(result);
-                foreach (var treeNode in treeNodes)
-                {
-                    var parentNode = treeNode.Parent;
-                    TestGroup group = treeNode.Parent?.Tag as TestGroup;
-                    if (group == null)
-                    {
-                        continue;
-                    }
-
-                    // Category nodes are updated either as soon as a test result for all children is present or at the end of the test run (method OnTestRunFinished)
-                    if (TestResultExistsForAllChildren(group))
-                    {
-                        int groupImageIndex = _displayStrategy.CalcImageIndexForGroup(group);
-                        parentNode.SelectedImageIndex = parentNode.ImageIndex = group.ImageIndex = imageIndex;
-                    }
-                }
-            }
-        }
-
         public override TestGroup[] SelectGroups(TestNode testNode)
         {
             List<TestGroup> groups = new List<TestGroup>();
@@ -115,26 +90,6 @@ namespace TestCentric.Gui.Presenters
 
                 return x.Name.CompareTo(y.Name);
             });
-        }
-
-        private bool TestResultExistsForAllChildren(TestGroup group)
-        {
-            foreach (TestNode testItem in group)
-            {
-                ResultNode result = _displayStrategy.GetResultForTest(testItem);
-                if (result == null)
-                {
-                    return false;
-                }
-
-                var imageIndexInGroup = DisplayStrategy.CalcImageIndex(result.Outcome);
-                if (imageIndexInGroup < TestTreeView.SuccessIndex)
-                {
-                    return false;
-                }
-            }
-
-            return true;
         }
 
         #endregion
--- a/src/TestCentric/testcentric.gui/Presenters/CategoryGrouping.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/CategoryGrouping.cs
@@ -28,7 +28,7 @@ namespace TestCentric.Gui.Presenters
 
         #region Overrides
 
-        public override string ID => _includeAncestors ? "CATEGORY_EXTENDED" : "CATEGORY";
+        public override string ID => "CATEGORY";
 
         public override void Load(IEnumerable<TestNode> tests)
         {
--- a/src/TestCentric/testcentric.gui/Presenters/CategoryGrouping.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/CategoryGrouping.cs
@@ -52,13 +52,17 @@ namespace TestCentric.Gui.Presenters
                 foreach (var treeNode in treeNodes)
                 {
                     var parentNode = treeNode.Parent;
-                    if (parentNode != null)
+                    TestGroup group = treeNode.Parent?.Tag as TestGroup;
+                    if (group == null)
                     {
-                        var group = parentNode.Tag as TestGroup;
-                        if (group != null && imageIndex > group.ImageIndex)
-                        {
-                            parentNode.SelectedImageIndex = parentNode.ImageIndex = group.ImageIndex = imageIndex;
-                        }
+                        continue;
+                    }
+
+                    // Category nodes are updated either as soon as a test result for all children is present or at the end of the test run (method OnTestRunFinished)
+                    if (TestResultExistsForAllChildren(group))
+                    {
+                        int groupImageIndex = _displayStrategy.CalcImageIndexForGroup(group);
+                        parentNode.SelectedImageIndex = parentNode.ImageIndex = group.ImageIndex = imageIndex;
                     }
                 }
             }
@@ -111,6 +115,26 @@ namespace TestCentric.Gui.Presenters
 
                 return x.Name.CompareTo(y.Name);
             });
+        }
+
+        private bool TestResultExistsForAllChildren(TestGroup group)
+        {
+            foreach (TestNode testItem in group)
+            {
+                ResultNode result = _displayStrategy.GetResultForTest(testItem);
+                if (result == null)
+                {
+                    return false;
+                }
+
+                var imageIndexInGroup = DisplayStrategy.CalcImageIndex(result.Outcome);
+                if (imageIndexInGroup < TestTreeView.SuccessIndex)
+                {
+                    return false;
+                }
+            }
+
+            return true;
         }
 
         #endregion
--- a/src/TestCentric/testcentric.gui/Presenters/CategoryGrouping.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/CategoryGrouping.cs
@@ -60,7 +60,8 @@ namespace TestCentric.Gui.Presenters
                     Groups.Add(group);
                 }
 
-                groups.Add(group);
+                if (!groups.Contains(group))
+                    groups.Add(group);
             }
 
             if (groups.Count == 0)
--- a/src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
@@ -88,6 +88,10 @@ namespace TestCentric.Gui.Presenters
             _view.ResetAllTreeNodeImages();
         }
 
+        public virtual void OnTestRunFinished()
+        {
+        }
+
         // Called when either the display strategy or the grouping
         // changes. May need to distinguish these cases.
         public void Reload()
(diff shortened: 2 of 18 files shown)


rowo360:
Some details about the **first commit**, which has the focus to update tree icons for **category groups**.

The essential part is in the class `GroupDisplayStrategy` - now the method `_grouping?.OnTestFinished(result)` is called instead of the method `ApplyResultToGroup(result, false)`. Unfortunately it's not visible in the diff-view, but we need to look to the code ourselves: The three grouping classes (CategoryGrouping, OutcomeGrouping and DuractionGrouping) already implemented the method `OnTestFinished`, but that code was not invoked anymore so far.

Looking into the git history I spotted commit #865 which was dealing with the elimination of the TreeViewElement and ITreeView. This commit exactly removed the call to `_grouping?.OnTestFinished(result)` and I'm rolling it back, now. But I think, it's for good. :-)

One concern might be that the method `ApplyResultToGroup(result, false)` is not called anymore for category grouping. But I think that it's for the best. Well, I already noticed some strange behavior sometime ago and wanted to raise a separate issue about it at times. Now I happen to stumble across the exact code that is responsible for this. Here's a screenshot about it:



I started a test run for one category (DisplayFormat: Testlist, GroupBy: Category (extended)). After the test run the category 'Feature_2' is not displayed anymore in tree and the test cases are listed under category 'None'. This behavior might be related to our discussion in #1106 , but actually the method `ApplyResultToGroup(result, false)` is responsible for moving nodes between group nodes. 
That's of course necessary for Duration and OutCome grouping, but the test categories will never change dynamically during a test run. That's why I think it's really right to stop using this method for categories - and it also fixes this strange behavior!

Next, I learned that there's an ImageIndex also in the TestGroup class. Sometime ago we decided to reset all ImageIndex at the start of a test run, but I missed to reset these ImageIndex properties in the TestGroups. And now these outdated ImageIndex at the TestGroups do some harm. Therefore I introduced a OnTestRunStarting() method at the Strategy class, which forward the method to the TestGrouping, which can perform the reset.

With these changes the existing method `CategoryGrouping.OnTestFinished()` gets invoked again, which already implemented the update of the tree icons for category nodes. I am almost satisfied with the existing behavior. The icon is always the right one! And I think this is the most important thing.
Less important, but still noticeable, is that the icon is set a little too early. Currently it's already set when the first test fixture (Fixture list mode) or first test case (Test list mode) is finished. It might be updated when the resulting state changes with some subsequent tests. 

But overall I expect that the icon should only be updated when the last test in this category has been completed. I think that's how I know the behavior so far. But let's discuss - I tried an improved solution with commit no.3

rowo360:
Some details about the **second** commit, which has the focus to update tree icons for **outcome** and **duration** groups.

Actually, the solution for the categories also works for outcome and duration grouping. So one solution might be to move that code from CategoryGrouping to GroupingStrategy class, so that it can be used for all three kind of groupings. I made an attempt and unfortunately notice now, looking at the commit, that I missed to remove that code again (method `UpdateTestResult(ResultNode result)` - it's dead code, now).
But the little drawback of this solution is also noticable here: for example the icon of node 'Passed' gets immediately updated after the first test finished. Therefore I made an attempt for a different solution:

It's based on the consideration that I can only actually set the icon correctly once the last test has been run. I cannot predict if a test will run 'slow' or 'fast' or if a test will 'pass' or 'fail' and finally in which group that test belongs to. Therefore, it is impossible to determine the correct time when the last test of one of these groups was executed and you can now set the icon for this group. 

Therefore I suggest to set the icon of these group nodes at the end of a test run!
Technically I'm handling the RunFinished event, passing it from the TreeViewPresenter via the Strategy to the Grouping. And most important the Grouping class can use an existing method `Strategy.CalcImageIndexForGroup(group)` for determining the group icon. So this essential logic must not be implemented.

rowo360:
Some details about the **third** commit, which enhance the update of tree icons for **category groups**.

I think that the icon of category nodes can either be updated as soon as a test result for all children is present or at the end of the test run. Maybe I'm oversimplifying it. And it shouldn't be 'test result for all children', but 'last test result' of that category. (imagine that I execute only a subset of tests of a category). But I have the feeling that it will be difficult to determine that a test result is really the last one of one category.

Therefore I stick with the 'test result for all children' approach and I consider it as an improvement compared to setting the icon too early.

rowo360:
Sorry for the long comments and all the text - I hope it helps with understanding and discussion.

CharliePoole:
This will take me a while to review. :-)

src/TestCentric/testcentric.gui/Presenters/CategoryGrouping.cs
```diff
@@ -52,13 +52,17 @@ public override void OnTestFinished(ResultNode result)
...
-                            parentNode.SelectedImageIndex = parentNode.ImageIndex = group.ImageIndex = imageIndex;
-                        }
+                        continue;
+                    }
+
+                    // Category nodes are updated either as soon as a test result for all children is present or at the end of the test run (method OnTestRunFinished)
+                    if (TestResultExistsForAllChildren(group))
+                    {
+                        int groupImageIndex = _displayStrategy.CalcImageIndexForGroup(group);
+                        parentNode.SelectedImageIndex = parentNode.ImageIndex = group.ImageIndex = imageIndex;
```
CharliePoole, line 65: Throwing out an idea here... tell me what you think.

What if all the updating of parent nodes was handled by the view? Of course, we use the presenter to make testing easier but it adds complexity too. My logic is as follows:
1. Everything in the tree has an index and that index is supposed to show the "worst" result from all its children. It seems like it should not matter whether the parent is another test, a category or any other group node.
2. So what  if we did this update to higher levels at the view level, dealing only with tree nodes?
3. When a node's index is being updated, that node could propagate the same index to all higher level nodes IFF the level of the node were not already worse than that update.
4. We would work our way up the tree until there was no parent OR until no update was needed.
 
Would this be simpler than what we do now?

src/TestCentric/testcentric.gui/Presenters/CategoryGrouping.cs
```diff
@@ -113,6 +117,26 @@ private void AddGroup(TestGroup group)
...
+                var imageIndexInGroup = DisplayStrategy.CalcImageIndex(result.Outcome);
+                if (imageIndexInGroup < TestTreeView.SuccessIndex)
+                {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
```
CharliePoole, line 139: Seems like a lot of work per update, but see earlier comment about letting the view do it.

src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
```diff
@@ -83,6 +83,15 @@ public virtual void OnTestFinished(ResultNode result)
                 _view.SetImageIndex(treeNode, imageIndex);
         }
 
+        public virtual void OnTestRunStarting()
+        {
+            _view.ResetAllTreeNodeImages();
```
CharliePoole, line 88: Is this needed for the NUnit tree?

CharliePoole (changes requested):
Two main changes...

1. I can see we are getting a lot of complications in the code due to the introduction of multiple strategies and because we are trying to use MVP properly. In the original NUnit GUI (V2), we simply gave the parent node the "worst" color index comparing its present setting to the setting of a child test that had just completed. This depends on the numerical order of the indices and on the policy of always propagating errors up the tree display. 

You could either do this in the view or in the presenter itself, working only with tree nodes. Either approach is less "pure" than what we are doing now, but even so might be simpler. 

If this makes sense to you, I suggest trying this change before the following suggestion because I suspect it will simplify the remaining code.

2. As we discussed this AM, let's just have a Category grouping in the UI rather than Category and Category (extended). In CategoryGrouping, we should change the semantics so it works like Extended Category in the TestList strategy only. IOW, if the user asked for a TestList, we will assume they want Categories to work as if they are inherited. I don't think this will be surprising to most people.

src/TestCentric/testcentric.gui/Presenters/TreeViewPresenter.cs
```diff
@@ -99,11 +99,16 @@ private void WireUpEvents()
...
                 _model.ClearResults();
-                _view.ResetAllTreeNodeImages(); 
+                Strategy.OnTestRunStarting();
                 CheckPropertiesDisplay();
                 CheckXmlDisplay();
             };
 
+            _model.Events.RunFinished += (ea) =>
+            {
+                Strategy?.OnTestRunFinished();
```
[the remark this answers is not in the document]
rowo360: Yes indeed, I can remove the '?' operator here. I was probably a little too careful - I'll remove it and double-check it.

src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
rowo360: Yes, this is also required for the NUnit tree. When starting a test run, we want that the images of all nodes are reset to the default (regardless of the current display format: NUnit, FixtureList or TestList). 
This statement was previously located in the TreeViewPresenter.WireUpEvents() - RunStarting event handler, but I moved it along to the new method Strategy.OnTestRunStarting().

src/TestCentric/testcentric.gui/Presenters/CategoryGrouping.cs
rowo360: I agree with you that this code part is somehow complex and we should consider how to simplify it.
If I look to the implementation of `TestGrouping.OnTestFinished`, it looks pretty close to your suggestion. That implementation is using the existing method `GroupDisplayStrategy.CalcImageIndexForGroup` - ok, that's not the 'bubble up' approach you mentioned, but instead an approach to get the 'worst' test result of any children.

But overall, I consider this approach used for the DurationGrouping and OutcomeGrouping as rather simple. We just need the OnTestRunFinished() event and update the tree node images of the 'Passed', 'Fast'... nodes at this point of time.

One idea: I could remove the entire `CategoryGrouping.OnTestFinished` method! That means that there's no difference between those Groupings anymore. I think that would simplify the code also! There's only one drawback from user point of view, that the image of a category node is updated only at the end of the test run and not immediately when the last test case of that category is finished.
That's not 100% accurate, but I believe it's acceptable. If users complain heavily about it, it can be improved at that point in time. Is this suggestion acceptable for you?

I have also thought about the 'bubble up' approach, which I expect to work properly. We only have to observe or decide at what time the parent should be updated? Is it right away when the first test case result of any child arrives? And of course it must be updated as soon as further test results arrive. That's different than the current behavior, but also fine!
Or we update the parent node only at that point of time, when the test case result of the last child arrives - but that starts to get complicated again.

Maybe I'm already too involved in the implementation, to get a clear picture.
But currently I propose just to remove `CategoryGrouping.OnTestFinished` method to reduce complexity.

CharliePoole: Seems like a reasonable compromise. We could then live with it through the next beta and see how it seems both to us and users.

rowo360:
The context menu 'By Category (Extended)' is removed now from user point of view.



Internally the category grouping still supports both modes (normal and extended mode). This was and is still controlled by a constructor parameter 'includeAncestors'. The only difference is that this parameter is set now depending on the current Strategy (FixtureList vs TestList).

I also checked the system behavior if a VisualState file with 'Category_Extended' mode is loaded. In that case the system will use the default grouping ('Assembly') - but most important the system doesn't crash.

I also removed the 'Category_Extended' handling from some tests. These tests didn't start to fail, but I prefer to clean up those right away.

CharliePoole (approved):
This is looking good to me. I'm going to bed and you're still working, so I'm approving now. When you're ready, merge if GitHub permits, which is what I hope I managed to get set up. If not, just remove the draft status and I'll do it in the morning.

rowo360:
I added some tests for the different grouping and strategy classes. Some of them are pure unit tests (testing a single class) - some of them are testing a set of classes.

rowo360:
Actually, I thought I was really done with this task now. But while testing I noticed an edge case, which I like to solve right away.
It's possible that a user defines the same test category at the test fixture **and** the test case!

Unfortunately, this means that the test cases are inserted twice when working with 'Test list' mode and 'Category' grouping.


This behavior is caused by the CategoryGrouping class which will find the category at the test case and the test fixture.

CharliePoole:
I agree. Look at the code that creates the filter for the category. It should not be creating duplicates in the selection.

Issue #1130, opened 2024-09-30
Crash: Start test run using context menu while test is running


rowo360:
**Describe the bug**
When a test run is started all toolbar buttons like 'Run all' or 'Run selected' are disabled until the test run is finished.
However the context menu entry 'Run' is still enabled.

**To Reproduce**
1. Load a project and start a test run 
    Important: the tests should run slow, so that there's enough time to execute the next steps **before** this run is finished.
2. Open tree context menu and select 'Run'
3. Crash

**Expected behavior**
I believe the context menu entry 'Run' should be disabled while a test is running. That's the identical behavior to all those toolbar buttons.

**Screenshots**
<img src="https://github.com/user-attachments/assets/6af7f2d6-b0b4-4b3a-a63e-1e91df9b51b3" width="600">

**Environment (please complete the following information):**
 - TestCentric Version: latest developer build (and probably also 2.0.0-beta5)
 - OS: Windows 11

rowo360:
The actual crash is happening from the progress bar! Somehow it get confused by the number of tests in the test run and the maximum value. I didn't look to the details here, because I think we should stop this use right from the start and disable the context menu in this case.
Probably the 'Debug' context menu entry should be disabled as well.

CharliePoole:
They should be added to the relevant tests as well.

rowo360:
This issue is solved by PR #1131 => closed

PR #1131, opened 2024-10-01
Disable context menu entry 'Run' and 'Debug' during a test run

rowo360:
This PR solves #1130 by disabling both context menu entries 'Run' and 'Debug' during a test run.
Here's a screenshot:

I wandered back and forth between 2 possible solutions: either setting the Enabled state whenever the context menu is opened, or set to disable in event RunStarted and enable again in event RunFinished.

From my point of view the first option is a bit simpler - because only one code location is taking care about the Enabled state. So I choose this approach.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/TreeViewPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/TreeViewPresenter.cs
@@ -436,6 +436,10 @@ namespace TestCentric.Gui.Presenters
 
             var layout = _model.Settings.Gui.GuiLayout;
             _view.TestPropertiesCommand.Visible = layout == "Mini";
+
+            // If a test is already running, no new test run should be started.
+            _view.RunContextCommand.Enabled = !_model.IsTestRunning;
+            _view.DebugContextCommand.Enabled = !_model.IsTestRunning;
         }
 
         #endregion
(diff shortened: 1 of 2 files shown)


src/TestCentric/testcentric.gui/Presenters/TreeViewPresenter.cs
```diff
@@ -436,6 +436,10 @@ private void InitializeContextMenu()
 
             var layout = _model.Settings.Gui.GuiLayout;
             _view.TestPropertiesCommand.Visible = layout == "Mini";
+
+            // If a test is already running, no new test run should be started.
+            _view.RunContextCommand.Enabled = !_model.IsTestRunning;
+            _view.DebugContextCommand.Enabled = !_model.IsTestRunning;
```
CharliePoole, line 442: I think it's possible to have a situation where the tests are taking a while to unload or reload and the user right-clicks. So you should take that into account by checking `_model.HasTests`. See the `UpdateViewCommands` method in the main presenter for the full set of things we check in the main view.

CharliePoole (changes requested):
Looks good except for the one change. As usual, treat it as a suggestion and either do it or comment why you are not doing it, then merge. :-)

src/TestCentric/testcentric.gui/Presenters/TreeViewPresenter.cs
rowo360: Thanks for the suggestion. I updated this code part and consider the HasTests flag now!

rowo360:
Hmm strange - GitHub still disables the 'Merge pull request' button for me and states that I still need to address the requested changes ?!?!

CharliePoole (approved):

CharliePoole:
OK, Let's work within the bounds for now and I'll look at authorizations more when I can.

Issue #1111, opened 2024-09-19
Select checkbox of a category tree node and run selected tests => crash


rowo360:
**To Reproduce**
1. Open test project which contains some test categories
2. Change tree display format to 'test fixture' and grouping to 'By category'
3. Show checkboxes in tree
4. Select checkbox of one test category 
5. Start test run by toolbar item 'Run selected tests'
=> crash

Here's a screenshot:
<img src="https://github.com/user-attachments/assets/c9e6b8d3-2efa-48ab-893c-386139e68d65" width="250">

Please note: this crash happens not only with category tree nodes, but with all kind of group nodes.

**Environment:**
- TestCentric Version - latest developer build 2.0.0-dev00451
-  OS: Windows 11

rowo360:
I've the impression that this crash is caused again by some `TestNode` vs. `TestGroup` handling.
The actual crash is happing in this code location, but I believe it's not the root cause of this bug.

<img src="https://github.com/user-attachments/assets/1ebf7f1c-8098-4fe6-9fe8-f34641d6f9ec" width="700">

I think the critical part is that the TestSelection contains one 'null' item. 
(BTW: this NullReferenceException is happing in some bog statement - but the scenario does not work if you simply skip this log statement)

rowo360:
I think that the root cause is instead at the point where the SelectedTests are determined for the model. That's done in the TreeViewPresenter class. There's an EventHandler which is invoked whenever a checkbox in the tree is checked/unchecked.

<img src="https://github.com/user-attachments/assets/744e45d3-858f-445f-9a57-4b3e38844e80" width="400">

I think it's obvious right away from the screenshot that the handling for type `TestGroup` is missing.
I expect that this can be solved by adding all the children of a test group to the TestSelection. The children are the actual test cases or test fixtures, but I'm not aware that there's a TestGroup within a TestGroup possible.

Thinking about this code part, I wonder if the items in TestSelection must be distinct? That's actual regardless from this current issue: if I check the checkbox from a TestFixture and one of his test cases, this TestSelection will contain two items. And when starting a test run for this selection the TestRunner will receive the XML Ids of both tests.
Do you know if the runner is smart enough to run the tests only once?

CharliePoole:
[...]

PR #1132, opened 2024-10-01
Check/Uncheck tree node - consider test groups

rowo360:
This PR is a proposal for fixing issue #1111 .
Let's discuss and improve this proposal. If we come to a final conclusion, I'll add several tests for this use case.

The basic idea of this solution is that all children of a TestGroup are added to TestSelection. We cannot add the TestGroup itself, because it is unknown to NUnit framework. The children might be Testcases or TestFixtures nodes.

Please note also the LINQ statement selectedTests.Distinct() to ensure that all items are added only once.

From user point of view it's possible to select multiple group nodes now and execute this set of tests:

Another observation (independent from this issue and PR): if the user checks a group node or test fixture all tests of that group or of that test fixture are executed. It makes no difference if the user selects additionally any test case inside this group/test fixture:

I think that behavior is totally fine.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/TreeViewPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/TreeViewPresenter.cs
@@ -17,6 +17,7 @@ namespace TestCentric.Gui.Presenters
     using System.Drawing;
     using System.IO;
     using TestCentric.Gui.Controls;
+    using System.Linq;
 
     /// <summary>
     /// TreeViewPresenter is the presenter for the TestTreeView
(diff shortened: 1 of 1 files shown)


CharliePoole:
I just updated the recipe to version 1.3.2 to resolve the crashing error under AppVeyor.

Also, see my comment on #1111 and think about whether at least some of this can be resolved in the code for TestSelection. It seems to me that adding duplicate ids should just be ignored and adding null should throw.

rowo360:
Yes, it feels a little bit odd viewing the AppVeyor failures, but to continue with commits and merging. I immediately worry whether I have overlooked something. So it's good that you update the recipe! :-)

I think it's best that I close this PR now, isn't it? 
Because we'll no longer pursue this proposal / commit.
And start a new PR as soon as the TestSelection adaptation is finished.

CharliePoole:
@rowo360 Are you using version 1.3.2 of the recipe? That's  the one that switched the version of GitVersion to 5.12, which is the  last 5.x release. Earlier releases have this crash problem in certain situations while the 6.x releases have breaking changes that would require a lot of script updates.

PR #1133, opened 2024-10-02
Update to TestCentric.Cake.Recipe version 1.3.2

rowo360:
Update to TestCentric.Cake.Recipe version 1.3.2

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -4,7 +4,7 @@
 const string REF_ENGINE_VERSION = "2.0.0-dev00023";
 
 // Load the recipe
-#load nuget:?package=TestCentric.Cake.Recipe&version=1.3.1
+#load nuget:?package=TestCentric.Cake.Recipe&version=1.3.2
 // Comment out above line and uncomment below for local tests of recipe changes
 //#load ../TestCentric.Cake.Recipe/recipe/*.cake
 
--- a/package-tests.cake
+++ b/package-tests.cake
@@ -50,6 +50,7 @@ public static void DefinePackageTests()
     //Assemblies = new[] { new ExpectedAssemblyResult("mock-assembly.dll", "netcore-1.1") }
     //        }));
 
+    if (!BuildSettings.IsRunningOnAppVeyor)
     PackageTests.Add(new PackageTest(1, "Net50Test", "Run mock-assembly.dll under .NET 5.0",
         "net5.0/mock-assembly.dll",
         MockAssemblyExpectedResult("Net60AgentLauncher")));
@@ -75,6 +76,7 @@ public static void DefinePackageTests()
             Assemblies = new [] { new ExpectedAssemblyResult("aspnetcore-test.dll", "Net60AgentLauncher") }
         }));
 
+    if (!BuildSettings.IsRunningOnAppVeyor)
     PackageTests.Add(new PackageTest(1, "AspNetCore50Test", "Run test using AspNetCore under .NET 5.0",
         "net5.0/aspnetcore-test.dll",
         new ExpectedResult("Passed")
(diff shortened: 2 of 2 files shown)


rowo360:
This already looks much better than before (build+test is starting), but finally the pipeline seems to hang at this step (see AppVeyor log):
------------------------------------
Run mock-assembly.dll under .NET 5.0
------------------------------------
Running C:/projects/testcentric-gui/package/tests/nuget/TestCentric.GuiRunner.2.0.0-ci00486-UpdateCakeRe/tools/testcentric.exe with arguments net5.0/mock-assembly.dll  --work=C:/projects/testcentric-gui/package/results/nuget/TestCentric.GuiRunner//Net50Test/ --run --unattended --full-gui

Is this related to the build script or to the AppVeyor environment?

rowo360:
Ok, looking into the AppVeyor logs shows up that this .NET runtime is not installed anymore:

Microsoft.WindowsDesktop.App 5.0.17 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]

CharliePoole:
Yes, it's one of annoyances with AppVeyor that what's supported by the agents changes from time to time. I've seen this one before, in fact, back when .NET 5.0 was the newest runtime. :-) You can use the value of `BuildSettings.IsRunningOnAppVeyor` to conditionally add that test in package-tests.cake.

In the past, I have added code to the script to install unsupported runtimes on the fly. However, while it's possible, it's not made as simple as it is in some other CI servers.

rowo360:
Now I have to admit that I feel a little lost as to exactly where I could go from here. ;-)
I opened the TestCentric.Cake.Recipe solution, found the package-test.cake file. But I assume that this file should not be changed. So I looked for the usages of the PackageTest class and found some in file agent-factory.cake

Looking up and down without a specific goal I finally spot these lines in agent-factory.cake:


Is this the correct code location at all?

CharliePoole:
Ah you found that! :-) AgentFactory is a class used to generate the code for new pluggable agents, of which we have a large number. I actually started moving away from it but that's not really relevant here. It's not used in build.cake for the GUI.

In the GUI project, the package tests are right in the root, in `package-tests.cake`, which is loaded by `build.cake`. You need to find the line that adds the .NET 5.0 test and wrap it in `if (!BuildSettings.IsRunningOnAppVeyor { EXISTING CODE };

This may not help with the internal details but there is a website for how to use the recipe: https://recipe.test-centric.org

rowo360:
Thanks! Yes, that's the right file - I finally found it!
I will start an attempt now, but will be away for a short while afterwards...

CharliePoole:
OK, the failure message "your HEAD has moved after repo normalisation" occurs for PRs because the temporary merged HEAD is detached. That's an unresolved bug in GitVersion? Cake? AppVeyor? I don't know. :-(

I'm going to disable PR runs in appveyor and merge to main on my desktop.

CharliePoole:
Even better. I was able to merge it from the web page by checking a box to override the PR. That kicked off a build of main resulting in pushing 2.0.0-dev00488 to MyGet.

I believe that in future, things will just merge for you and the build of main will proceed automatically for you. We'll see. :-)

The only risk of merging without testing the PR is if more than one person is working on conflicting updates. That's not happening right now, fortunately, since I'm working mostly in the v1-build branch. When I do anything to main, I'll send you a note to warn you to update your branches. For that reason it's best not to have too many branches at once and to not create a PR too soon. If you need comments on a feature branch, I can diff it myself in order to comment.

I'll get a GitHub workflows build in place as soon as possible, so hopefully this situation won't last a long time.

rowo360:
Very nice 👍!
Thanks for fixing this "your HEAD has moved..." issue and performing the merge. And cheers - the build is working again :-) 
It immediately feels much more reassuring when there is a green dot next to a build.

And good suggestion regarding PRs and branches - I'll do that. And of course I'm also curious to see how things will look with GitHub workflows in the future.

PR #1136, opened 2024-10-08
Refactor TestSelection class to implement IEnumerable interface

rowo360:
This PR resolves issue #1111 

The root cause of that crash was that the code didn't handle checking/unchecking TestGroups nodes so far.

This solution improves the TestSelection class so that either TestNodes or TestGroups maybe added. For that purpose the internal implementation was changed: the TestSelection class exposes only the interface IEnumerable\<TestNode\> and uses a private Hashset for the list of nodes internally. So the choice about using a Hashset or List is only an internal implementation detail of that class.

As a consequence several code locations must be adapted. These adaptations were not conceptual changes, but mostly using Linq to access the IEnumerable\<TestNode\>.

### DIFF
--- a/.github/workflows/testcentric-gui-ci.yml
+++ b/.github/workflows/testcentric-gui-ci.yml
@@ -1,0 +1,79 @@
+﻿name: TestCentric.GuiRunner.CI
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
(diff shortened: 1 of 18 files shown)


CharliePoole:
If this is actually ready and only blocked because of AppVeyor, I can merge it on my machine and push straight to main.

Issue #1138, opened 2024-10-13
Crash - toolbar button 'Run selected tests'


rowo360:
**Describe the bug**
I noticed this crash when clicking on the toolbar button 'Run selected tests'.

<img src="https://github.com/user-attachments/assets/5cc60c3a-4fb0-4716-b1bd-0559a7f4870c" width="300">

This crash occurs only:
- A project is just opened
- The option 'Show checkboxes' is enabled
- The 'checked' state of any state was not changed at all

There must be an additional condition for this crash, but unfortunately I can't name it exactly. First I thought it's related to the current Strategy/Grouping, but repeating the same steps which were previously causing the crash suddently work.
That's strange and my only explanation for this is that an event may not be fired...

Actually I want to click 'Run all tests' and accidently hit 'Run selected tests' ;-)

**Environment (please complete the following information):**
- TestCentric Version - latest developer build
- OS: Windows 11

rowo360:
The actual crash occurs because of a `NullReferenceException`:

<img src="https://github.com/user-attachments/assets/ea5ac6e4-59b8-400a-8dc3-b5522b300582" width="600">

The `SelectedTests` is null here and this kind of input is not accepted by the method `_model.RunTests`

In common the `SelectedTests` property is set when checking/unchecking tree nodes or even when loading a project (with or without VisualState file), but for some reasons this is not triggered in my use case !?!

I propose to fix this issue simply by checking this null case. Another option might be to initialize the `SelectedTests' not with null, but with an empty TestSelection. But I prefer the simplest approach right now without dealing with all details.

rowo360:
This kept me wondering why the behavior seems so random. Usually you can quickly explain the behavior and it just seems random until you identify the root cause. That's why I started another attempt doing some further debugging/logging of this use case:
I didn't came to a final conclusion, but I describe my current investigation - maybe you notice something:

In common the property SelectedTests gets set in this use case (load project) by the TreeViewPresenter.AfterCheck:
<img src="https://github.com/user-attachments/assets/0e48d15e-156d-4bab-a78c-0339ddf84adb" width="400">

This method is invoked whenever the checked state (checkbox) of a tree node is changed. And in common this method gets called during project load, too. And in general: if this method is called, the SelectedTests property won't be null! It must therefore be the case that the method is not called in the crash scenario.

In common this method is invoked while loading the VisualState and applying the VisualState to each node:
[...]

PR #1139, opened 2024-10-13
Fix crash when 'run selected tests' but no tests are selected

rowo360:
This PR fixes issue #1138.

Unfortunately I didn't manage to detect the actual root cause of this crash. As described in the issue repeating the same steps which were previously causing the crash suddently work. That doesn't feel good, but I'd rather fix the problem now than keep looking for a long time.

So the property `SelectedTests` might be set to null in some scenarios, which will cause the crash when executing 'Run selected tests'. Therefore I propose to fix it simply by handling this case now.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
@@ -954,7 +954,8 @@ namespace TestCentric.Gui.Presenters
         private void RunSelectedTests()
         {
             var testSelection = _model.SelectedTests;
-            _model.RunTests(testSelection);
+            if (testSelection != null)
+                _model.RunTests(testSelection);
         }
 
         private void RunFailedTests()
--- a/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
@@ -231,6 +231,8 @@ namespace TestCentric.Gui.Presenters
                 }
             };
 
+            _model.Events.SelectedTestsChanged += (e) => UpdateViewCommands();
+
             _settings.Changed += (s, e) =>
             {
                 switch (e.SettingName)
@@ -754,9 +756,10 @@ namespace TestCentric.Gui.Presenters
             bool hasFailures = _model.HasResults && _model.ResultSummary.FailedCount > 0;
 
             _view.RunAllButton.Enabled =
-            _view.RunSelectedButton.Enabled =
             _view.DisplayFormatButton.Enabled =
             _view.RunParametersButton.Enabled = testLoaded && !testRunning;
+
+            _view.RunSelectedButton.Enabled = testLoaded && !testRunning && _model.SelectedTests != null && _model.SelectedTests.Any();
 
             _view.RerunButton.Enabled = testLoaded && !testRunning && hasResults;
 
@@ -954,8 +957,7 @@ namespace TestCentric.Gui.Presenters
         private void RunSelectedTests()
         {
             var testSelection = _model.SelectedTests;
-            if (testSelection != null)
-                _model.RunTests(testSelection);
+            _model.RunTests(testSelection);
         }
 
         private void RunFailedTests()
--- a/src/TestCentric/tests/Presenters/Main/CommandTests.cs
+++ b/src/TestCentric/tests/Presenters/Main/CommandTests.cs
@@ -250,9 +250,17 @@ namespace TestCentric.Gui.Presenters.Main
         [Test]
         public void RunButton_RunsSelectedTests()
         {
-            // TODO: Specify Results and test with specific argument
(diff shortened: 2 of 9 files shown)


src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
```diff
@@ -954,7 +954,8 @@ private void RunAllTests()
         private void RunSelectedTests()
         {
             var testSelection = _model.SelectedTests;
-            _model.RunTests(testSelection);
+            if (testSelection != null)
+                _model.RunTests(testSelection);
```
CharliePoole, line 958: How are we able to get here if there are no selected tests? While I agree that the code should check for null or empty on general principle, the overall strategy in the UI has been to disable any controls that are not usable at a given point in time. So, if there are no selected tests,  this code ought to be unreachable.

CharliePoole (changes requested):
I can see that a test selection must  be capable of being empty but it doesn't seem to me that we should  be calling the model with an empty selection.

Additionally, if no tests are selected, it seems like the controls and menu items used to run selected tests should be disabled.

rowo360:
That idea is even better: disable the toolbar item if no tests are selected! I like it!
I prepared a solution for this idea and added a new commit.
Basically I introduced a new event `SelectedTestsChanged` in the model. This event gets fired whenever some nodes are checked or selected in the tree and the TreeViewPresenter updates the `SelectedTests` property. The TestCentricPresenter class listens to this event and updates the enable state of the toolbar item accordingly.

Here are some screenshots to get a better impression:
In the first image the toolbar button is disabled because no node is checked, in the second image a node is checked and thus the toolbar button is enabled. The last image is without the option 'Show checkboxes' - in that case the toolbar button is always enabled - I didn't notice a scenario in which no node is selected.

rowo360:
While working on this issue I wondered if the toolbar tooltip is always accurate?
The tooltip states 'Run selected tests' - but wouldn't it be more accurate to state 'Run checked tests' when the checkboxes are shown in the tree? If no checkboxes are shown, we can continue to use 'Run selected tests' as the tooltip. 
That might be helpful to understand why the toolbar button is disabled.

CharliePoole:
In my mind, when I first did this "selected" meant highlighted OR checked when checks are enabled. I agree with you that this might be confusing to the user. However, I think there are actually three states here, not two, in the way it is implemented.

1. If no checkboxes are shown, selected means selected, i.e. highlighted.
2. If checkboxes are shown _and nothing is checked_ it still means the same thing.
3. if checkboxes are shown and one or more tests are checked, then "selected" means checked.

I suggest you first verify that it works as I say, especially the second condition. If it does, then I think you would change the language in the tooltip each time it displays based on whether anything is checked rather than whether checkboxes are shown.

CharliePoole:
Is a new event needed? I'm pretty sure we were already responding to changed selection. It might be a generic property change to our settings.

src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
```diff
@@ -754,10 +756,11 @@ private void UpdateViewCommands(bool testLoading = false)
             bool hasFailures = _model.HasResults && _model.ResultSummary.FailedCount > 0;
 
             _view.RunAllButton.Enabled =
-            _view.RunSelectedButton.Enabled =
             _view.DisplayFormatButton.Enabled =
             _view.RunParametersButton.Enabled = testLoaded && !testRunning;
 
+            _view.RunSelectedButton.Enabled = testLoaded && !testRunning && _model.SelectedTests != null && _model.SelectedTests.Any();
+
```
CharliePoole, line 763: Will it let you use `_model.SelectedTests?.Any()` here>

CharliePoole (approved):
Looks good to me. Thanks!

My one inline comment is at your option. If you want to change the tooltip, feel free to add another commit but check out my general comment about it first.

src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
rowo360: Unfortunately that's rejected by the c# compiler: 
Operator '&&' cannot be applied to operands of type 'bool' and 'bool?'

rowo360:
> Is a new event needed? I'm pretty sure we were already responding to changed selection. It might be a generic property change to our settings.

Yes indeed, there's already a similar event: the model has a property ActiveTestItem and any change of this property triggers a SelectedItemChanged event. The ActiveTestItem is set by the TreeViewPresenter whenever the selected tree node changes. But checking/unchecking tree nodes will not change the tree node selection and thus not change the ActiveTestItem. But instead only the `SelectedTests` property of the model is updated. So, I added this missing event to the `SelectedTests` property to comply to the pattern.
BTW: there's also no setting change event while checking/unchecking tree nodes

CharliePoole:
FYI, once you merge the branch is pretty useless so you can delete it.

Issue #1137, opened 2024-10-12
Show details about test duration


rowo360:
**Describe the bug**
It's currently possible to group all tests 'By Duration', so that the user get some insights about slow running tests. That's already fine!

It would be even more helpful for the user if can view the exact test duration somehow. So that he can identify some bottleneeks easily and and knows where to start an improvement.

The test duration is already provided by NUnit, but it's not visualized so far.

**Expected behavior**
The test duration is shown for each test case and test fixture in the test tree.

**Environment (please complete the following information):**
-  TestCentric Version - latest developer build
-  OS: Windows 11

rowo360:
I suggest adding the test duration to the end of the texts of the treenodes. Here's a screenshot:

<img src="https://github.com/user-attachments/assets/fb9b85fc-5231-431d-82d8-f72f247e8508" width="350">

The test duration is available for TestFixture and TestCase nodes.
When switching between the tree Display strategies or groupings, the entire tree will be rebuild, so that the tree node names will be set again.


Overall, I had the impression that this solution will work and is beneficial.
However when I tried out this solution for a while, I asked myself why it's limited to the 'Duration' grouping at all. Isn't it beneficial for the user to get insights to the test duration in whatever strategy/grouping?

So, I came up with an idea to add a new context menu item:
<img src="https://github.com/user-attachments/assets/bc68b0cb-a04f-4cc5-83a8-ad20f4c787ca" width="450">

As a result the test duration can be viewed in the NUnit tree too!
<img src="https://github.com/user-attachments/assets/96188aee-a7a3-41ef-8957-bdb40493b010" width="400">

In this view the test duration is also available for the 'Namespace' nodes - but that's also fine!
If you like the idea about showing the test duraction in all kind of strategies/grouping + the context menu approach, I will continue to work out this solution.

CharliePoole:
I like the idea. Two thoughts...

1. Is this a context menu item or a setting? Context menu items were intended to be transient, i.e. till you close the app. Settings were intended to be sticky. I think we have blurred that distinction somewhat by use of visual state, but let's think about it anyway.

2. Do we need a feature to sort the nunit tree by duration? I.e. a new top-level sort menu item?

Some of this will take a while to work out. So go ahead with it but don't do a PR till after the beta6 release, which is unfortunately taking me longer than I thought it would. 

Celebrating my 82nd birthday today, so I'll be back on Sunday. :-)

rowo360:
[...]

PR #1143, opened 2024-10-18
Show/hide test duration in test nodes

rowo360:
This PR close issue #1137 by displaying the test duration in the tree nodes.
The test duration is shown in square brackets behind the tree node name.

The user can turn on/off this feature using the tree context menu:

The current state (turned on/off) is saved in the settings, so that the last active state is restored when restarting TestCentric. The state is **not** stored in the VisualState file.

From technical point of view the TreeNodes are created by the `DisplayStrategy `class, therefore I placed the code there (the core of the implementation is done here). It handles these use cases:
- OnTestFinished: update tree node name of the finished test
- OnTestRunStarting: reset all tree node names - that's identical to our behavior regarding the tree node images. So that no outdated results from a previous run are shown.
- Turn on/off features: update all tree node names
- Switch display strategy/grouping: If feature is turned on, the new strategy/grouping will display test durations right from the start.

If no test results are available or the feature is turned off, the tree node names remain as before.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
@@ -81,12 +81,16 @@ namespace TestCentric.Gui.Presenters
         {
             int imageIndex = CalcImageIndex(result.Outcome);
             foreach (TreeNode treeNode in GetTreeNodesForTest(result))
+            {
                 _view.SetImageIndex(treeNode, imageIndex);
+                UpdateTreeNodeName(treeNode);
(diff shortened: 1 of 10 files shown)


rowo360:
@CharliePoole :
I'm aware that we are already focused on the next issues. But may I draw your attention to this PR? 
It would be great if we could finalize this before I start with the next issues and potential merge conflicts.
And some review is required as the next step - thanks a lot!

CharliePoole:
Right! I held off till after the release and then forgot it.

CharliePoole (approved):

Issue #1152, opened 2024-10-20
Re-introduce the ability to list fixtures without namespaces in the NUnit Tree


CharliePoole:
As discussed in #1146, we want this option in NUnitTree, similar to what was once available in the NUnit V2 GUI. 

It somewhat duplicates the purpose of the FixtureList strategy but that's acceptable especially as we will probably phase out the other strategies once we have everything we need in the NUnitTree display.

CharliePoole:
Closed by PR #1153

Issue #1151, opened 2024-10-20
NUnit Tree: fold namespace nodes into one single node


rowo360:
This is a sub-issue of issue #1146.

We believe it's beneficial to fold all Namespace nodes into a single tree node, if there is just a single sub-namespace under the namespace. This folding will improve the overview in the tree.

PR #1153, opened 2024-10-21
NUnit tree: add option to show/hide namespace nodes; add folding of namespace nodes

rowo360:
This PR solves issue #1151 and #1152 which both are dealing with the NUnit tree:
- list fixtures without namespaces
- fold namespace nodes

From user point of view there's a new submenu item which controls if the namespace nodes are shown or hidden in general.

This menu item is only enabled for the NUnit tree, but disabled if FixtureList or TestList mode is active.

Here's a screenshot: Namespaces are shown on left side, on right side they are hidden

In addition the namespaces are getting folded (if applicable). Here's an example:

From a technical point of view, several aspects had to be taken into account. 
The core of this issue is implemented in the class NUnitTreeDisplayStrategy: it's taking care about omitting namespaces and folding of namespaces. I decided not to change the base method `MakeTreeNode` which is used by all Strategies, but to add the implementation in the NUnitTreeDisplayStrategy class itself. So it's kept independent, avoiding any unintended side effects. However the recursive call to create the nodes for the children are done here, now.
One important line of code is how to detect a namespace reliable - thanks to your hint about SetUpFixtures, I consider this case as well.

When folding the namespaces, I decided to assign the 'deepest' of the folded TestNodes to the resulting TreeNode. All other folded TestNodes will not have a corresponding TreeNode. Some special handling was required for the TreeNode name of folded namespaces. Thanks to the new feature 'Show test duration', the tree node name must be adapted whenever 'Show test duration' is switched on/off and therefore it's not kept untouched after tree node creation. I had the impression that it's difficult to determine the folded name afterwards, so I decided to keep a list of all folded names in the NUnitTreeDisplayStrategy. That list can be used, whenever a tree node name is requested.

VisualState/Settings:
I implemented the ShowNamespace option identically to the Strategy or Grouping option: that means that it's stored in the settings and in the VisualState file. The VisualState file ensures that the same visual tree representation is restored whenever a test project is loaded. And the setting option ensures that the last selected option is applied when opening a project without VisualState file.

And I extended some tests for these new use cases.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
@@ -176,14 +176,19 @@ namespace TestCentric.Gui.Presenters
             return string.Format("{0} ({1})", group.Name, group.Count());
         }
 
+        protected virtual string GetTestNodeName(TestNode testNode)
+        {
+            return testNode.Name;
+        }
+
         private string GetTestNodeDisplayName(TestNode testNode)
         {
-            string treeNodeName = testNode.Name;
+            string treeNodeName = GetTestNodeName(testNode);
 
             // Check if test result is available for this node
             ResultNode result = _model.GetResultForTest(testNode.Id);
             if (_settings.Gui.TestTree.ShowTestDuration && result != null)
-                treeNodeName = testNode.Name + $" [{result.Duration:0.000}s]";
+                treeNodeName += $" [{result.Duration:0.000}s]";
 
             return treeNodeName;
         }
--- a/src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
@@ -136,7 +136,7 @@ namespace TestCentric.Gui.Presenters
 
         public TreeNode MakeTreeNode(TestNode testNode, bool recursive)
         {
-            string treeNodeName = GetTestNodeDisplayName(testNode);
+            string treeNodeName = GetTreeNodeDisplayName(testNode);
             TreeNode treeNode = new TreeNode(treeNodeName);
             treeNode.Tag = testNode;
 
@@ -176,14 +176,14 @@ namespace TestCentric.Gui.Presenters
             return string.Format("{0} ({1})", group.Name, group.Count());
         }
 
-        protected virtual string GetTestNodeName(TestNode testNode)
+        protected virtual string GetTreeNodeName(TestNode testNode)
         {
             return testNode.Name;
         }
 
-        private string GetTestNodeDisplayName(TestNode testNode)
+        private string GetTreeNodeDisplayName(TestNode testNode)
         {
-            string treeNodeName = GetTestNodeName(testNode);
+            string treeNodeName = GetTreeNodeName(testNode);
 
             // Check if test result is available for this node
             ResultNode result = _model.GetResultForTest(testNode.Id);
@@ -217,7 +217,7 @@ namespace TestCentric.Gui.Presenters
             if (testNode == null)
                 return;
 
-            string treeNodeName = GetTestNodeDisplayName(testNode);
+            string treeNodeName = GetTreeNodeDisplayName(testNode);
             _view.InvokeIfRequired(() => treeNode.Text = treeNodeName);
         }
 
--- a/src/TestCentric/testcentric.gui/Presenters/NUnitTreeDisplayStrategy.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/NUnitTreeDisplayStrategy.cs
@@ -8,7 +8,8 @@ using System.Windows.Forms;
 
 namespace TestCentric.Gui.Presenters
 {
-    using System.IO;
+    using System.Collections.Generic;
+    using System.Linq;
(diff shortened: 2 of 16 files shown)


CharliePoole:
FYI, I'm having trouble with this one, in part because there are two different changes and it's hard to pull it apart that way, but mostly because it's very difficult to visualize what it looks like and how it works. I think I'll pull it down and run it.

CharliePoole:
I built and ran it and now I see what you're doing. There are a few changes I'd like to see, which I'll get into the review after lunch. But here's are a few points in case you are up and about at this time and want something to do...

I think the UI for showing namespace as an additional choice under NUnitTree works. If we succeed in getting rid of the other two strategies, then we can put the option on the main menu as if it were a new strategy. But for the submenu, I'd like to see two radio buttons... Show Namespaces and Hide Namespaces. The verbs make it clearer I think.

For the folding operation, it all looks good except for the XML display. I think it should show the higher level rather than the deepest level. In this case, showing full info seems to trump consistency.

rowo360:
Thanks for your feedback - I'll adapt these parts 👍 
- Replace the single submenu item with these two submenu items you mentioned. And only one of them can be checked.
- Assign the top most TestNode to a folded Namespace node

But I'm already heading to bed now, so I'll start tomorrow :-)
And next time I'll try to prepare a smaller commit, so that the overview is easier!

src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
```diff
@@ -176,14 +176,19 @@ public string GroupDisplayName(TestGroup group)
             return string.Format("{0} ({1})", group.Name, group.Count());
         }
 
+        protected virtual string GetTestNodeName(TestNode testNode)
+        {
+            return testNode.Name;
+        }
+
```
CharliePoole, line 183: The name of this method is slightly confusing. Up till now, each tree node representeed only one test node. But now, with folding, it can map to multiple namespaces. So... each test node has a name and the tree node _may_ have a combined name. It would be great if this method and the next one could reflect that in some way. GetTreeNodeName and GetTreeNodeDisplayName?

src/TestCentric/testcentric.gui/Presenters/NUnitTreeDisplayStrategy.cs
```diff
@@ -18,6 +19,8 @@ namespace TestCentric.Gui.Presenters
     /// </summary>
     public class NUnitTreeDisplayStrategy : DisplayStrategy
     {
+        private IDictionary<TestNode, string> _foldedNodeNames = new Dictionary<TestNode, string>();
+
```
CharliePoole, line 23: Did you examine whether the base class dictionary, `_nodeIndex`, which is used for other purposes, could be expanded to serve the purpose of _foldedNodeNames? If not, we could look at it at a later time to decide whether it would simplify things.

src/TestCentric/testcentric.gui/Views/TestCentricMainView.cs
```diff
@@ -405,13 +408,24 @@ private void InitializeComponent()
...
             this.displayFormatButton.ToolTipText = "Tree Display Format";
+
+            // 
+            // nunitTreeShowNamespaceMenuItem
+            // 
+            this.nunitTreeShowNamespaceMenuItem.Name = "NUNIT_TREE_SHOW_NAMESPACE";
+            this.nunitTreeShowNamespaceMenuItem.Size = new System.Drawing.Size(198, 22);
+            this.nunitTreeShowNamespaceMenuItem.Tag = "NUNIT_TREE_SHOW_NAMESPACE";
+            this.nunitTreeShowNamespaceMenuItem.Text = "Namespace";
+
```
CharliePoole, line 419: Let's make the text "Show Namespace" so it's clear what it does. I'd actually prefer to see both Show Namespace and Hide Namespace exclusive radio buttons to make it completely clear, but I think we will redo this whole menu thing at least one more time so it could wait.

CharliePoole:
I couldn't figure out where to add this in the code, so I'll just make a general comment.

I think that the assignment of the deepest node to a folded tree node works in all situations except one: when we display the XML. In that case alone, I'd like to see the XML from the topmost node down. I realize that you have it set up to always use the same tree node for all purposes and this is a possibly messy complication. So if you prefer we can merge it the way it is and revisit at a later time.

rowo360:
Yes, right! This single menu item looks a little lost - using two menu items is much better.

rowo360:
I assign the top most TestNode of all folded Namespace nodes to the TreeNode, now. I'm really fine with this decision!
The 'XML display' now works as you expected.
Some tests start failing after this changes - but that's a good sign that those use cases are covered :-)

rowo360:
And I adapted those method names as you suggested:
`GetTreeNodeName` and `GetTreeNodeDisplayName`

rowo360:
I also had a quick look to the `_nodeIndex` Dictionary of the base class, which you mentioned.
It might be possible to use it also for getting the combined name of folded namespace nodes, but needs some more investigations. So I prefer to keep the current approach `_foldedNodeNames` for now.

I will start the commit now, and if everything is ok and succeeds, it's ready to merge from my point of view.

CharliePoole (approved):
This looks good... I'll merge

Issue #1148, opened 2024-10-19
Filter NUnit Tree Display by Outcome


CharliePoole:
As a design issue, we need to decide

* Level of selection: i.e. just success, failure, not run as we have been doing or should we support the full range of possible outcomes, including Inconclusive, Warning, Error? What about sub-levels?
* Hide vs dim unfiltered nodes in tree?
* What do we do if we have a fixture only failure and fixtures are not displayed? A namespace failure?
* How does the user activate the feature?
* What right clicks are available? What do they do? Corner cases, e.g. Right click and run failed fixture with some test failures and some successes under it. What happens?
* Do we do this one incrementally? I.e. simply first, more later?

rowo360:
Here are some considerations regarding the outcome filtering (some of these applies also to duration/category filtering).

I believe we need to perform the outcome filtering ourself, because that's not a kind of filtering which NUnit supports natively. I imagine that the filtering is done on TestNode level and as a result of the filtering each TestNode has a bool property assigned (for example IsVisible). When building up the tree, this property can be evaluated to decide if a node should be created or not. Whenever the filter changes, the entire tree is build up once again (`Strategy.Reload()`).

> Hide vs dim unfiltered nodes in tree?

I prefer to hide all filtered out nodes in the tree, so that I get a compact visualization of the tree displaying only those content which I'm currently interested in.

Generally speaking a TestNode will be visible if either it matches the filter itself or if any child matches the filter. So for example, if only all 'Passed' tests should be shown and a fixture contains failed and passed test cases, the fixture will be shown and all his passed test cases. And I would continue to show the fixture node as failed. That might be confusing on first sight, because I filtered for passed tests only. But the state of the fixture is nonetheless 'Failed' and starting to display anything different in the tree due to this filtering seems to increase complexity.

> What do we do if we have a fixture only failure and fixtures are not displayed? A namespace failure?

Currently we don't start with showing/hiding fixture nodes. But I expect that we would show the namespace failure in this case because the outcome state is propagated to the parent nodes.

> Level of selection

I propose to support the identical set of outcomes which is present as OutcomeGrouping:
"Failed", "Passed", "Ignored", "Inconclusive", "Skipped", "Not Run"
I didn't check it, but I expect that none of these require a special handling - but all of them can be handled identically. So there's no additional effort to support one more or less. What do you have in mind with sub-levels?

> How does the user activate the feature?

I propose to add a new DropDownMenuItem which contains all outcome filters:
[...]

PR #1155, opened 2024-10-24
Provide filter logic of TestNodes; starting with filtering by outcome

rowo360:
This PR is the first incremental step for issue #1148.
It provides the basic filter functionality of test nodes by supporting filtering by outcome. There's only one single use case supported yet: test results are already present from a test run and afterwards an outcome filter is applied programmatically. Then the test tree will display only the filtered nodes.

Further use cases will be supported in the next incremental steps:
- UI to configure a filter
- Apply active filter already during test run
- Apply commands on filtered nodes only (for example 'Run selected')

So, it's not the intention that this filter functionality is already perfect. But I believe it's a good foundation for the next steps.
These changes do not yet have any effect on the user as the filters cannot be activated, so this PR could be merged without causing any side effects.

If things go as planned, the new filter class can be extended to support also duration and category filtering - and maybe name filtering sometime in the future. And also the other way around: I hope that we won't have to adapt any other code locations if we extend the filters. This applies especially for the test tree: displaying a filtered tree and the filter functionality is well separated.

I tested the filter functionality by misusing one of the existing toolbar buttons. So for example setting the filter programmatically to:
`_model.TestCentricTestFilter.OutcomeFilter = new List<string>() { "Not Run", "Failed" };`

### DIFF
--- a/src/TestModel/model/TestNode.cs
+++ b/src/TestModel/model/TestNode.cs
@@ -33,6 +33,7 @@ namespace TestCentric.Gui.Model
 
         public TestNode(XmlNode xmlNode)
         {
+            IsVisible = true;
             Xml = xmlNode;
 
             // It's a quirk of the test engine that the test-run element does;
@@ -76,6 +77,10 @@ namespace TestCentric.Gui.Model
         public bool IsAssembly => Type == "Assembly";
         public bool IsProject => Type == "Project";
 
+        /// <summary>
+        /// Controls if the TestNode should be visible or hidden in the TestTree
+        /// </summary>
+        public bool IsVisible { get; set; }
         public int TestCount => IsSuite ? GetAttribute("testcasecount", 0) : 1;
         public RunState RunState => GetRunState();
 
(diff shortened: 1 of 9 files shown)


CharliePoole:
I'm a TDD guy, of course, so I'd like to see a few tests even if you are still experimenting. However I'm OK if you want to wait. If you add tests, I think we would only need one for the main presenter of the type "wnen the user changes the filter in the view the model is notified" and one for the tree view presenter "when the model changes the view is changed."

src/TestModel/model/TestNode.cs
```diff
@@ -76,6 +77,10 @@ public TestFilter GetTestFilter()
         public bool IsAssembly => Type == "Assembly";
         public bool IsProject => Type == "Project";
 
+        /// <summary>
+        /// Controls if the TestNode should be visible or hidden in the TestTree
```
CharliePoole, line 81: "In the NUnit Test Tree" right?

CharliePoole (approved):
This looks great for the purpose. My comments are all minor, so I'll merge and you look a them on the next time you pass. 

I realize IsVisible has to be an attribute of the test node because we don't even create a testnode if it's false. It's real name should be something strange like "ShouldDisplayInTree" and only applies to the NUnitTree strategy at the moment.  In future we might want to eliminate it as a property and make it a method of the tree display strategy or of the presenter itself, I'm not sure.

PR #1159, opened 2024-11-24
Update to latest TestCentric.Engine dev release

### LINKED ISSUE #1144: 2.0.0-beta6 force stop endless loop test - cleanup
Hello,

I forced to stop a endless loop in a test.
After it is sopped, I can execute the test again but I cannot stop it anymore because the stop button is grayed out.
In case I close the GUI, it tells me that a test is currently running.

The endless while loop is also preventing logs.
In case endless loop is removed logs are shown at text output:
"=> name1.name2.name3_1.name3_1.test1
test1test2test3"

Start and end time are wrong displayed in case of a forced stop.

Can be reproduced with attached example project.
[Solution1.zip](https://github.com/user-attachments/files/17431464/Solution1.zip)

Best regards,
Martin

rowo360:
This PR closes #1144 by updating the project reference of the TestCentric.Engine.
The details about the actual fix can be found in the PR [213 ](https://github.com/TestCentric/testcentric-engine/pull/213) to that project.

When using the latest version of the TestCentric.Engine, the Stop and Kill button works reliably. The behavior was checked using the hung-tests.dll for different .Net framework versions (4.62, Core 3.1, .Net 7.0...)

As this was my first time updating a project referecence, I was a little surprised at the version number of the dev releases. I initially thought that the highest number will be the latest dev release. But I notice that it's different here by checking the release dates. So I finally chose the version 2.0.0-dev00008. Is there any smart approach to detect the latest dev release differently?

### DIFF
--- a/src/TestModel/model/TestCentric.Gui.Model.csproj
+++ b/src/TestModel/model/TestCentric.Gui.Model.csproj
@@ -15,7 +15,7 @@
   
 	<ItemGroup>
 		<PackageReference Include="Mono.Options" Version="6.12.0.148" />
-		<PackageReference Include="TestCentric.Engine" Version="2.0.0-dev00006" />
+		<PackageReference Include="TestCentric.Engine" Version="2.0.0-dev00008" />
 		<PackageReference Include="TestCentric.Engine.Api" Version="2.0.0-beta6" />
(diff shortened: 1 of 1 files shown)


CharliePoole:
Good work. Sorry about the version issue. I normally manage the dev builds manually, deleting any are too early. But looking at the dates is the simplest way to be sure.

PR #1167, opened 2024-12-19
Fix TreeView flickering

### LINKED ISSUE #1165: Execute tests - TreeView is flickering
**Describe the bug**
I notice some annoying flickering of the test tree when executing a large amount of tests.
In my case the assembly contains 5000 unit tests which are running pretty fast.
First analysis shows up that this flickering issue is a side effect of the implementation for issue #1137 - which wants to show the test duration in test tree nodes.

**To Reproduce**
1. Open test assembly which contains several fast running tests (for example the attached one)
2. Run 'All tests'
=> Observe test tree flicking

[Sample test project.zip](https://github.com/user-attachments/files/18052712/MyLibraryLarge.Test.zip)

**Expected behavior**
I expect that there's no flickering.

- TestCentric Version: Current developer build, but I assume that this issue exists also in previous beta versions
- OS: Windows 11
- Target Runtime for your test assembly: NET Framework 4.8

rowo360:
This PR closes #1165 which reports a flicking in the Treeview when executing tests.

As mentioned in the issue description this bug is caused by the 'Show duration' implementation. This implementation updates a tree node name by adding the test duration as a suffix right away whenever a test finished.

I propose to solve this issue by doing one step back: instead of updating the tree node names on TestFinished event, we'll update them on the TestRunFinished event. That means that the TreeView update is triggered only once - which will solve the flickering.

Obviously this is one step back for the user as he cannot observe the test duration anymore during a test run, but needs to wait until the test run finish. Especially for long running test runs (for which test duration is most helpful), that's not the best solution.
However I didn't find any simple solution, which takes care about all use cases. And unfortunately WindowsForms TreeView triggers a TreeView refresh whenever a TreeNode name is changed.
So overall I want to fix this flickering effect first, which is annoying and hits all users regardless if 'Show duration' is activated or not.

Additional note:
If test run was executed previously and afterwards show test duration is activated, the test duration will be shown immediately - this is kept unchanged.

Some further ideas:
- Introduce a Timer based approach to update the Tree node names. So, the Tree node names are updated during a test run only every 10s (for example)
- Attention this idea is a major design switch - nonetheless I'm mentioning it.
We can introduce additional columns next to the tree view. One of these columns will show the 'Test duration'. So instead of updating the tree node names, we need to update the column cells. I hope that updating cells won't cause a flickering effect as well. But overall, it must be evaluated first, if such columns are beneficial at all and for which other information they can be used (for example categories)?
And it's worth to mention that WindowsForms doesn't support a 'TreeView with additional columns' out-of-the-box. But there are several examples in the web which might be helpful.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
@@ -84,18 +84,20 @@ namespace TestCentric.Gui.Presenters
             foreach (TreeNode treeNode in GetTreeNodesForTest(result))
             {
                 _view.SetImageIndex(treeNode, imageIndex);
-                UpdateTreeNodeName(treeNode);
             }
         }
 
         public virtual void OnTestRunStarting()
         {
             _view.ResetAllTreeNodeImages();
-            UpdateTreeNodeNames();
+            if (_settings.Gui.TestTree.ShowTestDuration)
+                _view.InvokeIfRequired(() => UpdateTreeNodeNames());
         }
 
         public virtual void OnTestRunFinished()
         {
+            if (_settings.Gui.TestTree.ShowTestDuration)
+                _view.InvokeIfRequired(() => UpdateTreeNodeNames());
         }
 
         // Called when either the display strategy or the grouping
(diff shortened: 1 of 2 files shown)


CharliePoole (approved):
I think what you have is a good solution that lets us move forward. I'm approving and merging. I'd like to consider alternatives at some point but not in a way to slow down progress on other things.

CharliePoole:
Alternatives I think we should consider in a future iteration...

1. Finding out why this happens at all so as to prevent it at the source. After all, it doesn't happen when we update the image for a test after it is run.
2. Using a timer as you suggested
3. Displaying the duration in some other way, e.g. in a tooltip or with an icon overalay.

I'm not really in love with the multi-column design.

PR #1169, opened 2024-12-22
Update tree node images - add transparent background

### LINKED ISSUE #1168: Update tree node icons
**Describe the bug**
I propose that we update the icons used in the tree view for the test outcomes.
Here's a screenshot to show the context: 

These icons are available in four different styles: 'Default', 'Circles', 'Classic' and 'Visual Studio'. All of them should be updated.

An update is necessary because most of these icons do not have a transparent background. That missing transparent background gets visible when using these icons in the outcome filter toolbar. (Introduced for feature #1163)

As a consequence the icons for the style 'Circles' and 'Classic' must be converted from jpg to png files, because transparency is only supported in png format.

And finally I noticed that some icons in the 'Circles' and 'Classic' style are not displayed correctly on a 4k monitor. I didn't notice it on other displays, so it might be related to scaling. But ultimately it doesn't look pretty.

**Expected behavior**
All these icons should use a transparent background and should be displayed correctly on all kind of displays (incl. 4K)

 - TestCentric Version 2.0Beta
 - OS: Windows 11

rowo360:
This PR fixes #1168 by updating the tree node images.
I separated the changes into 5 individual commits in order to improved the overview.


rowo360:
The first commit is actually the only code change. It's taking care that the images are not displayed with any artifacts on 4k monitors by setting the ColorDepth of the ImageList to Depth32Bit. We didn't set this property previously explicitly, but instead it's default value was applied which is Depth8Bit.
This default value (Depth8Bit) is also applied after the ImageList is initialized from an ImageStream in the designer code. So overall, I decided to remove this initialization from the ImageStream completly (+ the ImageStream in the resource file).
The method LoadAlternateImages is called anyway whenever the application is started and loads the images and initialize the ImageList. This happens in case a SettingFile exists or in case of the first startup in which no SettingsFile exists.
So I adapt the method LoadAlternateImages that it's initialize the ImageList now.

rowo360:
The 2nd commit is about adding transparent background to the images of the Style 'Default' and 'Visual studio'. I'm adding some screenshots now, because otherwise it is not possible to understand the differences.

I only needed to adapt the 'Ignored' image for the 'Default' style:


In the 'Visual Studio' style the transparent background was missing in several images, so I adapted them.
(Upper line: old images; Bottom line: new images)

rowo360:
The 3rd commit is a about adding transparent background to the images of the style 'Classic'.
Again here's the screenshot:


In addition I needed to change the file format from jpg to png to support the transparency.

rowo360:
The 4th commit is a about adding transparent background to the images of the style 'Circle'.
This time I adjusted the images by painting the circle with a uniform color. I have the impression that this makes the images look cleaner. Here's the screenshot:


Also this time I needed to change the file format from jpg to png to support the transparency.

rowo360:
The last commit is dealing with the Build.cake script.
It must be adapted to deal with the new png files instead of the jpg files for the 'Classic' and 'Circle' style.

CharliePoole:
Any reason not to use the same images for classic as for circles, just editing them to add the checkmark, etc.? We could consider that after we live with them for a while. I'd also like to check to see whether Microsoft has new versions of their downloadable images, which I used to create the Visual Studio style.

CharliePoole (approved):
Thanks for your work,  this looks great! I noted a few things for us to look at later on in the process, but I'm merging this one.

PR #1171, opened 2025-01-12
Disable TextFilter when no project is loaded

rowo360:
I have noticed one issue with the new TextFilter (#1161) that I would like to fix with this PR:

We actually want the filters to be disabled when no project is loaded. And that's how we implemented it already for the OutcomeFilter. 
However I missed to add this for the new TextFilter. This PR now makes up for this!
(File: DisplayStrategy.cs and NUnitTreeDisplayStrategy.cs)

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
@@ -76,6 +76,7 @@ namespace TestCentric.Gui.Presenters
         {
             ClearTree();
             _view.OutcomeFilter.Enabled = false;
+            _view.TextFilter.Enabled = false;
         }
 
         public virtual void OnTestFinished(ResultNode result)
--- a/src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
@@ -75,8 +75,7 @@ namespace TestCentric.Gui.Presenters
         public void OnTestUnloaded()
         {
             ClearTree();
-            _view.OutcomeFilter.Enabled = false;
-            _view.TextFilter.Enabled = false;
+            _view.EnableTestFilter(false);
         }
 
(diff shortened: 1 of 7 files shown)


rowo360:
While testing this functionality, I noticed another issue with the filters which doesn't work as expected:
We have added one button to show/hide the filter in the toolbar. This button should only be visible for the NUnit display strategy, but not for the Fixture/TestList display strategy.
That works when loading a test project and the strategy from the VisualState file is applied. But it won't work if the strategy is changed interactively - that's strange but I must have missed that one.

I fixed this issue by calling method `UpdateViewCommands` whenever the DisplayFormat changes. (File: TestCentricPresenter.cs)

src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
```diff
@@ -76,6 +76,7 @@ public void OnTestUnloaded()
         {
             ClearTree();
             _view.OutcomeFilter.Enabled = false;
+            _view.TextFilter.Enabled = false;
```
CharliePoole, line 79: Wouldn't it be cleaner for the view to have a single property to enable filters?

src/TestCentric/testcentric.gui/Presenters/NUnitTreeDisplayStrategy.cs
```diff
@@ -50,6 +50,7 @@ public override void OnTestLoaded(TestNode testNode, VisualState visualState)
                 SetDefaultInitialExpansion();
 
             _view.OutcomeFilter.Enabled = true;
+            _view.TextFilter.Enabled = true;
```
CharliePoole, line 53: Ditto.

CharliePoole (approved):
Looks good. If it's easy to do, I suggest a single view property to enable filters.

rowo360:
You are definitely right. This is much cleaner!
I'll add a new method void EnableFilterElements(bool enable) to the view.
Thank you for reviewing so carefully and suggesting improvements!

CharliePoole:
Great work!

PR #1173, opened 2025-01-15
Keep expand/collapse state of tree nodes when changing the tree filter

rowo360:
This PR is another contribution to the test filter functionality (#1148, #1161, #1147).

This PR propose to keep the expand/collapse state of the tree nodes constant, whenever a test filter is modified. For example if a user expanded certain tree nodes and afterwards apply a test filter, all those tree nodes which are still visible after the filtering should remain their expand/collapse state. 

The idea is that the user is familiar with the view after the filtering and does not have to adjust the expand/collapse state again right away. So overal this PR tries to improve the usability when working with test filters.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
@@ -102,12 +102,13 @@ namespace TestCentric.Gui.Presenters
 
         // Called when either the display strategy or the grouping
         // changes. May need to distinguish these cases.
-        public void Reload()
+        public void Reload(bool applyVisualState = false)
         {
             TestNode testNode = _model.LoadedTests;
             if (testNode != null)
             {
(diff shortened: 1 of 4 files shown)


rowo360:
I have the impression that I don't have to invent anything new for this functionality. Instead the expand/collapse state is already handled by the VisualState file, so I just want to apply it in this use case.

Whenever a test filter is modified, the `DisplayStrategy.Reload()` method is invoked. This method in turn calls the `OnTestLoaded(TestNode, VisualState)` method. So far the second parameter was set to null, which means that some default handling for expanding/collapsing tree nodes is applied.
Now I determine the VisualState first and pass it to the OnTestLoaded method - that's all to keep the expand/collapse state constant.

rowo360:
Just for the sake of completeness here are two screenshots showing the existing code when no VisualState is passed to the OnTestLoaded() method:

Issue #1149, opened 2024-10-19
NUnit Tree Display by duration


CharliePoole:
Design decisions...

* Do we need slow/medium/fast groupings or can this just be a continuous thing like: display all test fixtures taking more than 5000 ms? Does anyone care about the medium and fast ones anyway?
* Should we consider a sort as opposed to a filter?

PR #1179, opened 2025-02-08
Sort tree nodes by name or duration

rowo360:
This PR solves #1149 by adding a context menu that allows the tree to be sorted by duration.

Here's a brief overview of the functionality:
- the tree nodes can be either be sorted by name or duration
- it support ascending or descending order
- if duration sorting is selected, the 'Show duration' option is automatically enabled
- if duration sorting is selected, but no test durations are available at all, the testnodes are sorted by name
- if duration sorting is selected, but test durations are only available partly (subset of tests was executed), the name sorting is applied in general, but the subtree with duration results is sorted by duration.
- the sorting options are **not** stored in the settings or VisualState file
- however if a new project is loaded, the current active sorting options are applied to the newly loaded tree.
- The focus of this issue is the sorting in the NUnit tree display format. However the sorting works also with the Fixture and Test list display format.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/DisplayStrategy.cs
@@ -11,8 +11,6 @@ namespace TestCentric.Gui.Presenters
     using Model;
     using Model.Settings;
     using Views;
-	using Elements;
-    using System.IO;
     using System.Linq;
 
     /// <summary>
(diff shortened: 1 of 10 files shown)


rowo360:
The idea of solving this problem via a context menu was discussed already previously.
Another solution idea is introducing a duration filter.
For example: filtering by slow, medium or fast tests.

Thats idea can also be implemented well from a technical point of view. However it requires that the boundaries for slow, medium and fast tests must be adjustable by the user somehow. Because we do not expect to be able to define a value that fits all users equally well. Moreover there are concerns that users are really interested in showing only the fast running tests at all. 
Instead, the most important use case will be to quickly identify the long-running tests. And this can be easily achieved with the solution via the context menu.

rowo360:
The Windows Forms TreeView class already supports sorting functionality. It only requires a `IComparer` implementation for the comparision of tree nodes. Therefore I implemented 3 `IComparer` classes:

- Comparing the Fullname of a TestNode
- Comparing the Name of TestNode
- Comparing the Duration

rowo360:
One additional comment independent from this issue:
I noticed that the sorting is not as expected if you hide a namespace. Here's a screenshot:



When hiding the namespaces the tree nodes remain in the identical order, resulting that there's no alphabetical order anymore.

With this PR the tree nodes are sorted alphabetical now also when hiding the namespaces.

PR #1180, opened 2025-02-09
Add toolbar button to reset test filters

rowo360:
This PR contributes to the filter functionality: #1148, #1147 and #1161.

It adds a new button to the test filter toolbar to reset all filters at once.

As you would expect from a reset button, it clears all filters, refreshes the tree and the filter UI elements.
(For example: clears the Filter TextBox)

And having this reset functionality in place, it's also invoked when a project is unloaded or reloaded.


rowo360:
Here are some comments from a technical point of view:

I noticed that there are two different requirements when a filter is reset. If the reset is triggered by the user, the tree must be updated to reflect the new filter. However if the filter is reset, because of a project unload/reload, it's not required to update the tree - because the tree is cleared or rebuild anyway.
So overall, I decided to add a new parameter to the reset method:
`void ResetAll(bool suppressFilterChangedEvent = false)`

I just want to avoid that an unnecessary tree rebuild is invoked in case of project unload/reload.

rowo360:
The category filter dialog is designed as a non-modal dialog. So the reset might be triggered while the dialog is opened.
Therefore some additional implementation in class `CategoryFilterDialog` was required. If the 'SelectedItems' property is changed programmatically from a client, the dialog must update the checked items in the ListBox.

rowo360:
Also the textbox for the text filtering requires some adaptation to support the reset functionality. So far this TextBox uses the TextChanged event to start a timer and observe if no further input is made wihtin a short period of time. Now the TextBox uses the OnKeyUp event isntead of the TextChanged event for this purpose.

When reseting the filter, the Text of the Textbox must be set to an empty string programmatically, but it should not start the timer in this use case. So the OnKeyUp event better fits here.

rowo360:
I noticed that I missed one essential use case when reloading tests: the test categories may have changed!
Therefore I applied the same approach for reloading as for closing/loading project. The test category filter dialog gets closed first and afterwards initialized with the new set of available test categories.

PR #1181, opened 2025-02-09
Fix category filtering for "No category"

rowo360:
This PR fixes one issue with the category filtering in a special use case (#1147)

If categories have been defined for all methods of a TestFixture and now "No category" is activated for category filtering, the TestFixture will still be displayed in the tree, but without any children. Here's a screenshot:

It's expected that in this case also the TestFixture is filtered out and thus not be displayed in the tree anymore.

### DIFF
--- a/src/TestModel/model/Filter/CategoryFilter.cs
+++ b/src/TestModel/model/Filter/CategoryFilter.cs
@@ -39,7 +39,8 @@ namespace TestCentric.Gui.Model.Filter
 
         public bool IsMatching(TestNode testNode)
         {
-            if (_condition.Any() == false)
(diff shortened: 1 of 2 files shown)


rowo360:
Luckily this use case doesn't break the category filtering implementation. It's just an exceptional case which were not considered so far. Also all existing filtering tests continue to run successfully with the required adaptation. However tests were added to cover this use case.

In general the category filtering should consider only test-case nodes, but not TestFixtures nodes. The test-case nodes already determines all categories from itself and from its TestFixture to decide if the category filter matches.
And for TestFixture node the general filtering rule is applied: it will be only visible, if any of his child nodes is visible.

PR #1190, opened 2025-02-16
TestCentricProject - revisit IsDirty flag to reduce number of prompts to save changed project

### LINKED ISSUE #1187: Do you want to Save MessageBox pops up too frequently
**Describe the bug**
Whenever a project was modified and is getting closed, we prompt the user and ask if he wants to save his changes.
However I've the impression that this MessageBox shows up also in use cases in which the project was not modified.
For example the message box is shown in these use cases:
- Load an existing tcproj file and close it
- Load one file from the recent files list (dll or tcproj)

**To Reproduce**
1. Load a tcproj file
2. Close project
3. Messagebox "Do you want to Save?" shows up

**Expected behavior**
Overall I think that we show this MessageBox only if the user actively modified the test project. For example, if he created a new project or if he added a test file to a project.

 - TestCentric Version:2.0.0-beta6
 - OS: Windows 11

rowo360:
This PR fixes #1187 by reducing the use cases in which the "Do you want to save?" MessageBox show up.

The MessageBox is shown now only in these two use cases:
- The user actively created a new project
- The user actively added a test to an existing project

In contrast the MessageBox is not shown anymore, if a user loads an existing tcproj file. Or if a test is opened from the recent file list - this includes the use case if the last recent file is opened when starting the application.

I think that we don't need to ask the user again in the later use case. He already made his decision in his previous session and it seems to be bit intrusive if we ask over and over again.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
@@ -653,7 +653,10 @@ namespace TestCentric.Gui.Presenters
         {
             var files = _view.DialogManager.SelectMultipleFiles("New Project", CreateOpenFileFilter());
             if (files.Count > 0)
-                _model.CreateNewProject(files);
+            {
+                _model.CreateNewProject();
+                _model.AddTests(files);
+            }
         }
 
         private void OpenProject()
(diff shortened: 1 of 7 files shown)


rowo360:
From technical point of view I made these essential decisions:
- I want to continue that the `TestCentricProject` class determines itself the IsDirty flag - instead of that it can be set from outside.
- Therefore the IsDirty flag is initially set to false, after constructing a TestCentricProject object.
- Also if a project (tcproj) is loaded, the IsDirty flag is set to false
- Only when a TestCentricProject is modfied, the IsDirty flag is set to true. 
  Modification means that a test file is added to the project

The code for the use cases 'Create new project' and 'Add test' are adapted to call the new method `AddTests `so that the IsDirty flag is set accordingly.

rowo360:
While working on this issue I used multiple test files in one project for the first time. No doubt - that makes sense, but it was out of my focus so far. I think I need to have a closer look to this functionality and check how certain use cases will work.
For example, I wonder how a user can remove a test file from a project? Or how the VisualState files will be applied?

Issue #1193, opened 2025-02-24
Improve the overview of test results (part 2)


rowo360:
**Describe the bug**
This issue is a follow-up issue of #1191 which is targeting to improve the overview of the test results.
With the previous work, all information is now available on the Test Result Tab. However, this tab is now too crowded. Here's a screenshot:

<img src="https://github.com/user-attachments/assets/46ff814c-6ede-4b6d-b27a-17cebc34babf" width="800">

I propose these enhancements:

- Show 'TestOutputSubView' only if selected test contains any output
- Show DetailList and SourceCode/ErrorBrowser only if a test case is selected (in all other cases the content is anyway empty)
- Reduce height of DetailList ListBox in case there's only one single assertion
- Adapt visualization of 'TestResultSubView'
  - Remove Message label (it's unused) and Assertion information (info is also available in detailList)
  - Show number of passed/failed/notRun .. tests if a TestFixture/TestSuite/TestAssembly node is selected
  - If test case node is selected, minimize the required information to save space

rowo360:
Here are some final screenshots from the solution:

The elements shown in the "test result" tab are reduced  to display only the relevant information.
For example if a 'test-suite' test node is selected in the tree node, the TestResultSubView shows a list of all kind of test types existing in the test result.
<img src="https://github.com/user-attachments/assets/32622284-a29c-4fdd-9078-b3425913d653" width="700">

rowo360:
This list might be larger or smaller depending on the tests. Here's another example:

<img src="https://github.com/user-attachments/assets/67a1a5f3-6a21-4538-a607-f348a5602012" width="700">

rowo360:
This applies also to nodes which contain no test results.
Please note also, that in all these cases so far, the ErrorBrowser and DetailList is not shown yet. They contained no information in these use cases in the past, so they are hidden now.

<img src="https://github.com/user-attachments/assets/2f9ec5fe-6928-4845-9219-97e18235f037" width="700">

rowo360:
The ErrorBrowser and DetailList are shown if a test-case node with an error is selected.
Both these elements are kept unchanged.
The information in the TestResultSubView is minimized to save space.

<img src="https://github.com/user-attachments/assets/1aa6ec06-d68b-4018-87c5-548ecbf8ff35" width="700">

rowo360:
In case there are multiple assertions all of them all displayed in the DetailList - that's also unchanged - I just mentioning it and show a screenshot.

[...]

PR #1194, opened 2025-03-01
Issue 1193

rowo360:
This PR solves #1193 by adapting the visualization of the TestResultView.

I already added some screenshots in issue #1193 to get an impression about the new user experience.

Technical view:
I separated the changes in several individual commits to get a chance for reviewing.
Most of the changes are related to UI elements. I comment on some parts next

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/ErrorsAndFailuresPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/ErrorsAndFailuresPresenter.cs
@@ -176,7 +176,9 @@ namespace TestCentric.Gui.Presenters
 
         private void InitializeTestOutputSubView()
         {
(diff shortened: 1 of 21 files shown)


rowo360:
I decided to introduce a new presenter class: `TestResultSubViewPresenter` 
That one is responsible for the logic of view `TestResultSubView`. So far this view was controlled along by the `ErrorsAndFailuresPresenter`.
It doesn't contain any complex logic, but by introducing also interfaces for these classes it's possible to create some tests for these use cases.

rowo360:
I introduced a new class `TestResultCounts` to determine the number of 'Failed', 'Passed'... tests of a selected TestNode.
Instead of adding this new class, I started with using the existing class `TestResultSummary` - which has a similar intention. However I failed to utilize this class for my purpose, because the numbers were not accurate in my use case.
So I came to the conclusion that either I need to adapt the TestResultSummary class or add a new one.
I decided for the second option to avoid any negative side effects.

PR #1208, opened 2025-03-19
Menuitem for 'Display format NUnit': simplify 'show namespace' submenu

### LINKED ISSUE #1204: NUnitTree Show/Hide Namespace can be simplified.
I like how NUnitTree is checkable but also shows the show/hide namespace options, but I think it could use a few improvements...

1. Suppress displaying the submenu on hover if NUntiTree is not checked. If the user checks it, then show the submenu. This, of course, means overriding some default behavior of Windows Forms.
2. Instead of two items in the submenu for show and hide, just have one item with the checkmark on or off. Label can be Show namespaces.

rowo360:
This PR fixes #1204 by replacing the two submenu items "Show namespace" and "Hide namespace" with one single submenu item.

In addition this submenu is only available, if the NUnit display format is checked. If another display format is checked, the submenu is not available.

If a display format is selected, the menu is not closed automatically anymore. Instead it's kept open, so that the user can select the appropriate submenu item right away.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
@@ -272,7 +272,7 @@ namespace TestCentric.Gui.Presenters
                         _view.GroupBy.SelectedItem = _settings.Gui.TestTree.FixtureList.GroupBy;
                         break;
                     case "TestCentric.Gui.TestTree.ShowNamespace":
-                        _view.ShowNamespace.SelectedIndex = _settings.Gui.TestTree.ShowNamespace ? 0 : 1;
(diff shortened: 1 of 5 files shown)


rowo360:
From technical point of view:
we can use the ToolStripMenuItem event `Closing` to suppress that the menu is automatically closed when an item is clicked. The event parameter provides some information about the closing reason to act accordingly.
This applies for the entire 'Display format' menu item now and not only the 'NUnit' menu item. But as we plan to adapt the 'Fixture list' and 'Test list' menu item too, that's also fine.
BTW: although the menu is not closed automatically, the tree view is updated to the newly selected display format right away.

rowo360:
I decided to keep the entire logic for adding/removing the submenu in the View class. From my point of view that's a pure UI part, so involving the Presenter class for this step seems to be an overhead.

PR #1209, opened 2025-03-20
Add menu item to remove test subpackage from test project

### LINKED ISSUE #1202: Remove test file from test project
**Describe the bug**
The current system doesn't support to remove a test file from a test project. We only support to add test files.
Here's a screenshot in which I accidently added a wrong test assembly. But I cannot remove it now using the UI.

**To Reproduce**
1. Create new test project
2. Add multiple test files to project
=> Try to remove one test files

 - TestCentric Version: Current developer build, but I assume that this issue exists also in previous beta versions
 - OS: Windows 11

rowo360:
With this PR it is now possible to remove a test package from a test project.
The user can trigger this functionality either by a context menu entry or by a keyboard (delete/backspace key).

This menu item is only available for tree nodes of type `IsAssembly`. For all other types of tree nodes (TestFixture, Test case ...) it's not visible. This operation is also not available during a test run. 

Moreover this operation is also not available if the test project contains only one test package. That might not be necessary, but I don't want to introduce a new state - a test project without any test package. But I preferred to be cautious here.

This PR close #1202.

### DIFF
--- a/src/TestCentric/testcentric.gui/Elements/ControlElements/KeyCommand.cs
+++ b/src/TestCentric/testcentric.gui/Elements/ControlElements/KeyCommand.cs
@@ -1,0 +1,45 @@
+// ***********************************************************************
+// Copyright (c) Charlie Poole and TestCentric contributors.
+// Licensed under the MIT License. See LICENSE file in root directory.
+// ***********************************************************************
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Windows.Forms;
+
+namespace TestCentric.Gui.Elements
+{
+    /// <summary>
+    /// This class is responsible to handle key events of a control
+    /// </summary>
+    internal class KeyCommand : IKeyCommand
+    {
+
+        public event CommandHandler KeyUp;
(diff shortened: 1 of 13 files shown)


rowo360:
One point of interest might be the new interface `IKeyCommand `and associated new class `KeyCommand` in Elements folder. That's not exactly a view or control element likes all other classes in this folder. But I think it has the same purpose of hiding the Windows Forms details from the Presenter classes. In this case it's the keyup/keydown event, but also all those Windows Forms keycodes. So, I think it fits to this folder fine.

There are actually many more details to keyboard handling: ctrl, alt key pressed or other key modifiers. I omit all these cases for now, because I was only interested in simple key events.

rowo360:
I managed to remove the subpackage from the test project by using the SubPackages collection - as you suggested:
`SubPackages.Remove(subPackage);`

Otherwise, the remove process is the same as the add process: after the package is removed, the current test project is unloaded and loaded again. That might not be optimal for performance reasons, but results into a consistent state.

Issue #1205, opened 2025-03-16
Eliminate FixtureList Group by Assembly


CharliePoole:
This is redundant with NUnitTree, per our discussion.

PR #1211, opened 2025-03-22
Add submenu for grouping to 'Fixture list' and 'Test list' menu item

### LINKED ISSUE #1206: Change dropdown submenus for FixtureList, TestList to use same approach as for NUnitTree
That is, if TestList or FixtureList is checked, a submenu shows the availble GroupBy items. This has two big advantages:
1. Only the relevent items appear, so none have to be disabled.
2. The user can click to select the TestList/FixtureList and then click the grouping rather than having to bring up the menu twice.

rowo360:
This PR closes #1206 by rearranging the grouping submenu items.
As proposed in the related issue, the grouping items are placed now as submenu items of the 'Fixture list' and 'Test list' menu item. This means that the DisplayFormat menu now only has 3 entries - here's a screenshot:

The 'Fixture list' supports only three groupings: duration, category and outcome. The assembly grouping is removed as proposed by issue #1205. Also the fixture grouping is not shown anymore - previously it was disabled for 'Fixture list'.

The 'Test list' still supports all kind of groupings. Here's a screenshot:

These submenu items are only available if the specific display format is checked, otherwise there's no submenu for that display format. Moreover the display format menu is not getting closed automatically, if a display format is selected. Instead the submenu is shown, so that the user can immediately select his prefered grouping.

### DIFF
--- a/src/TestCentric/testcentric.gui/Elements/ToolStripElements/CheckedToolStripMenuGroup.cs
+++ b/src/TestCentric/testcentric.gui/Elements/ToolStripElements/CheckedToolStripMenuGroup.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) Charlie Poole and TestCentric contributors.
 // Licensed under the MIT License. See LICENSE file in root directory.
 // ***********************************************************************
@@ -190,7 +190,7 @@ namespace TestCentric.Gui.Elements
 
         public void InvokeIfRequired(MethodInvoker _delegate)
         {
-            if (ToolStrip.InvokeRequired)
+            if (ToolStrip != null && ToolStrip.InvokeRequired)
                 ToolStrip.BeginInvoke(_delegate);
             else
                 _delegate();
--- a/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
@@ -964,18 +964,10 @@ namespace TestCentric.Gui.Presenters
 
             switch (displayFormat)
             {
-                case "NUNIT_TREE":
-                    _view.GroupBy.Enabled = false;
-                    break;
                 case "TEST_LIST":
-                    _view.GroupBy.Enabled = true;
                     _view.GroupBy.SelectedItem = _settings.Gui.TestTree.TestList.GroupBy;
                     break;
                 case "FIXTURE_LIST":
-                    _view.GroupBy.Enabled = true;
-                    // HACK: Should be handled by the element itself
-                    if (_view.GroupBy is Elements.CheckedToolStripMenuGroup menuGroup)
-                        menuGroup.MenuItems[1].Enabled = false;
                     _view.GroupBy.SelectedItem = _settings.Gui.TestTree.FixtureList.GroupBy;
                     break;
             }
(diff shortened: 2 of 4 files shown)


rowo360:
I briefly considered whether I should create separate grouping menu items for 'Fixture list' and 'Test list' in the view. And also whether I should create two separate CheckedToolStripMenuGroup instances in the view. But finally I didn't notice any essential difference, so I kept it unchanged.

src/TestCentric/testcentric.gui/Elements/ToolStripElements/CheckedToolStripMenuGroup.cs
```diff
@@ -190,7 +190,7 @@ public void EnableItem(string tag, bool enabled)
 
         public void InvokeIfRequired(MethodInvoker _delegate)
         {
-            if (ToolStrip.InvokeRequired)
+            if (ToolStrip != null && ToolStrip.InvokeRequired)
```
CharliePoole, line 193: Is it even possible to create a an instance without a ToolStrip?

CharliePoole (approved):
Looks good. I assume you'll merge in the change I made to add UNGROUPED.

src/TestCentric/testcentric.gui/Elements/ToolStripElements/CheckedToolStripMenuGroup.cs
rowo360: Yes, that's a new state now. These submenu items (byCategory, byOutCome...) are created in method `InitializeComponent`, but are not added as a child to any control yet. Instead this is done dynamically now, whenever the FixtureList or TestList menu items is getting checked. At that point in time they are added as submenu items either to the FixtureList or TextList menu item.

So these submenu items are parentless from Windows Forms perspective at this point in time and the CheckedToolStripMenuGroup class fails to retrive a ToolStrip (parent) from that submenu items.

There are alternative solutions:

- We can pass in a ToolStrip instance. For example by adding a new constructor in class CheckedToolStripMenuGroup
- Instead of dynamically adding the submenu items, we are duplicating them so that both menu items FixtureList and TextList have their own dedicated set of submenu items. We can set the visbility of the submenu items according to the checked state of the parent menu item, so that the submenu is hidden completely.

I think I will try out the first approach after merging the 'ungroup' feature.

CharliePoole: I tend to avoid doing anything dynamically unless it's actually necessary, as when the content of the menu changes based on events that only happen at runtime. In this case, in the original menu before those changes, I shared the items and disabled one of them but that was just me being lazy. :-)

I'd consider your second option, which only requires copying some code and prevents any crosstalk between the two strategies. Another principle I like is to keep the view very very simple because the view itself is barely tested at all. 

But you're the guy doing it so I'm OK with whichever approach you take.

rowo360: Ok, good points - since I've always thought about this myself, but without coming to a final conclusion, I'll just give it a try. If the result looks simple, I'll abondon this PR and create a new one. :-)

CharliePoole: Assuming the changes are small, thee's no need to abandon the PR. Once you add a commit, both the feature branch and the pr will update. The only time you need to abandon a PR is if the original starting point (in main) at the time you created it changes.

rowo360:
It was now easier for me to create a new branch with a new PR. (#1216)
So I close this PR without a merge now.

PR #1213, opened 2025-03-24
Add UNGROUPED uption to FixtureList and TestList

### LINKED ISSUE #1207: Should FixtureList/TestList have an "ungrouped" option?
Currently, the user is required to select some sort of grouping. A straight, ungrouped list would have advantages in some situations. For example the user could sort the list any way desired.

@rowo360 What do you think? I'm making this a question, but it would change to a feature if we want to do it.

CharliePoole:
Fixes #1207 

Note that some issues with testing arose and I added but then commented out "UNGROUPED" in a test case for that reason. I think there is some issue with how we are using the default settings for both strategy and  groupid. I'll file an issue with my questions.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/FixtureListDisplayStrategy.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/FixtureListDisplayStrategy.cs
@@ -51,6 +51,11 @@ namespace TestCentric.Gui.Presenters
             switch (groupBy)
             {
                 default:
+                case "UNGROUPED":
+                    foreach (TestNode fixture in GetTestFixtures(testNode))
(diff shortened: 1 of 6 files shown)


rowo360:
That looks great - as you said - the implementation is really simple and straightforward.
Here's only one remark: the new menu item is 'Simple list', isn't it? As an alternative I can also imagine: "No grouping" or "Ungroup". But please use the term that best fits as a native speaker.

rowo360:
BTW: I expect some merge conflicts with open PR #1211. But I'll handle this. 
I'll wait until this PR is merged and will update my changes so that the new menu item is also moved to the submenu.

rowo360 (approved):

PR #1216, opened 2025-03-27
Add submenu for grouping to 'Fixture list' and 'Test list' menu item

### LINKED ISSUE #1205: Eliminate FixtureList Group by Assembly

### LINKED ISSUE #1206: Change dropdown submenus for FixtureList, TestList to use same approach as for NUnitTree

rowo360:
This PR rearrange the grouping submenu items.
As proposed in the related issue (#1206), the grouping items are placed now as submenu items of the 'Fixture list' and 'Test list' menu item. This means that the DisplayFormat menu now only has 4 entries - here's a screenshot:

The 'Fixture list' supports only four groupings: simple list, duration, category and outcome. The assembly grouping is removed as proposed by issue #1205. Also the fixture grouping is not shown anymore - previously it was disabled for 'Fixture list'.

The 'Test list' still supports all kind of groupings. Here's a screenshot:

These submenu items are only available if the specific display format is checked, otherwise there's no submenu for that display format. Moreover the display format menu is not getting closed automatically, if a display format is selected. Instead the submenu is shown, so that the user can immediately select his prefered grouping.

This PR closes #1205 and closes #1206.

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
@@ -260,10 +260,10 @@ namespace TestCentric.Gui.Presenters
                         UpdateViewCommands();
                         break;
                     case "TestCentric.Gui.TestTree.TestList.GroupBy":
-                        _view.GroupBy.SelectedItem = _settings.Gui.TestTree.TestList.GroupBy;
+                        _view.TestListGroupBy.SelectedItem = _settings.Gui.TestTree.TestList.GroupBy;
                         break;
                     case "TestCentric.Gui.TestTree.FixtureList.GroupBy":
-                        _view.GroupBy.SelectedItem = _settings.Gui.TestTree.FixtureList.GroupBy;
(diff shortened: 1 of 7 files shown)


rowo360:
Based on our discussion on PR #1211 I decided to restart the implementation on a new workspace and so it's also resulting in a new PR. Instead of dynamically assigning the submenu items to either the 'TestList' or 'FixtureList' menu item, each menu item has its own dedicated submenu items now.
I favour this approach over my first attempt because it's using the standard Windows Forms design: each Control is assigned to a parent - always. And we don't need to take care about any exceptional cases.

The number of changed lines might be larger compared to my first approach (more difficult to review), but most of the lines are related to the newly added controls. And I also rename some controls to differentiate between 'byCategoryMenuItem' for testlist and fixture list.

So the essentail part is that the interface `IMainView` contains two CheckedToolStripMenuGroup items now:
- ISelection TestListGroupBy { get; }
- ISelection FixtureListGroupBy { get; }

PR #1217, opened 2025-03-28
Remove and replace functionality of TestRunSummary display

### LINKED ISSUE #1212: Improve test run summary
**Describe the bug**
In this issue, an enhancement is proposed to improve the test result summary. The proposal is a mix of the summary in TestCentric 1.6 with some new elements.

But first a few thoughts on why an improvement is helpful - here's a screenshot of a test run summary:

- My biggest point is that we are presenting a lot of information here. A lot of textual information which must be read and understood by the user.
- I can imagine that many users are overwhelmed by this amount of information and just skip it. Perhaps just because we have the information available, we shouldn't present everything at once.
- All this information requires some space to present.
- The show/hide summary button helps to reduce the space. But I've doubts if users will recognize this button at all.
- Unfortunately not all information of the summary is visible. The last line is shown only partly (see screenshot).
- The summary gets hidden when starting a test run and shown again when the test run finished. This makes the UI a bit restless for me - especially for quick running test. (some UI elements disappear and two seconds later they are shown again).

**Expected behavior**
Overall, I propose that the user can understand the test result summary in a glance. And yes indeed, that means I propose to present less information. Of course, that's a tradeoff - some users might miss some information. And I expect that we'll discuss which part is most important. But nonetheless I think currently we present too much information.

I already made one attempt to improve the test run summary. So I can explain what I've in mind - first one screenshot and in the comments below some explanations.

That's the test run summary of the same test - but I think this can be understood more quickly.

CharliePoole:
Fixes #1212


CharliePoole:
Continuing work on this as a work in process....

First commit: Entirely remove the TestRunSummary display and the button that displays/hides it.

Second Commit: Modify the status bar to use icons rather than text. Add a stopwatch icon for duration. Eliminate test count and test runs.

NOTE: Test count and test runs were always equal in all my tests. Both can be deduced by adding up all the panels that are displayed.

CharliePoole:
@rowo360 There's more that could be done, but I'm running short of time and I'd like to get what's done merged so we are both using the same GUI as we work.

rowo360 (approved):
That's all fine - I'm curious to see the statusbar in action :-)

PR #1219, opened 2025-04-01
Centralize ImageSet management for views that need it

### LINKED ISSUE #1218: Image set selection for icons used in tree and elsewhere
While working on #1212, I have encountered some roadblocks around the image sets we use for icons. This issue documents the problems I have seen. We could incorporate the needed changes in the work I'm doing for #1212 or work this in a separate PR before continuing with #1212. 

Problems I see...
1. Some of the image sets are out of date and no longer seem very useful.
2.  Because of features we have added, the icon image sets are now loaded in four different places in the code.
3. There is no Warning icon, although the Ignored icon is in the form of a warning.
4. Some of the icons themselves could be improved.
While those are separate problems, they are probably worked more efficiently in one PR. Details...

**Image Sets** I propose removing the "Circles" image set and combining "Classic and Default" into one called "Classic". The two are very similar and "Default" isn't really a name, it's just the one we use if no other is selected.

**Centralization** We should create a central class that manages image sets and loads the current set only once. Each element that needs the icons should get them from that class. The location of that class should reflect the fact that the images are no longer only used in the tree but in other parts of the GUI as well. We could also change the settings so that the selected image set is under Gui rather than Gui.TestTree, although this is a minor consideration.

**Warning Icon** Each image set we retain should have a Warning icon. There are two choices regarding Ignored... (1) We can keep it as a separate icon, in which case  the two icons must be different in appearance. (2) We can give the user a choice to display Ignored tests as a warning or as skipped. This would add a small amount of complexity but reflects the fact that many some users consider Ignored as a neutral result, whereas the original NUnit developers considered it as a bad thing to be highlighted in some way.

**Icon improvements** Classic could be made to look more professional. Visual Studio doesn't actually reflect the current icons used in VS 2022 but a much older version.

I lean towards postponing #1212 and working this separately since there are some decisions involved. @rowo360 What do you think?

CharliePoole:
Fixes #1218

### DIFF
--- a/src/TestCentric/testcentric.gui/OutcomeImageSet.cs
+++ b/src/TestCentric/testcentric.gui/OutcomeImageSet.cs
@@ -1,0 +1,62 @@
+// ***********************************************************************
+// Copyright (c) Charlie Poole and TestCentric contributors.
+// Licensed under the MIT License. See LICENSE file in root directory.
+// ***********************************************************************
+
+using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using TestCentric.Engine;
+
+namespace TestCentric.Gui
+{
+    /// <summary>
+    /// An ImageSet holds a group of images used as icons in various parts of the GUI.
+    /// </summary>
+    public class OutcomeImageSet
+    {
+        private static readonly string[] REQUIRED_FILES = new[] {
+            "Inconclusive.png", "Success.png", "Failure.png", "Ignored.png", "Skipped.png" };
+
+        private string _imageSetDir;
+
+        public OutcomeImageSet(string imageSetDir)
+        {
+            Guard.ArgumentValid(IsValidImageSetDirectory(imageSetDir), $"Directory {imageSetDir} does not contain an image set.", nameof(imageSetDir));
+
+            _imageSetDir = imageSetDir;
+            
+            Name = Path.GetFileName(imageSetDir);
+        }
+
+        public string Name { get; }
+
+        public static bool IsValidImageSetDirectory(string dir)
+        {
+            foreach (string file in REQUIRED_FILES)
+                if (!File.Exists(Path.Combine(dir, file)))
+                    return false;
+
+            return true;
+        }
+
+        // Counter used for testing
+        public int LoadCount { get; private set; } = 0;
+
+        public Image LoadImage(string imgName)
+        {
+            if (_images.ContainsKey(imgName))
+                return _images[imgName];
+
+            LoadCount++;
+            
+            return _images[imgName] = Image.FromFile(Path.Combine(_imageSetDir, imgName + ".png"));
+        }
+
+        private Dictionary<string, Image> _images = new Dictionary<string, Image>();
+    }
+}
--- a/src/TestCentric/testcentric.gui/Presenters/ImageSetManager.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/ImageSetManager.cs
@@ -1,0 +1,76 @@
+// ***********************************************************************
+// Copyright (c) Charlie Poole and TestCentric contributors.
+// Licensed under the MIT License. See LICENSE file in root directory.
+// ***********************************************************************
+
+using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using TestCentric.Engine;
+using TestCentric.Gui.Model;
+using TestCentric.Gui.Views;
+
+namespace TestCentric.Gui.Presenters
+{
+    public class ImageSetManager
+    {
+        private static readonly string BASE_DIR = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, Path.Combine("Images", "Tree"));
+
+        private ITestModel _model;
+        private IMainView _view;
+
+        public ImageSetManager(ITestModel model, IMainView view)
+        {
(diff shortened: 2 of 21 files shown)


CharliePoole:
@rowo360 I merged this so I can get back to working on #1212, which it is blocking. Even so, an after-the-fact review would be helpful and I can incorporate any changes into the issue- branch 1212.

After thinking it over, I decided to have the main presenter hold the `ImageSetManager` and for the manager to basically work like an adjunct to the presenter, calling methods on the various components that need to refresh their images when a change occurs.

Currently, it is providing images for the `TestTreeView`, `TestResultSubView` and `TreeSettingsPage`. With the changes in #1212, I'll add the `StatusBarView` as well.

src/TestCentric/testcentric.gui/SettingsPages/TreeSettingsPage.cs
```diff
@@ -30,11 +31,14 @@ public class TreeSettingsPage : SettingsPage
         private static readonly string[] imageExtensions = { ".png", ".jpg" };
 
         private static string treeImageDir = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, Path.Combine("Images", "Tree"));
```
rowo360, line 33: Is it possible that this member is not required anymore?
Same question regarding member imageExtensions?

src/TestCentric/testcentric.gui/OutcomeImageSet.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            if (_images.ContainsKey(imgName))
+                return _images[imgName];
+
+            LoadCount++;
+            
+            return _images[imgName] = Image.FromFile(Path.Combine(_imageSetDir, imgName + ".png"));
+        }
+
+        private Dictionary<string, Image> _images = new Dictionary<string, Image>();
```
rowo360, line 60: Don't know if it's best practice or I'm only used to it: should all members listed at begin of class?

src/TestCentric/testcentric.gui/OutcomeImageSet.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using TestCentric.Engine;
+
+namespace TestCentric.Gui
+{
+    /// <summary>
+    /// An ImageSet holds a group of images used as icons in various parts of the GUI.
+    /// </summary>
+    public class OutcomeImageSet
```
rowo360, line 20: Was it intentional that this class is located in the main folder? And not in the Presenters subfolder?

rowo360, line 20: And just another remark: we could add an interface IOutcomeImageSet.
So that the ImageSetManager is the only one using the concrete class, but all usages only dependent on the interface. Just to enfoster testing and to loosing the coupling - well, we are barely testing the view code. But I'm just mentioning it.

rowo360 (commented):
Looks fine - I added some minor comments.
I think you mentioned the most interesting already in a code comment yourself: the ImageSetManager should use an event driven approach.

src/TestCentric/testcentric.gui/SettingsPages/TreeSettingsPage.cs
CharliePoole: Right. Both gone in #1212 now.

src/TestCentric/testcentric.gui/Views/ITestResultSubView.cs
```diff
@@ -18,7 +18,7 @@ public interface ITestResultSubView
         /// <summary>
         /// Load test outcome images 'Passed', 'Failed'...
         /// </summary>
-        void LoadImages(string imageSet);
+        void LoadImages(OutcomeImageSet imageSet);
```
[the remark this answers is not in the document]
CharliePoole: Yes. I actually want to standardize it for all the subviews that need it. Right now some TreeVew uses property OutcomeImages and the others use LoadImages. I'll standardize it in #1212, maybe use an interface `IUseImageSet`

src/TestCentric/testcentric.gui/Presenters/ImageSetManager.cs
```diff
@@ -0,0 +1,76 @@
...
+        // This method has built-in knowledge of which views use the
+        // CurrentImageSet and what method to call on each.
+        // TODO: Use an event-oriented approach and let views pull
+        // changes as they need it, provided it doesn't duplicate
+        // too much code.
+        private void UpdateViews(OutcomeImageSet imgSet)
+        {
+            _view.TreeView.OutcomeImages = imgSet;
+            _view.TestResultSubView.LoadImages(imgSet);
+        }
```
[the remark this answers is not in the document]
CharliePoole: I'll look at that but it would mean moving the manager to the model, which somehow seemed wrong to me.

src/TestCentric/testcentric.gui/OutcomeImageSet.cs
CharliePoole: That's the old practice, which of course originated in the dark ages with single-pass compilers... even two-pass found it helpful. But C# allows placing definitions where they are used, so I'll often place a private member right by the only class that uses it, usually in front of it but this time it's after. Just my opinion of course.

src/TestCentric/testcentric.gui/OutcomeImageSet.cs
CharliePoole: Probably a good idea. I'll do that.

CharliePoole: Postponed for now - affects too many places.

Issue #1191, opened 2025-02-21
Improve the overview of test results


rowo360:
**Describe the bug**
In this issue, an enhancement is proposed to provide a better overview of the results of a test run.

I think that it is currently difficult to take in the test results with a quick glance. 
From my point of view it's because the information about the test results is distributed in different places.
On the one hand the tab 'Test properties' contains some information. However it's mixed with the info about test properties, so you need to set the focus on the proper part.
On the other hand the tab 'Errors and Failures' contains also information about test results. However some information is redunant.

Here's a screenshot:
<img src="https://github.com/user-attachments/assets/53942543-e7a2-4124-ab4b-41dc99ddd3ec" width="700">

**Expected behavior**
I don't propose that we change the visualization of test results. I just suggest that we rearrange the different parts, so that relevant information can be quickly overviewed.

**Environment (please complete the following information):**
- TestCentric Version:2.0.0-beta6
- OS: Windows 11

rowo360:
One possible solution would be to separate the two tabs:

- 'Test Properties' will only contain information about the tests
- 'Test Results' (former 'Errors and Failures') will contain all information about the result of the test run

We can switch automatically to the 'Test Results' tab when a test run finished.

rowo360:
Issue is solved with associcated PR:
Overall the test properties and test results are clearly separated and it supports the user to keep the focus. However further improvement is required to simplify the visualization of the test results. This will be done by a separate issue.

PR #1221, opened 2025-04-20
Fix for showing test result overview for test groups

rowo360:
This PR contributes to the previous work done for #1191 and #1193.
That work shows the test results of a selected tree node in the 'Test result tab'. However it did not work yet, if a **TestGroup** node was selected. In particular if the Tree is grouped and a category node or outcome node is selected.
Here's a screenshot: 

This PR solves the described behavior for TestGroup.


rowo360:
The code only handled `TestNode` objects so far, but not `TestGroup` objects. This was changed by using the common interface `ITestItem`. When retrieving the test counts for a group, all children must be iterated and their TestResult counts must be accumulated.

Issue #1220, opened 2025-04-10
Fail to load IDriverFactory extension


rowo360:
**Describe the bug**
This bug is targeting **TestCentric 1.6 and 1.7**
We recently extended our IDriverFactory extension to support also .Net 8.0. This results into this folder structure:

- NUnit.ConsoleRunner
- NUnit.ConsoleRunner.NetCore
- DriverFactoryExtension
  - Net48
  - Net80
- TestCentric

Each subfolder (Net48 and Net80) contains a version of the DriverFactory for his target framework.

This works fine using NUnit.ConsoleRunner for Net48 tests and NUnit.ConsoleRunner.NetCore for Net80 tests. Each runner loads and uses the respective DriverFactory extension.

However TestCentric fails to load the IDriverFactory extension at all now. The dialog showing the loaded extensions remains empty. And thus the DriverFactory is not used anymore and in our case the tests fails to load.

Or in other words: The presence of the Net80 folder (containing the DriverFactory for Net80) is already sufficient that TestCentric cannot load a Net48 test using its DriverFactory. If I delete the Net80 folder, TestCentric is successful again.

**Expected behavior**
I can imagine a two-step solution:
- It's possible to load Net48 tests using its DriverFactory even if a Net80 version of that DriverFactory exists.
- Full support of DriverFactory extensions targeting different frameworks and selecting the appropriate one

The first step is more important for now.

**Environment:**
 - TestCentric Version 1.7
 - OS: Windows 11
 - Target Runtime for your test assembly: .NET Framework 4.8

CharliePoole:
@rowo360 
First some questions...

* Is this your own company's extension?
* Is your primary goal to fix version 1 or version2 of TestCentric?

Regarding the merge problem. In fixing any issue, you have to first decide whether to target V1 or V2. For V1, let BRANCH be v1-build, for V2 let it be main. Then,

1. Make sure BRANCH is up to date, using pull or fetch.
2. Create your feature branch from BRANCH
3. Make your fixes.
4. Do a pull request, making sure to set the target to BRANCH.

In your case, it looks as if you created the branch from v1-build and then did the PR to merge to main. The easiest way to fix this is to close your PR without merging but don't delete the issue branch. Then create a new PR targeting v1-build. That will be much easier to review since the current PR also wants to apply every change from 1.7.0 to the present.

That said, I'll try to look at only your own commits and comment on them. :-)

Charlie

CharliePoole:
[...]

Issue #1088, opened 2024-08-12
Testcentric 2.0beta - IDriverFactory extension


rowo360:
**Description**
I made an attempt to activate a IDriverFactory extension using **TestCentric-Gui 2.0.0-beta5** - unfortunately without success. While trying to get it activated, I searched for the missing detail and also looked into the TestCentric source code and log files. Finally I came to the conclusion that there's a gap in the TestCentric agent for this use case, so I raised this bug. But please review my attempt in general, if I'm on the right track.

**Overview**
The IDriverFactory extension is successfully detected by the TestCentric GUI. It's listed in the extension dialog with the state enabled. However the agent fails to load the IDriverFactory and there's an error message in the GUI while loading a test.
![grafik](https://github.com/user-attachments/assets/63b71e5b-8894-4758-b203-b90d39287feb)

Exception callstack:
![grafik](https://github.com/user-attachments/assets/9f7d6525-970b-4c84-9c82-1c2d7fa7174e)

**IDriverFactory implementation**
- The DriverFactory implements interface TestCentric.Engine.Extensibility.IDriverFactory and is basically empty for the start
- The resulting assembly is located in a folder 'MyTestCentricAddin'
- In the TestCentricGUI folder the '\*.addin' file is extended with the path to the addin folder
  ../MyTestCentricAddin/*.dll
- With these steps the new addin is already successfully detected by the TestCentricGUI
- Next I added also an '\*.addin' file for the agent. It's located in the folder:
  TestCentric.Extension.Net462PluggableAgent.2.5.0\tools\agent
  And contains a similar line:
  ../../../MyTestCentricAddin/*.dll

My assumption was that these steps would be sufficient to enable the extension also for the agent.

**Observations**
I'm adding some observations from analyzing this issue - it might be helpful.
I activated the detailed logging and compared the log file from the agent with the log file from the GUI:

Here's the essential part from the GUI log file:
21:45:39.493 Info  [ 1] ServiceManager: Initializing ExtensionService
21:45:39.493 Info  [ 1] ExtensionManager: Initializing ExtensionPoints...
21:45:39.493 Info  [ 1] ExtensionManager: Assembly: testcentric.engine
21:45:39.493 Info  [ 1] ExtensionManager: Assembly: TestCentric.Engine.Api
[...]

PR #1222, opened 2025-04-28
Fix presence of different .Net versions of a IDriverFactory extension

rowo360:
This PR contains a proposal for fixing #1220  in **TestCentric 1.7.0**

The crucial question, however, is whether we actually want to have a new release **1.7.1** at all. 
The arguments in favor would be that this issue is currently blocking: As described in the issue the extension works fine using NUnit console, but TestCentric fails to use the extension at all now. That means that TestCentric can no longer run .Net 4.8 tests - the presence of the Net80 folder (containing the DriverFactory for Net80) is already sufficient that TestCentric cannot load a Net48 test using its DriverFactory.

The other option would be to fix this only in version 2.0. However I think a TestCentric 2.0 release is not expected in the near future and we also already noticed that some further work is required to support extensions in 2.0 in general (#1088).

This PR solves only the crucial part of the issue: to load Net48 tests using its DriverFactory even if a Net80 version of that DriverFactory exists. 

@CharliePoole: What's your opinion about this one?


rowo360:
Here’s an overview of the sequence how TestCentric processes the extensions and fails:

-  TestCentric detects both versions of the extension 4.8 and 8.0
- In method ExtensionService.ProcessCandidateAssembly() it processes both extension assemblies and comes to the conclusion that the version 8.0 is the better version for that assembly (method IsBetterVersion)
- So overall only the 8.0 version survives and the 4.8 version is dropped
- However before activating the extension there is a check if the target framework can be loaded at all in method ExtensionService.FindExtensionsInAssembly() – calling method CanLoadTargetFramework
- This sequence is executed identically in both processes: TestCentric.Gui and TestCentric.Agent
- The TestCentric.Gui process uses always the .Net 4.8 framework – so the CanLoadTargetFramework() method returns false and therefore a .Net 8.0 DriverFactory extension is never listed in the extension dialog
- The TestCentric.Agent process for a .Net 4.8 test uses also the .Net 4.8 framework – so again CanLoadTargetFramework() method returns false
- So overall the DriverFactory is not used anymore for a .Net 4.8 test


And some screenshots from the code:

rowo360:
The proposed fix is guided by the NUnit console code base and how the NUnit console handle this use case. Luckily the two code bases are almost identical in this part, so it’s easy to compare and find the difference.

The NUnit console code base contains an additional check in method `ExtensionService.ProcessCandidateAssembly()` if the target framework can be loaded at all – by calling the existing method `CanLoadTargetFramework`. The proposed fix just adds this check in the TestCentric code base as well. That means that before performing the `IsBetterVersion `check, it’s first check if the extension is applicable at all.
I don’t expect any side effects as this check (CanLoadTargetFramework) is executed already now – only a few steps later in that sequence.

If considering the TestCentric.Gui process use case:

- Both versions of the extension exists (4.8 and 8.0): version 8.0 is dropped by check CanLoadTargetFramework and so version 4.8 is shown in the extension dialog (behavior in 1.7.0: extension dialog shows no active extension)
- Only extension 8.0 exists: version 8.0 is dropped and extension dialog shows no active extension (same behavior as in 1.7.0)
- Only extension 4.8 exists: version 4.8 is shown in the extension dialog (same behavior as in 1.7.0)

If considering the TestCentric.Agent process use case when both versions 4.8 and 8.0 of the extension exists:
- Executing 4.8 test: version 8.0 is dropped by check CanLoadTargetFramework and so version 4.8 is used
- Executing 8.0 test: version 4.8 is dropped by check IsBetterVersion and version 8.0 is used

rowo360:
BTW: I guess something is wrong with this PR - maybe I used the wrong branch to create it (branch: v1-build)?
But that can be fixed easily - I think the discussion about the issue and the fix itself is more important.

CharliePoole:
Your fix for V1 is correct and I think the issue is a critical bug. So we'll do a 1.7.1 release. See some more comments in the issue. You'll have to delete and redo this PR first, however.

rowo360:
Perfect, thanks for the hint to handle this PR correctly - I'll close this PR and create a new one, now.

PR #1223, opened 2025-05-04
Fix presence of different .Net versions of a IDriverFactory extension

rowo360:
This PR contains a proposal for fixing #1220 in TestCentric **1.7.0**

The crucial question, however, is whether we actually want to have a new release 1.7.1 at all.
The arguments in favor would be that this issue is currently blocking: As described in the issue the extension works fine using NUnit console, but TestCentric fails to use the extension at all now. That means that TestCentric can no longer run .Net 4.8 tests - the presence of the Net80 folder (containing the DriverFactory for Net80) is already sufficient that TestCentric cannot load a Net48 test using its DriverFactory.

The other option would be to fix this only in version 2.0. However I think a TestCentric 2.0 release is not expected in the near future and we also already noticed that some further work is required to support extensions in 2.0 in general (#1088).

This PR solves only the crucial part of the issue: to load Net48 tests using its DriverFactory even if a Net80 version of that DriverFactory exists.

### DIFF
--- a/src/TestEngine/testcentric.engine.core/Services/ExtensionService.cs
+++ b/src/TestEngine/testcentric.engine.core/Services/ExtensionService.cs
@@ -338,6 +338,13 @@ namespace TestCentric.Engine.Services
                 {
                     var candidate = new ExtensionAssembly(filePath, fromWildCard);
 
+                    if (!CanLoadTargetFramework(Assembly.GetEntryAssembly(), candidate))
+                    {
+                        log.Debug("  Unable to load this assembly");
+                        return;
+                    }
+
+
                     for (int i = 0; i < _assemblies.Count; i++)
                     {
                         var assembly = _assemblies[i];


rowo360:
Here’s an overview of the sequence how TestCentric processes the extensions and fails:

-  TestCentric detects both versions of the extension 4.8 and 8.0
- In method `ExtensionService.ProcessCandidateAssembly()` it processes both extension assemblies and comes to the conclusion that the version 8.0 is the better version for that assembly (method `IsBetterVersion`)
- So overall only the 8.0 version survives and the 4.8 version is dropped
- However before activating the extension there is a check if the target framework can be loaded at all in method `ExtensionService.FindExtensionsInAssembly()` – calling method `CanLoadTargetFramework`
- This sequence is executed identically in both processes: TestCentric.Gui and TestCentric.Agent
- The TestCentric.Gui process uses always the .Net 4.8 framework – so the `CanLoadTargetFramework()` method returns false and therefore a .Net 8.0 DriverFactory extension is never listed in the extension dialog
- The TestCentric.Agent process for a .Net 4.8 test uses also the .Net 4.8 framework – so again `CanLoadTargetFramework()` method returns false
- So overall the DriverFactory is not used anymore for a .Net 4.8 test


And some screenshots from the code:

CharliePoole (approved):
Looks good... notice the following adminstrative changes, which I made for you...
1. Removed the assignment of the PR to me
2. Requested me as the reviewer of the PR
3. Added "Fixes #1223" to your initial comment to cause automatic closure

PR #1227, opened 2025-05-06
Add NUnitTree grouping 

### LINKED ISSUE #1226: Add grouping support for NUnit tree display strategy
**Describe the enhancement**
We are currently supporting the grouping of the test tree only for the FixtureList and TestList strategy. However this grouping functionality is completely missing for the NUnit display strategy. As this grouping functionality is helpful to set the focus on certain aspects of the tests, I propose to add grouping also for the NUnit tree - namely I propose grouping by category, by outcome and by duration. 
(This idea was first mentioned in #1195).

In contrast to the FixtureList and TestList grouping the NUnit tree grouping will display tree nodes for the assembly and namspaces - building up a tree hierarchy of tests - instead of a plain list of tests. This test hierarchy (which is well known from the NUnit tree) provides some kind of structured view to the tests and supports to execute individual parts within the test hierarchy.

**Screenshots**
Here are some screenshots from the enhancement:

  <br/>

 - TestCentric Version: 2.0.0-beta

rowo360:
This PR closes #1226 by providing grouping functionality to the NUnit tree display strategy.

I plan to implement the functionality incrementally and provide separate commits for indivdual parts. So it's easier to review and comment on that parts. Here's an overview of the individual steps:

1. Extend UI + save grouping in VisualState + Setting
2. Implement grouping
3. Support folding of namespaces
4. Update tree nodes icons during/after a test run
5. Improve performance for regrouping

### DIFF
--- a/src/TestCentric/testcentric.gui/Presenters/NUnitTreeDisplayStrategy.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/NUnitTreeDisplayStrategy.cs
@@ -52,7 +52,7 @@ namespace TestCentric.Gui.Presenters
             _view.EnableTestFilter(true);
         }
 
-        protected override VisualState CreateVisualState() => new VisualState("NUNIT_TREE", _settings.Gui.TestTree.ShowNamespace).LoadFrom(_view.TreeView);
+        protected override VisualState CreateVisualState() => new VisualState("NUNIT_TREE", _settings.Gui.TestTree.NUnitGroupBy, _settings.Gui.TestTree.ShowNamespace).LoadFrom(_view.TreeView);
 
         protected override string GetTreeNodeName(TestNode testNode)
         {
--- a/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/TestCentricPresenter.cs
@@ -239,6 +239,9 @@ namespace TestCentric.Gui.Presenters
                         UpdateTreeDisplayMenuItem();
                         UpdateViewCommands();
                         break;
+                    case "TestCentric.Gui.TestTree.NUnitGroupBy":
+                        _view.NUnitGroupBy.SelectedItem = _settings.Gui.TestTree.NUnitGroupBy;
+                        break;
                     case "TestCentric.Gui.TestTree.TestList.GroupBy":
                         _view.TestListGroupBy.SelectedItem = _settings.Gui.TestTree.TestList.GroupBy;
                         break;
@@ -463,6 +466,11 @@ namespace TestCentric.Gui.Presenters
             _view.ShowHideFilterButton.CheckedChanged += () =>
             {
                 _settings.Gui.TestTree.ShowFilter = _view.ShowHideFilterButton.Checked;
+            };
+
+            _view.NUnitGroupBy.SelectionChanged += () =>
+            {
+                _settings.Gui.TestTree.NUnitGroupBy = _view.NUnitGroupBy.SelectedItem;
             };
 
             _view.TestListGroupBy.SelectionChanged += () =>
@@ -916,6 +924,9 @@ namespace TestCentric.Gui.Presenters
 
             switch (displayFormat)
             {
+                case "NUNIT_TREE":
+                    _view.NUnitGroupBy.SelectedItem = _settings.Gui.TestTree.NUnitGroupBy;
+                    break;
                 case "TEST_LIST":
                     _view.TestListGroupBy.SelectedItem = _settings.Gui.TestTree.TestList.GroupBy;
                     break;
--- a/src/TestCentric/testcentric.gui/Presenters/TreeViewPresenter.cs
+++ b/src/TestCentric/testcentric.gui/Presenters/TreeViewPresenter.cs
@@ -133,6 +133,7 @@ namespace TestCentric.Gui.Presenters
                         Strategy.Reload();
                         break;
 
+                    case "TestCentric.Gui.TestTree.NUnitGroupBy":
                     case "TestCentric.Gui.TestTree.TestList.GroupBy":
                     case "TestCentric.Gui.TestTree.FixtureList.GroupBy":
                     case "TestCentric.Gui.TestTree.ShowNamespace":
@@ -338,7 +339,11 @@ namespace TestCentric.Gui.Presenters
         private void UpdateTreeSettingsFromVisualState(VisualState visualState)
         {
             _treeSettings.DisplayFormat = visualState.DisplayStrategy;
-            if (visualState.DisplayStrategy == "TEST_LIST")
+            if (visualState.DisplayStrategy == "NUNIT_TREE")
+            {
+                _treeSettings.NUnitGroupBy = visualState.GroupBy;
+            }
+            else if (visualState.DisplayStrategy == "TEST_LIST")
             {
                 _treeSettings.TestList.GroupBy = visualState.GroupBy;
             }
--- a/src/TestCentric/testcentric.gui/Views/IMainView.cs
+++ b/src/TestCentric/testcentric.gui/Views/IMainView.cs
@@ -76,7 +76,7 @@ namespace TestCentric.Gui.Views
         IViewElement DisplayFormatButton { get; }
         ISelection DisplayFormat { get; }
         ISelection TestListGroupBy { get; }
-
+        ISelection NUnitGroupBy { get; }
         ISelection FixtureListGroupBy { get; }
 
         IChecked ShowNamespace { get; }
--- a/src/TestCentric/testcentric.gui/Views/TestCentricMainView.cs
+++ b/src/TestCentric/testcentric.gui/Views/TestCentricMainView.cs
@@ -88,6 +88,11 @@ namespace TestCentric.Gui.Views
         private ToolStripMenuItem fixtureListMenuItem;
         private ToolStripMenuItem testListMenuItem;
         private ToolStripMenuItem nunitTreeShowNamespaceMenuItem;
+        private ToolStripSeparator nunitTreeSeparator;
+        private ToolStripMenuItem nunitTreeUngroupedMenuItem;
+        private ToolStripMenuItem nunitTreeByCategoryMenuItem;
+        private ToolStripMenuItem nunitTreeByOutcomeMenuItem;
+        private ToolStripMenuItem nunitTreeByDurationMenuItem;
         private ToolStripMenuItem textListUngroupedMenuItem;
         private ToolStripMenuItem textListByAssemblyMenuItem;
         private ToolStripMenuItem textListByFixtureMenuItem;
@@ -176,6 +181,9 @@ namespace TestCentric.Gui.Views
             DisplayFormat = new CheckedToolStripMenuGroup(
                 "displayFormat",
                 nunitTreeMenuItem, fixtureListMenuItem, testListMenuItem);
+            NUnitGroupBy = new CheckedToolStripMenuGroup(
+                "NUnitGroupBy",
+                nunitTreeUngroupedMenuItem, nunitTreeByCategoryMenuItem, nunitTreeByOutcomeMenuItem, nunitTreeByDurationMenuItem);
             TestListGroupBy = new CheckedToolStripMenuGroup(
                 "TestListGroupBy",
                 textListUngroupedMenuItem, textListByAssemblyMenuItem, textListByFixtureMenuItem, textListByCategoryMenuItem, textListByOutcomeMenuItem, textListByDurationMenuItem);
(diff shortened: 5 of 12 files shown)


rowo360:
**Commit: Extend UI + save grouping in VisualState + Setting**
This commit extends the UI to select a grouping of the NUnit Tree. These are again the familiar groupings: Category, Duration and Outcome. In addition, the UI contains the selection 'No grouping' to get the original NUnit view. Overall it looks like this:


The 'No grouping' is the default option and the current selection is saved in the VisualState file **and** in the settings file. The approach of saving the selection in both files is identical how the selected grouping of the FixtureList/TestList strategy is saved. This means even if there's no VisualState file present, the last selected grouping will be applied.

rowo360:
**Commit: provide main grouping logic**
This commit contains the main grouping logic - therefore I try to explain it in more detail.

When implementing the grouping, I have considered two different use cases: 
once when a test assembly is loaded and the entire tree has to be built up. And secondly, when a test has to be regrouped into a new group after a test run (for example: grouping by Outcome)
Instead of implementing both use cases separately, I want to choose an approach that supports both cases. With these thoughts in mind, I decided to build-up the tree by processing through the TestNodes representing test cases. This means that the same approach can also be used for the regroup, where you also have a test case node and have to sort it into an existing tree.

Overall, the approach consists of these steps:

1. Get a list of all test cases and process them one-by-one
2. Determine the names of the groups into which the test case is to be sorted (might be multiple groups in case of category grouping)
3. Create the TreeNode (if necessary) for the group (a root TreeNode)
4. Determine the path in the NUnit tree from the root to the test case (e.g. a list of TestNodes)
5. Iterate through this path and create the corresponding TreeNodes as children of the group node
6. In case the corresponding TreeNode already exists, it can be used instead.

With this approach, a grouped NUnit tree can be build up. And important, the same steps can also be used for the regrouping use case (regrouping is not part of this commit).
Here's a screenshot of the grouped NUnit tree (category, outcome and duration):

rowo360:
**Commit: provide main grouping logic (continued)**
In general, each TreeNode is connected to a TestNode via its tag property - this also applies for the NUnit grouping. The tree leaf nodes are connected to the TestNodes representing test cases. The inner tree nodes (Assembly, Namespace, Fixture) are not representing an existing TestNode entirely, but only a part of it. Therefore a TestGroup object is created for these tree nodes. This test group contains exactly those TestNodes that are available as children in the subtree. 

Here's an example for category grouping:


For Node1 a TestGroup is created containing 4 TestNodes: TestA, TestB, TestC and again TestA.
For Node2 a TestGroup is created containing 1 TestNode: TestA
For Node3 a TestGroup is created containing 11 TestNodes: I skip to list them...
For Node4 a TestGroup is created containing 3 TestNodes: TestD, TestE, TestF

So in this example, Node1 and Node3 represents only a part of the NUnit Assembly TestNode 'MyLibrary.Test.dll' and Node2 and Node4 only represents a part of the NUnit Fixture TestNode 'Feature1Tests'.

If the user selects such an inner TreeNode and starts a TestRun on it, the associated test group can be passed to the TestRun. So overall, only the child nodes of that subtree are executed.

rowo360:
**Commit: provide main grouping logic (continued)**
Here are some details about the code changes themself:

- I add all files related to the NUnit grouping in a separate folder: Presenters/NUnitgrouping
- I want to separate the grouping logic from the DisplayStrategy implementation. Therefore one base class `GroupingBase `and three derived classes `CategoryGrouping`, `DurationGrouping `and `OutcomeGrouping `are inserted.
- The described main grouping logic can be found in method `GroupingBase.CreateTreeNodes()`
- The creation of TreeNodes remain in the `DisplayStrategy` class. I add an interface `INUnitTreeDisplayStrategy` which is passed to the grouping classes so that they can trigger the creation of TreeNodes.
- I add a `Parent `property to the `TestNode `class. This property is required to determine the tree path given a leaf node - so I'm able to navigate the tree up.

Here's also a list of considerations which I finally dropped:

- The FixtureListStrategy/TestListStrategy already provides some grouping functionality and a base class GroupDisplayStrategy. I started to use this base class also for the NUnit grouping. But it turns out that I need to override almost all base implementations and add new virtual methods to comply to the new use cases, that only a tiny part remains. Overall, that doesn't feel beneficial at the end and I dropped this approach.
- The Linq `GroupBy` method provides build-in grouping functionality. We only need a list of objects and can group them by any property. I manged to use this Linq method successfully. However I dropped it because it was not such a big difference. And I need to do more research to find out how to create multiple groups in case of category grouping. I'm just mentioning it, in case you prefer this approach.
- While I was trying out Linq, I wondered if Linq can be used for extended grouping functionality. As stated, Linq just need a list of objects to start grouping and will return a set of lists grouped by one property. So this output can be used for another grouping operation - which means we can chain the grouping operations. For example this can be used to group by outcome within categories. Allthough this sounds interesting, I'm not convinced that this group-in-group functionality is really missing.
Or thinking in another direction: also the assembly, namespaces and fixtures can be considered as some grouping and Linq can be used to group them accordingly (in whatever order!). This sounds also interesting, but basically means that I need to drop all the available information which I already got from NUnit. I don't think any of this is necessary at the moment.
- I considered to create a model of the grouped NUnit tree first and afterwards start to create the Tree based on that model. Especially when it came to folding namespace nodes (described in later commit) which turned out to be rather complicated. But in the end, I never got to the point where I couldn't go on without a model. That's why I didn't do it.

rowo360:
**Commit: provide regroup logic**
This commit adds the regroup functionality for the outcome and duration grouping. It's getting triggered whenever an individual test is finished. The implementation of the regroup logic can be found in class `Regrouping.cs` and consists mainly of two steps: 
first the path from the old group to the leaf node is determined and the TestNode is removed from every TestGroup along this path. If in this process a TestGroup no longer contains any test nodes, the associated tree node can be removed. As a result of this step the TestNode is not represented anymore in the Tree. So, the next step can insert the TestNode into the new group again using the general approach of inserting TestNodes into the tree (described above).

Besides these two main steps the regrouping takes care to update tree node names of all changed tree nodes and to update the tree node icon. For example I extracted some parts dealing with the tree node names into a separate helper class, so that it can used from the `Regrouping` and `GroupingBase` class.

rowo360:
**Commit: support folding of namespace nodes**
I hadn't expected it, but it turned out that this part was the most complicated. The challenge was that you not only have to fold namespaces, but also unfold folded namespace nodes again if necessary. For example with outcome grouping , if the first test is passed, the test is sorted as the first and only one into the Passed group. Obviously all namespaces nodes of this test can be folded at this point in time. We also cannot predict whether another test will pass and which namespaces it contains. So it might occur that the folded namespaces nodes (which were accurate after the first test finished) must be unfolded again after the second test finished. This folding + unfolding namespace nodes is a hard challenge.

And although I found a solution for this, I ultimately discarded it: the solution was too error-prone - if something went wrong, everything fell apart.
Instead I was happy when I came up with an idea which is not 100% accurate, but which is simple. Instead of dynamically deciding whether namespace nodes can be folded based on the current existence of nodes in the subtree, I use the ungrouped NUnit tree for this decision. That means that the existing folding logic which is already in use for the ungrouped NUnit tree can be reused here. This approach is not 100% accurate, as in some cases some namespaces may not be completely folded - but I prefer inaccuracy to robustness. Nonetheless if required, this subtask can be improved in time.

Here's a screenshot:

Actually, one could have the impression that the namespace nodes 'MyLibrary.Test' and 'Folder_1' could also be folded together. However that's not the case because in the ungrouped NUnit tree, the 'MyLibrary.Test' node contains multiple namespace nodes 'Folder_1', 'Folder_2', 'Folder_3' - so no further folding is possible here.
(However, the namespace nodes 'MyLibrary' and 'Test' are folded accurately).

rowo360:
**Commit: set all tree node images on test run/loading**
The base class `DisplayStrategy` takes care about the image icon of tree nodes when loading the tree or when running a test. However it only takes care about tree nodes associated with **TestNodes**, but not for **TestGroups** - since we use TestGroups intensivly in the grouped NUnit tree, an extension is necessary. 

Without any extension the tree look liks this after a test run:


The tree node image icons must be considered in these two cases:

- When running a test
- When loading a tree and some test results are present from a previous test run (for example: by changing the selected grouping)

We receive test finish events from the Model for all kind of TestNodes - including Assembly, Namespace and Fixture nodes. We can use this event as a trigger to update the icon of the associated TestGroups. In other words: if the test finish event is triggered for a namespace node, there won't be any pending test results from a subnode, so we can start to calculate the appropriate icon of the assoicated tree nodes. I don't want to use the TestNode outcome provided by NUnit itself, because it represents the outcome of the complete TestNode - however the Treenodes in the grouped tree only represents a part of that TestNode.

Here's an example:


All these three TreeNodes represents a different part of the NUnit TestNode 'MyLibrary.Test'. And from NUnit perspective the outcome of the TestNode 'MyLibrary.Test' is failed. But for Node1 or Node3 which represents only a subset of tests from that TestNode, the outcome should be different from my point of view.

So, I want to use the test event as a trigger and calculate the outcome for the TestGroups. I considered two options for this: either iterate through all TestNodes of the TestGroup to get the individual tests results and finally determine an overall result or use the image index of the TreeNode children to get an overall result.
I decided for the later option, because it seems to be an overhead to ask all containing TestNodes for their results again. Especially since the TreeNode children already have this information.

This approach works fine, but it doesn't cover the top grouping nodes ('Slow' or 'CategoryA'), because there's no test event for these. Therefore, their image icon is determined after the test run finished using the same approach as above.

rowo360:
**Commit: improve regroup performance**
I used a project with 50 test fixtures with 100 test cases each, resulting in a total of 5000 test cases for this performance analysis. This is an artificial project in which each test case only consists of an Assert.IsTrue(true) statement. So these tests are running fast. I challenged the regroup functionality by running all tests for outcome grouping: this means all 5000 tests must be regrouped from 'Not run' to 'Passed'.

Here are some numbers of the test execution time for comparision:
Ungrouped: 12s
Fixture list + outcome grouping: 15s
Test list + outcome grouping: 120s
NUnit + outcome grouping: 125s

Well, this performance delay is not acceptable from my point of view and there is also a wild flickering of the tree during the test run, which is also very annoying. So, overall some improvement is required.
I used the tool DotTrace to analyze this use case and to get detailed insights about which code part is causing the delay. And this analysis clearly points to the tree redraw operation which is triggered whenever a test is regrouped - which is happening 5000 times here. This matches perfectly with the mentioned flickering...

A solution to this problem is not that difficult: we simply need to significantly reduce the number of redraw operations. So instead of immediately applying a test event to the tree, I have introduced a queue that buffers all incoming test events (the ResultNode). So, while running a test, this buffer increases in size, but without updating the tree view, yet.
The tree view update is triggered by a short timer, which takes all current elements in the queue and apply them on the tree. This will result into only one single redraw operation instead of many. So overall, this is some bulk processing approach.

With this approach I was able to reduce the test execution time for NUnit+outcome grouping to 14s.
I think that is an enormous improvement and acceptable - also the flickering is solved. I won't state that it's perfect - if the timer is triggered there's a quick, but noticable delay - but you can operate the tree while the test is running and regrouping is in process.
BTW: the timer is currently set to 800ms.

I think the approach is OK and hopefully easy to understand. From implementation side it requires a Timer and taking care about threading - so I took special take that nothing goes wrong here - the implementation can be found in class `RegroupTestEventQueue`.

rowo360:
I started working on this enhancement because we mentioned in a discussion a few weeks ago that this would be a cool feature. I have now reached a point where I am pleased with the functionality and implementation. I have tried to cover all types of use cases, so I am not aware of any outstanding issues/errors. As always, there might exists different views on some details of the implementation, but I hope at least not on the big chunks. At least it was my intention not to use magic tricks, but to build on existing approaches. But let's discuss it...

Now, I wanted to ask you for your opinion on this enhancement - I'm avoiding the term 'review', because I'm aware that this PR contains a lot of new code which is hard and challenging to review line-by-line. I spent some time to describe the general ideas and considerations of the individual commits in the comments above. I hope they are understandable and comprehensible. So it might also be an option to read through that comments to get the big picture (instead of looking into the code).

CharliePoole:
Thanks for all the explanatory comments. It will make it easier to follow what you are doing. I'll have some time for this over the weekend - hope that's not too long a wait.

rowo360:
@CharliePoole:
I just wanted to ask if it would be possible to continue with the merge for this PR now?
Of course I am totally aware that it's not the best approach (merging without reviewing) - especially when it includes several changes. But just in case you're short in time, it would be benefical for me to perform the merge - so that I can continue to work without risking any merge conflicts.

And in any case, I am still available to make any desired changes at a later date - definitely! So if you come accross some misbehavior or design flaw later, just give me a sign and I'll try to improve.

But let's do it in a way that is comfortable for you!

CharliePoole:
It has been waiting too long, so I simply merged it.

rowo360:
Thanks a lot!

Issue #1237, opened 2025-05-26
Update tree node icons


rowo360:
**Describe the enhancement**
I propose that we update certain tree node icons to provide a modern look & feel of the application. Ok, maybe that sounds too big - but basically I propose the following steps for this:

1. Update the image set 'Visual Studio' to the latest icons used by Visual Studio 2022
2. Update the image set 'Circles' 
3. Remove the image set 'Classic'
5. [Optional] Add an image set 'Visual Studio code'
6. [Optional] Add an image set 'Resharper'
7. [Optional] Rename image set 'Default' to 'TestCentric' or 'NUnit'

**Environment (please complete the following information):**
 - TestCentric Version: 2.0.0-beta6

rowo360:
The 'Classic' and 'Default' image set look pretty similar - so I propose to remove one for simplicity.
I choose the 'Classic' one for removal.

rowo360:
While looking around for test icons I came across **Visual Studio Code** and its test explorer. Some of the icons look really different here. Unusual for me at first glance - but some users might like them. Here's a screenshot:

<img src="https://github.com/user-attachments/assets/ffe9897d-eed2-497c-9db5-7aa9826e5894" width="300">

rowo360:
Also the Test explorer from **Jetbrains Resharper** tool uses an own set of icons. Some of them also unusual, but that might depend on the perspective!

<img src="https://github.com/user-attachments/assets/173510a1-21c7-454d-8287-1feedabaa2eb" width="400">


I noticed that Visual Studio code and also Resharper does not provide a separate icon for each state. However, I would not adapt this, but use the icons identically as these tools are doing it.
For example: Visual Studio code uses the same icon for state 'Ignore' and 'Warning'. And Resharper uses the same icon for state 'Ignore' and 'Explicit'.

CharliePoole:
Hi Roald @rowo360 !

Sorry, I've been neglecting you, so I'll try to answer you all in one shot here!!!


> 1. Update the image set 'Visual Studio' to the latest icons used by Visual Studio 2022

I agree completely. However, See Note 1, below.

> 2. Update the image set 'Circles'

Agreed. And the Radial Gradient option was my favorite as well.
  
> 3. Remove the image set 'Classic'
> 6. [Optional] Rename image set 'Default' to 'TestCentric' or 'NUnit'

[...]

PR #1238, opened 2025-05-27
Update tree node icons

rowo360:
This PR solves #1237 by providing new, updated tree node icons for the various image sets


rowo360:
I removed the 'Classic' image set completely. This includes to adapt some tests and also the build.cake script, which takes care to copy those images to the target.

rowo360:
I need to handle an additional use case:
Although we removed the image set 'Default', this image set name might be present in some TestCentricSettings.xml files. Well, just because it was the last selected image set by the user. Unfortunately our system crashes when trying to restore an unknown image set from the setting file.
So, I fixed this use case on the setting level: if AlternateImageSet is requested and it's the outdated value 'Default', it returns the new default value 'Classic'. The idea was to fix it centrally and not that all client of that property needs to take care about it.

CharliePoole:
@rowo360 All sounds good to me. T hanks.

PR #1241, opened 2025-06-01
Add tree node icon for running tests

### LINKED ISSUE #1240: Add tree node icon to identify running tests
**Describe the enhancement**
When a test run is started, it is currently not easy to identify which tests are still expected to produce results. Currently all tree node icons are set to the initial state when a test run starts. So, all tests look the same - regardless of whether they are included in the test run or not. This problem is not essential if you only have very fast tests or always run all tests. But in all other cases it's helpful to spot all running tests in a glance or to observe the progress.

Therefore I propose to introduce an additional tree node icon which states that this test is part of the test run, but did not provide a test result yet. The user has to wait until the test finished.
This intermediate icon will be replaced as soon as the test result is present with the accurate outcome icon.

Here's a screenshot from the expected enhancement:

With the new icon, the user can immediately see that the test run has been started with the node 'Feature2Tests' and that 'TestB' and 'TestC' will produce test results.

 - TestCentric Version: 2.0 beta6

rowo360:
This PR resolves #1240 by introducing a new tree node icon:
This icon states that this test is part of the test run, but did not provide a test result yet. The user has to wait until the test finished.

Here's a screenshot:

With the new icon, the user can immediately see that the test run has been started with the node 'Feature2Tests', that 'TestA' already succeeded and that 'TestB' and 'TestC' will produce test results.


rowo360:
I didn't create a new icon on my own, but take the icon used by Visual Studio for this purpose. I applied this icon for both our image sets 'Visual Studio' and 'Classic'. I tried to modify this icon for our 'Classic' image set first, but all these attempts were unsatisfactory. So, I prefer to use the existing one also for our 'Classic' image set.
However for the 'Circle' image set I created an icon on my own - some faded circle with some border - trying to express that filling that circle with some color will still come.

rowo360:
There was already a code part that reset all tree icons when a test run was started. I only need to modify this code part to apply a different icon in case the tree node is part of the test run. In order to answer this question I extended the ITestModel with a new method `IsInTestRun(TestNode testNode)`. This method uses the private class `TestRunSpecification` which contain a list of all TestNodes currently executed.

In a regular test run all tree nodes having this new icon will be updated during the test run. As soon as the outcome is present, the appropriate outcome icon will be set. So, this new icon is only an intermediate state and should not remain after a test run. This applies also if a test run is stopped. In this case a TestFinished event is raised for all remaining tests and thus the icon is updated.

There's only a gap for the 'Force stop test run' use case. In this case there are no TestFinished events anymore. Therefore I added some code part in the `TestRunFinished` event to revert all new 'running' icons to the original 'not run' icon.

PR #1244, opened 2025-06-10
Show test results from previous test runs

### LINKED ISSUE #1243: Show test results from previous test runs
**Describe the enhancement**
This isn't a brand new idea, but mentioned several times in previous issues. 
Sometimes I don't execute all tests at once, but execute one part of the tests one after another.
However whenever I start a new test run all previous test results are reset, so I cannot view these results later on.

Therefore I propose that test results are not reset anymore when starting a new test run. Instead I propose to display the tree node icons of previous test results differently (dimmed or faded) - so that's clear that these results are not part of the latest test run but from a previous one.

So overall the user has an accumated view of all test results from all previous test runs.

This described enhancement might also be applied when reloading tests - however that's not the focus now - but something for the future.

 - TestCentric Version: 2.0 beta6

rowo360:
This PR resolves #1243:

- the test results are not reset anymore before starting a new test run
- the set of tree node icons is extended to display test results from previous test runs with a separate icon. These new icons are almost identically to the original ones, but dimmed. For the image set 'Visual Studio' the orginal icons for this state are used.

Here is a screenshot for all 3 different image sets:

If the user selects a failed test of a previous test run, all details of this failure are displayed in the 'test result' tab. So, failures can be analyzed later on regardless if the result is from the latest or a previous test run.


rowo360:
The new tree node icons are added to the image list of the TreeView. I decided to place these new images not at the end of this list but just before the regular images. This approach works best when determining the ImageIndex for `TestGroups` which uses the max value of the containing tests. For example: a TestGroup with two tests (one test failed from a previous run and the other test passed from the latest run) will get the tree node icon 'Passed' (without dimming).

rowo360:
I'm changing this PR to state draft, now. I just want to reconsider my last comment, if this is really the expected behavior.

PR #1281, opened 2025-08-21
Automatically check/uncheck child nodes when parent is checked

### LINKED ISSUE #199: Check boxes of lower level tests should be selected once test of higher level is selected
I find it useful to have following in GUI runner:
Selecting test tree view with check boxes, and selecting one test which consist of other tests (e.g. with attribute Values) all under tests should be selected. If so implemented it is easy to just select out some tests (like inverse selection) and to test all others except those unselected. When there are 100 tests it really simplify running all tests except couple of them.

From https://bugs.launchpad.net/nunit-3.0/+bug/498178

rowo360:
This PR fixes #199 by automatically check/uncheck all child nodes when parent node is checked/unchecked.

Actually, only all child nodes need to be updated as soon as the check state of one node changes. However in addition we need to distingush if the action is triggered by the user or programmatically. An example for programmatically is while restoring the check state from the VisualState file or another example while we updating the child node state the TreeView invoke the event again.

Fortunately, the TreeView event provides an EventArgument that can be used to detect this.

### DIFF
--- a/src/GuiRunner/TestCentric.Gui/Views/TestTreeView.cs
+++ b/src/GuiRunner/TestCentric.Gui/Views/TestTreeView.cs
@@ -102,7 +102,15 @@ namespace TestCentric.Gui.Views
 
             treeView.AfterSelect += (s, e) => SelectedNodeChanged?.Invoke(e.Node);
 
-            treeView.AfterCheck += (s, e) => AfterCheck?.Invoke(e.Node);
+            treeView.AfterCheck += (s, e) =>
+            {
+                // Update checked state of all child nodes if action is triggered by user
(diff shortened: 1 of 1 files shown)


rowo360:
Although the first commit was already fully functional, I was not yet completely satisfied: while we're checking all child nodes (one-by-one), the AfterCheck event is raised for every change. That's an unnecessary overhead and slightly noticable in a huge test project.
Therefore I decided to enhance the approach by delaying the AfterCheck event until all child nodes are checked, which also means that it's invoked only once now.

Another observation:
Checking the child nodes is really only a visual assistance for the user. When we determine the TestSelection for a TestRun, we don't consider any child nodes as soon as we encounter a checked node. That checked node already states that all child nodes are included in the test run.

PR #1282, opened 2025-08-22
Remove dependency on Mono.Options

### LINKED ISSUE #1273: Replace Mono.Options with a new implementation
The GUI Runner support for command-line options is based on the `Mono.Options` nuget package. Unfortunately, the package is not signed, which prevents any of our assemblies that depend on it, directly or indirectly, from being signed. This is the immediate motive for replacement. In doing so we may want to use newer technology, such as `System.CommandLine`.

I'm initially characterizing this as a "design issue" until we decide exactly how it will be replaced.

CharliePoole:
Fixes #1273


src/GuiRunner/TestModel/GuiOptions.cs
```diff
@@ -177,14 +245,14 @@ public string GetHelpText()
             writer.WriteLine();
             writer.WriteLine("Options:");
 
-            WriteOptionDescriptions(writer);
+            //WriteOptionDescriptions(writer);
```
rowo360, line 248: Isn't it sufficient to iterate through all dictionary items and write the Aliases + Description to restore this functionality?

src/GuiRunner/TestModel/GuiOptions.cs
```diff
@@ -85,35 +89,99 @@ public CommandLineOptions(params string[] args)
...
+                return false;
+
+            string opt = arg.StartsWith("--")
+                ? arg.Substring(2)
+                : arg.Substring(1);
+            string val = null;
+            int delim = opt.IndexOfAny(['=', ':']);
+            if (delim > 0)
+            {
+                val = opt.Substring(delim + 1);
```
rowo360, line 156: I'm not sure about this one: a String.Trim(...) isn't required here?

rowo360 (approved):
Of course, I would have preferred it if we had used the `System.CommandLine` class. But I think it's just as important to choose pragmatic solutions in the right place. And without much knowledge of the differences between `System.Commandline` and `Mono.Options`, I think you managed to remove the dependency with just a few changes.

Regarding signing the assembly - I think it's fine if we sign the assemblies. But signing is only required for uniqueness in the GAC, isn't it? It's to avoid name collisions from different publishers. By default we don't place any assemblies in the GAC ourself - there's no msi or other installer, isn't it? So overall, it is a precautionary measure in case one of our assemblies ends up in the GAC.

PR #1324, opened 2025-09-18
Fix serialization of package settings

### LINKED ISSUE #1315: .Net framework: Compile test dll which is loaded in TestCentric - Cannot compile because it's being used by another process
**Describe the bug**
While developing some NUnit tests in Visual Studio, I already opened the test dll in TestCentric and ran the tests to confirm that the tests are working as expected.
However when editing and compiling the NUnit tests again, Visual Studio complains that the dll cannot be generated because the dll is used by another process. Visual Studio even provides a list of processes: it's the Net462 Pluggable Agent process.

This bug is identical to issue [55](https://github.com/TestCentric/TestCentric.Agent.Core/issues/55) which was already solved sometime ago, but somehow it has crept in again.

It only occurs for .Net framwork - .Net Core test assemblies seems not to be blocked by the agent process

 - TestCentric Version: 2.0 beta 7

rowo360:
This PR fixes #1315 

Although the issue is about one dedicated use case, it turns out to be a general issue of the package settings which might cause negative effects also in other use cases.

The package settings support different typed settings (string, int, bool) and a typed access `T GetValueOrDefault<T>(...)`. When serializing the package settings from the UI process to the agent process all settings are serialized as a string. So far, so good. However, when a setting is to be deserialized, it is not known which target type should be used (int, bool or string). That is why, until now, all settings were restored as strings during deserialization.
That means basically, that all int, bool settings were transformed to a string setting:
  ShadowCopyFiles=true => ShadowCopyFiles="true"
  TestRunTimeout=30 => TestRunTimeout="30"

Accessing such a setting afterwards `bool v = packageSettings.GetValueOrDefault(SettingDefinitionsShadowCopyFiles)` will always return false;

As a solution, I chose a simple approach that checks whether a bool or int is present by using the `TryParse` method. This works fine, but I'm aware that this solution is not perfect! For example if we intend to store the string "True" it will be converted to a bool, now. That's also an unexpected behavior, but I think that it's more a theoretical edge case. 

A different solution might include the type in the serialization stream. Or we might identify the target type by looking up the set of SettingDefinitions using the setting name. If you prefer one of these alternative approaches, I can try to improve the approach.

### DIFF
--- a/src/TestEngine/TestCentric.Engine.Api.Tests/PackageSettingsTests.cs
+++ b/src/TestEngine/TestCentric.Engine.Api.Tests/PackageSettingsTests.cs
@@ -1,0 +1,51 @@
+// ***********************************************************************
+// Copyright (c) Charlie Poole and TestCentric contributors.
+// Licensed under the MIT License. See LICENSE file in root directory.
+// ***********************************************************************
+
+using NUnit.Framework;
+
+namespace TestCentric.Engine
+{
+    [TestFixture]
+    internal class PackageSettingsTests
+    {
+        [TestCase(nameof(SettingDefinitions.ShadowCopyFiles), "True", true)]
+        [TestCase(nameof(SettingDefinitions.ShadowCopyFiles), "False", false)]
+        [TestCase(nameof(SettingDefinitions.ShadowCopyFiles), "false", false)]
+        [TestCase(nameof(SettingDefinitions.ShadowCopyFiles), "FALSE", false)]
+        [TestCase(nameof(SettingDefinitions.TestRunTimeout), "10", 10)]
+        [TestCase(nameof(SettingDefinitions.TestRunTimeout), "0", 0)]
+        [TestCase(nameof(SettingDefinitions.TestRunTimeout), "-100", -100)]
+        [TestCase(nameof(SettingDefinitions.SelectedAgentName), "Agent", "Agent")]
+        [TestCase(nameof(SettingDefinitions.SelectedAgentName), "", "")]
+        public void Add_Setting_AsString(string settingName, string stringValue, object expectedValue)
+        {
+            // Arrange
+            PackageSettings settings = new PackageSettings();
+
+            // Act
+            settings.Add(settingName, stringValue);
+
+            // Assert
+            var value = settings.GetSetting(settingName);
+            Assert.That(value, Is.EqualTo(expectedValue));
+        }
+
+        [TestCase("UnknownSetting", "Value", "Value")]
+        public void Add_UnknownSetting_AsString(string settingName, string stringValue, string expectedValue)
+        {
+            // Arrange
+            PackageSettings settings = new PackageSettings();
+
+            // Act
+            settings.Add(settingName, stringValue);
+
+            // Assert
+            Assert.That(settings.HasSetting(settingName), Is.True);
+
+            var value = settings.GetSetting(settingName);
+            Assert.That(value, Is.EqualTo(expectedValue));
+        }
+    }
+}
--- a/src/TestEngine/TestCentric.Engine.Api.Tests/PackageSettingsTests.cs
+++ b/src/TestEngine/TestCentric.Engine.Api.Tests/PackageSettingsTests.cs
@@ -8,7 +8,7 @@ using NUnit.Framework;
 namespace TestCentric.Engine
 {
     [TestFixture]
-    internal class PackageSettingsTests
+    internal class PackageSettingFactoryTests
     {
         [TestCase(nameof(SettingDefinitions.ShadowCopyFiles), "True", true)]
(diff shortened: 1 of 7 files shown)


CharliePoole:
In terms of the design intent, I originally looked at three options...

1. Decision on deserialized Type made made at each individual call site.
2. Including the target type in the stream itself.
3. Using a definition for each setting, which included the expected type and possibly a default value for each setting.

I chose 3 and I'd like to stick with that. Here's my thinking about each option...

1. This is what we did in NUnit V2 and V3 and TestCentric V1.. It's very flexible but, unsurprisingly, it's also very error-prone.
2. This is what `BinarySerializationFormatter` did for our communications. Since Microsoft, with all it's resources, was unable to make this secure, I figure it's better not to try it.
3. This approach is encapsulated in the `SettingDefinition` class we are using for V2 and will be in NUnit V4 as well. It's need when its used but some of our code doesn't use it yet, including the spot you found.

A `TestPackage` contains two sorts of settings, for our purposes, those we understand and those we don't. Those we don't (i.e. based on the name) are never interpreted or acted upon. They are merely passed on to the framework, in case the framework understands them. Framework settings, because of how they are passed may be of any object type. However, because they are sent from the agent, we have to deserialize them anyway. That means the engine needs to know the name and type of every setting, even framework settings. This is obviously a limitation and, what's worse, we aren't totally there yet.

So, for all that, handling ints and bools as a fallback may be a temporary solution, but we'll eventually need to do more as this area is far from complete.

rowo360:
I've mixed something up and need to take another close look at this.

src/TestEngine/TestCentric.Engine.Api.Tests/PackageSettingsTests.cs
```diff
@@ -0,0 +1,51 @@
...
+            settings.Add(settingName, stringValue);
+
+            // Assert
+            Assert.That(settings.HasSetting(settingName), Is.True);
+
+            var value = settings.GetSetting(settingName);
+            Assert.That(value, Is.EqualTo(expectedValue));
+        }
+    }
+}
```
CharliePoole, line 51: Nice. We've been missing tests in this area.

You are verifying correct use of pre-defined `SettingDefinition`  names but what about incorrect use? Can we guarantee that the provision of an incorrect value Type for a known setting will fail? From inspection, I think your code does guarantee that, but a test would be good. We should test for incorrect use of (string, bool, int) as well as objects that are not of those types.

In the wild, we may still some failures until we are correctly converting all Types used by the framework to an XML or other string representation but I think that's for another issue.

src/TestEngine/TestCentric.Engine.Api/SettingDefinitions.cs
```diff
@@ -341,6 +342,23 @@ public static class SettingDefinitions
...
+            // Get a list of all public properties
+            var properties = typeof(SettingDefinitions).GetProperties(BindingFlags.Public | BindingFlags.Static);
+
+            foreach (PropertyInfo propertyInfo in properties)
+                if (propertyInfo.Name == name)
+                    return propertyInfo.GetValue(null, null) as SettingDefinition;
+
+            return null;
+        }
+
```
CharliePoole, line 362: I'm of two minds here. On the one hand, we could do this one time only as a property of `SettingDefinitions`, constructed lazily. On the other hand, that's something else to go wrong in the future. :-( 

What do you think?

CharliePoole (changes requested):
I really like the overall shape of this change. I'd like some tests of the non-happy path. The idea of only constructing the property list once is optional as far as I'm concerned.

src/TestEngine/TestCentric.Engine.Api/SettingDefinitions.cs
rowo360: I added a  cache now, that we need to get the list of known SettingDefinitions only once. I think that this list won't change in runtime, so I hope that it won't break in future.

src/TestEngine/TestCentric.Engine.Api.Tests/PackageSettingsTests.cs
rowo360: I added some tests for unknown settings now and check that in this case the string is stored.

src/TestEngine/TestCentric.Engine.Api/SettingDefinitions.cs
CharliePoole: In fact, my intention was that the list of known `SettingDefinitions` should be fixed at compile time and not subject to change. That's the main reason for the separation between `PackageSetting(s)` and `SettingDefinition(s)`. Of course somebody could come up with a use case for defining new "known" settings on the fly, but I don't want to anticipate the need.

CharliePoole (approved):
This looks great.

PR #1326, opened 2025-09-21
Enhance save result dialog to support different results formats

### LINKED ISSUE #1313: Save test results in various formats
The user should be able to save the result of a test run, currently displayed in the GUI in any of the supported formats, both built-in and extension-based. Currently, the native format, `nunit3` is the only one shown when the menu item is clicked.

rowo360:
This PR fixes #1313 

First of all I verified that all NUnit `IResultWriter` extensions are properly recognized and displayed in the extension dialog. This works fine!

Next, I enhanced the SaveDialog. The FileFormat drop-down box now contains all available result formats. This allows the user to select their desired format. Nunit3 is the default option and the only option in case no IResultWriter extensions are available.

This approach using the FileFormat drop-down box replaces the previous approach using a submenu with all available extensions (see screenshot in issue). That approach worked as well, but it seems to me that the new approach is the more general one.

### DIFF
--- a/src/GuiRunner/TestCentric.Gui/Presenters/TestCentricPresenter.cs
+++ b/src/GuiRunner/TestCentric.Gui/Presenters/TestCentricPresenter.cs
@@ -81,6 +81,7 @@ namespace TestCentric.Gui.Presenters
             UpdateViewCommands();
             UpdateTreeDisplayMenuItem();
             UpdateRunSelectedTestsTooltip();
+            UpdateSaveResultFormatsMenuItem();
 
             WireUpEvents();
             _view.ShowHideFilterButton.Checked = _settings.Gui.TestTree.ShowFilter;
@@ -490,7 +491,6 @@ namespace TestCentric.Gui.Presenters
 
             _view.RunParametersButton.Execute += DisplayTestParametersDialog;
 
-            _view.SaveResultsCommand.Execute += SaveResults;
             _view.TransformResultsCommand.Execute += TransformResults;
 
             _view.OpenWorkDirectoryCommand.Execute += () => System.Diagnostics.Process.Start(_model.WorkDirectory);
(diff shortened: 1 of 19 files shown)


src/GuiRunner/TestCentric.Gui/Presenters/TestCentricPresenter.cs
```diff
@@ -84,10 +82,6 @@ public TestCentricPresenter(IMainView view, ITestModel model, GuiOptions options
             UpdateTreeDisplayMenuItem();
             UpdateRunSelectedTestsTooltip();
 
-            foreach (string format in _model.ResultFormats)
-                if (format != "cases" && format != "user")
-                    _resultFormats.Add(format);
-
```
CharliePoole, line 90: We need to support "user" and possibly "cases" - I'm not sure about that one.

The "user" format is the main reason I used a popup menu  rather than a File save dialog filter. That one is different from the others. If the user selects "user" they then need to specify the name of an xslt file to open in a file open dialog. Making them use a file save followed by a file open seems clumsy.

Probably, the "user" format will have to be the driver for this design.

Of course, the "ideal" solution would be to take over

rowo360:
Thanks for pointing out this special feature with the "user" and "cases" terms. I've also wondered why these two values exist and was about to ask. 😄  But since they had been filtered out until now, I left it that way for the time being.

So, the term "user" is used whenever some result transformation should be applied via a xslt file - got it! I never used it before myself, but I read about it in the NUnit3 console command line options now. And indeed we need a different UI to support this feature.

One question: how do we invoke a test result transformation using the API? Currently results are saved with these lines:
```
var resultWriter = Services.ResultService.GetResultWriter(format, new object[0]);
resultWriter.WriteResultFile(results.Xml, filePath);
```

Is the xslt file passed in via some arguments? Or is this done completely different?

rowo360:
Perfect - thanks for the hint 👍  
With that, I have successfully transformed a test result to a html file. Next, I will create the UI for it. I think I will leave the ‘Save result’ option unchanged and instead introduce another menu item called ‘Transform result’. This new menu item opens a small dialog box in which the user can specify both the xlst and the target file.

rowo360:
I added a new menu item to support the transformation of test results:


This functionality is kept separated from 'Save test results' because it requires the selection of a xslt file. Therefore the new menu item opens a dialog in which the user can enter the xslt and the target file.



The transformation process itself is executed using the 'user' IResultWriter - these lines are added to the `TestModel`:
```
            var resultWriter = Services.ResultService.GetResultWriter("user", new[] { xsltFile });
            var results = TestResultManager.GetResultForTest(LoadedTests.Id);
            resultWriter?.WriteResultFile(results.Xml, targetFile);
```

The GUI is responsible to catch any exceptions and display the output to the user, so that he's informed about the failure. I tested this behavior with none existing or invalid xslt files successfully.

PR #1340, opened 2025-10-24
Enable 'Select agent' menu item when using NUnit agents

rowo360:
This PR (re-)enables the 'Select agent' menu item after switching to the NUnit agents and NUnit extensibility.
Here's a screenshot:
image

Some comments about the actual code changes:

- Instead of creating a new instance of the `TestAgentService`, I propose to retrieve an instance using a `GetService `call. The existing code which creates a new instance missed to start the service. So, it was not initialized and contains no provides.
Using the GetService call we retrieve an initialized service. One drawback might be that we use a concrete class in the GetService call instead of an interface, but I copied this call from the AssemblyRunner class - so, I assume it's OK for now.

- After this change the method `RuntimeFramework.Parse` thrown an exception, because the input doesn't have the expected format. Luckily I spotted another existing method (`RuntimeFramework.FromFrameworkName`) which is capable to process the input and provide the FrameworkName. So, I switched to use this method.
Here's a screenshot of the runtime name:
image

### DIFF
--- a/src/GuiRunner/TestModel/TestServices.cs
+++ b/src/GuiRunner/TestModel/TestServices.cs
@@ -24,7 +24,7 @@ namespace TestCentric.Gui.Model
 
             ExtensionService = GetService<IExtensionService>();
             ResultService = GetService<IResultService>();
-            TestAgentService = new TestAgentService();
+            TestAgentService = GetService<TestAgentService>();
         }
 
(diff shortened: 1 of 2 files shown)


rowo360:
Oh, many tests are failing 🤔 - I didn't notice it in advance. I will fix these, but maybe you can confirm the approach in general first?

PR #1341, opened 2025-10-25
Further changes to support use of nunit.extensibility.dll

### LINKED ISSUE #1336: Eliminate use of wrapper classes for NUnit extensions
To do this, we will need to replace use of a large number of TestCentric API Types with the corresponding NUnit Types. This is a large change and will be done over multiple PRs.

CharliePoole:
Fixes #1336

### DIFF
--- a/src/GuiRunner/TestCentric.Gui/Presenters/AgentSelectionController.cs
+++ b/src/GuiRunner/TestCentric.Gui/Presenters/AgentSelectionController.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT License. See LICENSE file in root directory.
 // ***********************************************************************
 
+using System;
 using System.Collections.Generic;
 using System.Windows.Forms;
 using TestCentric.Engine;
@@ -40,6 +41,7 @@ namespace TestCentric.Gui.Presenters
             defaultItem.Checked = true;
             defaultItem.Enabled = true; // Should always remain enabled
             agentMenu.MenuItems.Add(defaultItem);
+            defaultItem.Click += OnAgentMenuItemClicked;
 
             foreach (var agentFullName in _model.AvailableAgents)
             {
@@ -52,6 +54,7 @@ namespace TestCentric.Gui.Presenters
                 var agentItem = new ToolStripMenuItem(agentName);
                 agentItem.Tag = agentFullName;
                 agentMenu.MenuItems.Add(agentItem);
+                agentItem.Click += OnAgentMenuItemClicked;
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/GuiRunner/TestCentric.Gui/Presenters/AgentSelectionController.cs
```diff
@@ -39,7 +39,7 @@ public void PopulateMenu()
             agentMenu.Enabled = agentsToEnable.Count > 1;
             if (agentMenu.Enabled)
             {
-                agentMenu.MenuItems.Clear();
+                //agentMenu.MenuItems.Clear();
```
rowo360: I assume that this is precisely the code which you mentioned that doesn't quite work yet. Currently the menu items for the agents are added over and over again when the pop-up menu is opened (and the dummy entry is displayed)

rowo360 (approved):

src/GuiRunner/TestCentric.Gui/Presenters/AgentSelectionController.cs
CharliePoole: I can see that what I checked in was not what I intended to check in. In the GUI, I often do simple experiments, e.g. if I remove this line, what will I see? This was one of those and in fact it's how I realized that it's not necessary to construct everything from scratch each time. I removed it but must not have committed the change before I pushed.

I already have the code to a point where there are two methods, one to construct and one to update based on what is being loaded, and it seems more understandable to me. I'll push it so you can see it.

src/GuiRunner/TestModel/TestModel.cs
```diff
@@ -117,11 +119,13 @@ public static ITestModel CreateTestModel(ITestEngine testEngine, GuiOptions opti
...
         // Services provided either by the model itself or by the engine
-        public ITestServices Services { get; }
+        public IServiceLocator Services { get; }
 
         public UserSettings Settings { get; }
 
-        public IList<string> AvailableAgents { get; }
+        // Since we generate a new name when agents are wrapped,the available agent list must created dynamically.
+        public IList<string> AvailableAgents =>
+            [.. TestEngine.Services.GetService<ITestAgentProvider>().GetAvailableAgents().Select((a) => a.AgentName)];
```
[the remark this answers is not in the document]
CharliePoole: You may notice I removed caching entirely from Services. I'm of two minds about keeping the class, although a very simple forwarding property  `Services => TestEngine.Services` may be convenient`. Or possibly `EngineServices` to make it clear that the engine is doing something for us. What do you think.

In general, I am feeling that the `TestModel` has become ore complicated than necessary, but that's for another issue I think. :-)

src/TestEngine/testcentric.engine/Services/TestAgency.cs
```diff
@@ -29,12 +30,11 @@ class TestAgency : ITestAgentProvider, ITestAgency, IService
 
         private const int NORMAL_TIMEOUT = 30000;               // 30 seconds
         private const int DEBUG_TIMEOUT = NORMAL_TIMEOUT * 10;  // 5 minutes
+        private const string AGENT_LAUNCHERS_PATH = "/TestCentric/Engine/AgentLaunchers";
```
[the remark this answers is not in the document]
CharliePoole: Yes... it seems that each application has to define it's own extension points. Using `TypeExtension` to define them only works cleanly if the same application (i.e. NUnit Engine) both defines and implements them. I experimented with using those extension points, but I like the use of `Extension` better. 

Of course, this implies that all extension points - at least all those being used at one time - should have a unique path. Therefore, beginning the path with the name of the application, possibly followed by the component, seems like the best pattern.

src/GuiRunner/TestCentric.Gui/Presenters/AgentSelectionController.cs
CharliePoole: @rowo360 It turns out that NUnit itself is returning a short agent name... i.e. "Net80AgentLauncher" rather than "NUnit.Engine.Agents.Net80AgentLauncher." I'll do a workaround for now.

src/GuiRunner/TestCentric.Gui/Presenters/AgentSelectionController.cs
```diff
@@ -33,68 +33,69 @@ public bool AllowAgentSelection()
...
                 foreach (ToolStripMenuItem item in agentMenu.MenuItems)
                 {
-                    if ((string)item.Text == selectedAgentName)
-                        item.Checked = isItemChecked = true;
-                    else
-                        item.Click += (s, e) =>
+                    string itemTag = item.Tag as string;
+                    item.Enabled = itemTag == "DEFAULT" || agentsToEnable.Contains(itemTag);
+
+                    item.Click += (s, e) =>
```
rowo360: I propose to move the Click event registration to method `PopulateMenu`: otherwise we are registering over and over again when opening the menu. I didn't notice any totally bad behavior, because the `!item.Checked` condition is only fulfilled on the first invocation. But nonetheless the event notification is invoked multiple times.

src/GuiRunner/TestCentric.Gui/Presenters/AgentSelectionController.cs
```diff
@@ -33,68 +33,69 @@ public bool AllowAgentSelection()
...
+                    };
                 }
+            }
 
-                if (!isItemChecked)
-                {
-                    defaultMenuItem.Checked = true;
-                    //packageSettings.Remove(SettingDefinitions.SelectedAgentName);
-                }
+            void EnsureSingleItemChecked(ToolStripMenuItem itemToCheck)
```
rowo360: This code section reminds me that we have a class `CheckedToolStripMenuGroup` which takes care that exactly one item of a set of ToolStripMenuItems is checked. So the presenter class would work basically with an ISelection.

Hmm, I think it's not necessary to change this entire code to that approach, now. But I just want to mention it 😄

CharliePoole: This code doesn't make a lot of use our Element structure, except for IPopup to kick things off. It would be cleaner to use it but really only for our own satisfaction.

Issue #1342, opened 2025-10-26
Convert Settings to use standard Windows application settings


CharliePoole:
Originally, user settings were supported by both the NUnit and TestCentric engines using an XML settings file. This facility was removed from both engines as breaking changes for version 4.0 of NUnit and 2.0 of TestCentric. In the case of NUnit, the settings file no longer exists as all information is passed using command-line options or environment variables.

TestCentric continues to have a settings file and the TestCentric engine retains some classes that support it, which are no longer present in the NUnit engine. For version 2.0, we should remove these files and start using Windows support based on the configuration file.

Part of this issue will involve designing how to store the configuration as well as how to interface it with our existing static classes that access the settings. Keeping that top-level interface is desirable in order to avoid wide-ranging changes to the GUI.

@rowo360 This is part of the GUI other than removing the classes in the engine itself that support it. Would you feel like taking this one on?

rowo360:
Yes of course, I can start to get familiar with the requirement and do the necessary adaptations. Let's start with some beginners questions:

So, we are talking about our `UserSettings` class, which is divided into `GuiSettings` and `EngineSettings`. All these settings are stored in one file TestCentricSettings.xml which is located in folder %appData%/local/TestCentric. But you have in mind some different location or different format as the standard Windows application settings?

The second part of this issue is about the EngineSettings which contains these indiviual settings:

- ShadowCopyFiles
- Agents
- RerunOnChange
- SetPrincipalPolicy
- PrincipalPolicy

I think we could move the RerunOnChange setting to the GuiSettings. But is one of the four setting still supported by NUnit engine and we can pass it to NUnit?

PR #1346, opened 2025-10-29
Use ConfigurationManager to store user settings

rowo360:
This PR is intended only as a draft for #1342 (and not the final solution). In this PR I evaluate the direct usage of the `ConfigurationManager` class to read/write user settings. 

- This approach fits well in our existing design. We just need to replace the IO-operations with the appropriate `ConfigurationManager` methods. That's all and really easy!
- The user setting file (user.config) is stored in folder AppData\Local\TestCentric\testcentric.exe_StrongName_l4olpu2u0j01dg13xxpj02ttammiohbl\2.0.0.0
This folder name looks suspicious, but it seems to be the common naming convention.

- The actual settings are stored as <key, value> pairs in the &lt;appSettings&gt; section. Here's a screenshot:
image
When getting familiar with settings I learned there's a difference between &lt;appSettings&gt; and &lt;applicationSettings&gt; and discussions about the pro/cons of each approach. Ultimately, &lt;applicationSettings&gt; is probably the newer approach which contains also type checking. I hope it will be used when evaluating the next approach.

- It may not look like it, but these three lines were not that easy. Only a lot of googling helped here and I'm still not convinced or didn't understand the concept in total why the first line isn't sufficient.
```
Configuration userConfigFile = ConfigurationManager.OpenExeConfiguration(ConfigurationUserLevel.PerUserRoamingAndLocal);
ExeConfigurationFileMap fileMap = new ExeConfigurationFileMap() { ExeConfigFilename = userConfigFile.FilePath };
Configuration configFile = ConfigurationManager.OpenMappedExeConfiguration(fileMap, ConfigurationUserLevel.None);
```

So, that's all for now - I want to keep this draft PR open to compare it with the other approaches

### DIFF
--- a/src/GuiRunner/TestModel/Settings/SettingsStore.cs
+++ b/src/GuiRunner/TestModel/Settings/SettingsStore.cs
@@ -8,12 +8,11 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.Globalization;
 using System.IO;
-using System.Runtime.Remoting.Contexts;
-using System.Text;
-using System.Xml;
 
 namespace TestCentric.Gui.Model.Settings
 {
+    using System.Configuration;
+
     /// <summary>
     /// SettingsStore implements the loading and saving of settings
     /// from an XML file and allows access to them via a key
@@ -25,7 +24,6 @@ namespace TestCentric.Gui.Model.Settings
 
         protected Dictionary<string, object> _settings = new Dictionary<string, object>();
 
-        private string _settingsFile;
         private bool _writeable;
         private object _myLock;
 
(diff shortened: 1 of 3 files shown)


CharliePoole:
Conceptually, this looks like a good approach. The Keys and Values are essentially the same as what we already store in TestCentricSettings.xml. 

It's hard to see any advantage to ApplicationSettings over AppSettings in our context but trying it out rather than just making an arbitrary decision is sensible.

I'll take a look at the code and failures as well.

src/GuiRunner/TestModel/Settings/SettingsStore.cs
```diff
@@ -8,12 +8,11 @@
...
 using System.Globalization;
 using System.IO;
-using System.Runtime.Remoting.Contexts;
-using System.Text;
-using System.Xml;
 
 namespace TestCentric.Gui.Model.Settings
 {
+    using System.Configuration;
+
```
CharliePoole: Slightly surprised you're using `SettingsStore` rather than replacing it, but it may make sense. We should be sure to eliminate anything from it not needed by the new implementation.

CharliePoole:
Failure on the pull request is due to the cake recipe not being found. Since the push build worked, it's possible that you may not have updated your local main before creating the branch from it.

rowo360:
I'm closing the draft PR now, as we prefer option 2 as the solution.

PR #1349, opened 2025-11-02
Use settings designer to store user settings

rowo360:
This is another draft PR for https://github.com/TestCentric/TestCentricRunner/issues/1342. Here I evaluate the usage of the setting designer to read/write user settings. Again it's intended to get an impression+overview about the approach in general. It's working successfully, but I didn't verified every detail yet - for example I'm aware that I broke the tests.

Essentially, this approach is identical to the approach using the ApplicationSettingsBase class (described [here](https://github.com/TestCentric/TestCentricRunner/pull/1347)). So many parts in that description applies also here. The key difference is that the setting class is generated here. To do this, you need to add a *.Settings file and in the code behind there's a *.Designer.cs file created. The UI looks like this - for example it's possible to add settings of type System.Drawing or TestCentric enums.

image

The generated class in the *.Designer.cs file is derived from `ApplicationSettingsBase` and uses attributes on the properties to define the scope and default values. Basically it's the same code I created manually in the previous approach.

image


rowo360:
The code generator creates partial classes, so if required we can extend a setting class with additional initialization or properties. For example the `RecentFiles` setting class has a setting property 'Entries' which is a list of all recent file names. In addition it has a property 'Latest' which access/modify the 'Entries' setting. That's still possible using this approach:

image

rowo360:
Closing this PR as we prefer option 2 as the solution.

PR #1355, opened 2025-11-19
Fix exception using SelectAgent menu

### LINKED ISSUE #1354: Crash when selecting the agent
**Describe the bug**
With the latest changes on our main branch there's a crash when selecting an agent using the main menu.

**To Reproduce**
1. Open test assembly targeting .Net 8.0
2. Open main menu 'Fiile' => 'Select agent' => 'Net90Agent'
=> Crash

In addition here's a screenshot from the crash location:

 - TestCentric Version: latest main branch
 - Target Runtime for your test assembly: .Net 8.0

CharliePoole:
Fixes #1354 

This actually fixes the original problem and two additional related problems that were hidden behind it. All of them relate to the use of the full name of the extension Type.

The basic problem is that `NUnit.Engine.TestAgentInfo` provides an `AgentName field` intended as a display name for the agent. We were using that value as the tag for each menu item, rather than the full name of the type, which is contained in the `ExtensionNode`. The key to this fix is that we are now ignoring the display name provided by the agent, which is OK for our existing agents but not a good recipe for the future.

Complicating matters, we are in the process of switching to use NUnit services in place of our own, so any changes we make will need to be ported to NUnit.    

@rowo360 In reviewing this, please try to suggest any longer term improvements. We may even want to implement a better solution on the TestCentric side now, for later porting to NUnit.

### DIFF
--- a/src/GuiRunner/TestCentric.Gui/Presenters/AgentSelectionController.cs
+++ b/src/GuiRunner/TestCentric.Gui/Presenters/AgentSelectionController.cs
@@ -84,13 +84,11 @@ namespace TestCentric.Gui.Presenters
                 EnsureSingleItemChecked(item);
 
                 // TODO: We currently need to use both settings. Investigate.
+                _model.TestCentricProject.RemoveSetting(SettingDefinitions.SelectedAgentName);
+                _model.TestCentricProject.RemoveSetting(SettingDefinitions.RequestedAgentName);
+
                 string itemTag = item.Tag as string;
-                if (itemTag == null || itemTag == "DEFAULT")
-                {
-                    _model.TestCentricProject.RemoveSetting(SettingDefinitions.SelectedAgentName);
-                    _model.TestCentricProject.RemoveSetting(SettingDefinitions.RequestedAgentName);
-                }
-                else
+                if (itemTag is not null && itemTag != "DEFAULT")
                 {
                     _model.TestCentricProject.AddSetting(SettingDefinitions.SelectedAgentName.WithValue(itemTag));
                     _model.TestCentricProject.AddSetting(SettingDefinitions.RequestedAgentName.WithValue(itemTag));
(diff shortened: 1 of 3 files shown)


rowo360:
My first thought was that we needed a DisplayName. But actually, it already exists with `TestAgentInfo.AgentName`.
So, it seems to me that all the necessary information is available, and the problem is more that we are working with strings in TestCentric from the ExtensionObject (TestAgengyInfo) the UI, but the setting (`SettingDefinitions.RequestedAgentName`) ultimately requires an identifier for the ExtensionNode.

So, two options came into my mind:
The `TestModel.AvailableAgents` returns a list of ExtensionNode objects. Each extensionNode object is assigned to the Tag property of the menu item. However the Text property of menu item uses the 'TestAgentInfo.AgentName' property by navigating from the extensionNode to the extensionObject. When selecting an item the extensionNode.TypeName can be set in the settings.

If we want to keep the ExtensionNode vs ExtensionObject handling out of the UI, another option might be to use a list of TestAgentInfo for the property `TestModel.AvailableAgents`. I would assign a TestAgentInfo to the Tag of a menu item. However when selecting an item we cannot set the settings directly, but need to query for the matching ExtensionNode object to retrieve its TypeName. Maybe that code can be located in the `TestAgentService`? For example: string GetAgentIdentifier(TestAgentInfo)
I think that's a good option.

Or a third option: the TestAgentInfo class can be extended to provide its identifier itself.

BTW:
I was wondering why we use the settings at all to define the requested agent. It's convient, but we are passing the value to the NUnit setting and we are reading it back from the NUnit setting to create the appropriate agent. NUnit itself doesn't require this setting for now. Am I'm right about this one? Of course, I'm not suggesting to change it, because I guess that would require a lot of changes. But just to confirm, that we are the only consumer of this setting currently and we can define it's content.
If sometime in the future, we pass the responsibilty for the agent creation to NUnit this might change of course.
Ok, just  want to share all my considerations... 😄

Issue #1360, opened 2025-11-25
Close test gap in package tests (x86 target)


rowo360:
**Describe the bug**
The analysis in issue #1356 shows up that there's a test gap in our package tests. We already have a wide range of package tests checking various combination of target frameworks+platforms, but the combination of that issue is missing.

So, the intention of this issue is to close this gap, so that we notice such kind of errors instantly and not via an issue from a user.

**Expected behavior**
A package test for .Net8.0 and x86 exists and is executed in our pipelines.

**Environment (please complete the following information):**
 - TestCentric Version 2.0 - beta 7
 - Target Runtime for your test assembly: .NET 8

PR #1361, opened 2025-11-26
Extned package tests for x86 platform

rowo360:
This PR solves #1360 by adding additional target frameworks to our test assembly **mock-assembly-x86**.
These tests are executed in our build scripts.

I removed some code parts in class `MockAssembly` which set the static string `AssemblyPath`. This change was required due to compile error using .Net core which complains that `Assembly.EscapedCodeBase` is not available. I could also introduce some preprocessor directives (\#if) for that purpose, but I came to the conclusion that the `AssemblyPath` is not used at all. So, I decided to remove this part.

Additional remark:
It's currently still unclear why this kind of tests failed using TestCentric 2.0 - beta 7. However luckily they are already succeeding using our latest developer builds, in which we switched to NUnit agents in the meantime. And the tests are executed successfully in our pipeline. Here's one part of the log file:
```
----------------------------------------
Run mock-assembly-x86.dll under .NET 7.0
----------------------------------------
```

### DIFF
--- a/package-tests.cake
+++ b/package-tests.cake
@@ -62,6 +62,22 @@ public static class PackageTests
             "net9.0/mock-assembly.dll",
             MockAssemblyExpectedResult("Net90AgentLauncher")));
 
+        GuiAndEngineTests.Add(new PackageTest(1, "Net60X86Test", "Run mock-assembly-x86.dll under .NET 6.0",
+            "net6.0/mock-assembly-x86.dll",
+            MockAssemblyX86ExpectedResult("Net60AgentLauncher")));
+
+        GuiAndEngineTests.Add(new PackageTest(1, "Net70X86Test", "Run mock-assembly-x86.dll under .NET 7.0",
+            "net7.0/mock-assembly-x86.dll",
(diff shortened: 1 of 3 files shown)


package-tests.cake
```diff
@@ -62,6 +62,22 @@ public static class PackageTests
             "net9.0/mock-assembly.dll",
             MockAssemblyExpectedResult("Net90AgentLauncher")));
 
+        GuiAndEngineTests.Add(new PackageTest(1, "Net60X86Test", "Run mock-assembly-x86.dll under .NET 6.0",
+            "net6.0/mock-assembly-x86.dll",
+            MockAssemblyX86ExpectedResult("Net60AgentLauncher")));
+
+        GuiAndEngineTests.Add(new PackageTest(1, "Net70X86Test", "Run mock-assembly-x86.dll under .NET 7.0",
+            "net7.0/mock-assembly-x86.dll",
+            MockAssemblyX86ExpectedResult("Net70AgentLauncher")));
```
CharliePoole: Since we don't have a .Net 6.0 or 7.0 launchers, I'd expect these to use the .Net 8.0 one.

PR #1362, opened 2025-11-26
Improve message when start logging without sufficient rights

### LINKED ISSUE #1358: Enhance message box when logging is requested without sufficient rights
**Describe the bug**
Testcentric supports some logging functionality to provide internal insights. This functionality is activated via a command line parameter and as a result a log file is created next to the TestCentric.exe file.

`testcentric.exe --trace=Debug`

However if the current user doesn't have write access in the current folder an exception is thrown during start-up. For example if installation was executed with admin rights (via chocolatey) and TestCentric is started as a common user.

Here's a screenshot of the exception:

**Expected behavior**
If the current user doesn't have sufficient rights to create a log file, we should display this information in a message box (in a user friendly manner). And disable logging for the session

 - TestCentric Version 2.0 - beta7
 - OS: Window 11
 - Target Runtime for your test assembly: .NET Framework 4.7.2,

rowo360:
This PR fixes #1358. The application checks on start-up if the user has sufficient rights to create a log file whenever logging was enabled. If not this new message box is shown, informing the user about the problem. Afterwards TestCentric is getting closed. 

So, overall instead of a message box with some internal info a more user friendly message box is shown

### DIFF
--- a/src/GuiRunner/TestCentric.Gui/AppEntry.cs
+++ b/src/GuiRunner/TestCentric.Gui/AppEntry.cs
@@ -52,6 +52,12 @@ namespace TestCentric.Gui
                 return 2;
             }
 
+            if (!ValidateLoggingOption(options))
+            {
+                MessageDisplay.Error($"Failed to enable logging. Might be due to missing access rights in folder {Environment.CurrentDirectory}. Please consider to start with admin rights.");
+                return 2;
+            }
+
             log.Info("Instantiating TestModel");
             ITestModel model = null;
             try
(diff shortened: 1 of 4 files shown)


CharliePoole:
@rowo360 I'd like to hold off on merging this until we figure out why it's happening. See my comment on #1358

src/GuiRunner/TestCentric.Gui/AppEntry.cs
```diff
@@ -113,6 +119,17 @@ public static int Main(string[] args)
             return 0;
         }
 
+        private static bool ValidateLoggingOption(GuiOptions options)
```
CharliePoole: Is the value not being validated within GuiOptions itself? ( can't remember)

If so, it may be better to do it there, but I'm OK with keeping it here for the time being... you  choice.

CharliePoole (approved):
Based on our discussion,  this looks fine but see my one inline question.

src/GuiRunner/TestCentric.Gui/AppEntry.cs
rowo360: In this method we validate that the user has write access to bin folder (only if logging is enabled) - so it's some additional validation besides what's done in the GuiOptions.

Issue #1396, opened 2025-12-18
Eliminate TestCentric Engine


CharliePoole:
We have been slowly reducing the content of the TestCentric Engine and relying on the NUnit Engine for more and more features. I've been doing this slowly in the background and recently we seem to have reached a tipping point, where it became much easier to phase out classes and interfaces. I'm now fairly sure this will be ready for the next beta release, so this issue is for the purpose of documenting the change.

This is a breaking change both in regard to previous betas and the version1 GUI. However, it's effect is limited to developers creating their own runners using our engine. I don't know of any myself but if they exist, they should switch to use of the NUnit engine. Folks who use the GUI directly should not be impacted.

@Row360 I'm expecting that this change will probably initially break a feature or two in the GUI, since some of our implementations add functionality to the NUnit engine. When I merge the final change to remove the engine, I'd like you to identify any problems that arise. We can initially just list them here. Then you and I can decide, issue by issue, whether to fix them in NUnit or compensate (if possible) in the test model.

I'm also thinking that it makes sense to change the repo back to it's original name, testcentric-gui, once we are only a gui. What do you think?

rowo360:
Here's another observation: it's about **test parameters**.
I added this kind of test case

<img width="500" src="https://github.com/user-attachments/assets/afbec8af-8769-4e5e-b256-729c9feba0e4" />

But although I defined the parameter in our test parameters dialog, the TestContext.Parameters dictionary is always empty during a test run. I already checked that the parameters are correctly set in the TestCentricProject settings. So, I think that the parameters are correctly passed to the NUnit engine/agent, but somehow get lost in between. I'll try to check it and will update this comment.

PR #1402, opened 2025-12-28
Fix bug when loading tcproj files

### LINKED ISSUE #1403: Error loading tcproj file
It's not possible to load a tcproj file: instead there's an error message box:

rowo360:
Fixes #1403 

This PR fixes #1403  (first mentioned in #1396 ([link](https://github.com/TestCentric/TestCentricRunner/issues/1396#issuecomment-3678559987) to comment):

It's not possible to load a tcproj file: instead there's an error message box:

@CharliePoole:
I'm not entirely sure about this fix, because it's based on searching through the commit history and looking on that changes. But in doing so, I came to realize that serializing/deserializing a TestPackage seems not be supported anymore or that we should not use it anymore. Instead there are two new methods in the `NUnit.Engine`: ToXml() and FromXml().

Saving a TestCentric project already used the ToXml(), but loading still uses the deserializing approach. Switching to helper method FromXml() solves this issue.

### DIFF
--- a/src/GuiRunner/TestModel/TestCentricProject.cs
+++ b/src/GuiRunner/TestModel/TestCentricProject.cs
@@ -98,26 +98,21 @@ namespace TestCentric.Gui.Model
         {
             ProjectPath = path;
 
-            using (StreamReader reader = new StreamReader(ProjectPath))
+            try
             {
-                XmlSerializer serializer = new XmlSerializer(typeof(NUnit.Engine.TestPackage));
+                string fileContent = File.ReadAllText(ProjectPath);
+                TestPackage newPackage = PackageHelper.FromXml(fileContent);
 
-                try
-                {
-                    var newPackage = (NUnit.Engine.TestPackage)serializer.Deserialize(reader);
+                foreach (var subPackage in newPackage.SubPackages)
+                    AddSubPackage(subPackage.FullName);
 
-                    foreach (var subPackage in newPackage.SubPackages)
-                    {
-                        AddSubPackage(subPackage.FullName);
-                    }
-
-                    LoadTests();
-                }
-                catch (Exception ex)
-                {
-                    throw new Exception("Unable to deserialize TestProject.", ex);
-                }
+                LoadTests();
             }
+            catch (Exception ex)
+            {
+                throw new Exception($"Unable to load TestProject from {path}", ex);
+            }
+
 
             IsDirty = false;
         }
@@ -133,22 +128,6 @@ namespace TestCentric.Gui.Model
             using (StreamWriter writer = new StreamWriter(ProjectPath))
                 writer.Write(this.ToXml());
         }
-
-        //public void Save(TextWriter writer)
-        //{
-        //    XmlSerializer serializer = new XmlSerializer(typeof(NUnit.Engine.TestPackage));
-
-        //    try
-        //    {
-        //        serializer.Serialize(writer, this);
-        //    }
-        //    catch (Exception ex)
-        //    {
-        //        throw new Exception("Unable to serialize TestProject.", ex);
-        //    }
-
-        //    IsDirty = false;
-        //}
 
         public void LoadTests()
         {


rowo360:
Here's an additional consideration: 
actually I wanted to add another test for this use case. Just because I think this is a really important use case which should be covered by a test. And as soon as it no longer works, a faulty test should reveal this.

That worked (in TestCentric.Gui.Model.Tests). But then I noticed that there is already a similar test and that this test runs successfully even without my changes. 😩  The problem with this test is that in this use case, the exception is thrown in NUnit.Engine, but we use a mock for the engine in the tests. This means we are not really testing the use case properly.

I came up with two ideas:
Either I can extend the engine mock so that it throws an exception and then check whether we are handling it correctly. This is good for a unit test, but not really what I want. Instead, for quality assurance, I would prefer an additional type of test: an integration test that tests certain top-level use cases with our components together with real NUnit.Engine and NUnit.Agents. Maybe this type of test might even include the TestCentric.GUI presenter logic, but the UI part should definitely not be included. So, besides loading/saving a project, we might even check if we can run/filter tests and that the result is as expected.

Do you think such kind of tests would be helpful or too much effort to create and maintain? Or not valuable or not in the sense of a pure unit test approach?

CharliePoole:
@rowo360 

I created issue #1403, which you are solving here. We need individual issues for anything we do in TestCentric to compensate for our use of the nunit engine vs the testcentric engine becaues  that's the only way the change will be recorded as a bug fix in the release notes. I also added the new issue to the list of sub-issues in  #1396, the "mother" issue.

src/GuiRunner/TestModel/TestCentricProject.cs
```diff
@@ -98,27 +98,22 @@ public void Load(string path)
         {
             ProjectPath = path;
 
-            using (StreamReader reader = new StreamReader(ProjectPath))
+            try
             {
-                XmlSerializer serializer = new XmlSerializer(typeof(NUnit.Engine.TestPackage));
+                string fileContent = File.ReadAllText(ProjectPath);
+                TestPackage newPackage = PackageHelper.FromXml(fileContent);
```
CharliePoole: Just make sure that `TestCentricProject` doesn't have any additional members, not handled by the `FromXml` method. I don't think it does but I can't check very easily right now.

rowo360: I checked all `TestCentricProject` members and it's fine! For example member `IsDirty` is not saved, but saving that one is indeed not intended.

rowo360:
I'm going to merge this PR now because the original issue is fixed. But while looking further around another question pops up. While looking into the content of a tcproj file I was a bit surprised about all the settings. Here's an example:
```
<TestPackage id="2">
  <Settings ShadowCopyFiles="True" InternalTraceLevel="Off" MaxAgents="0" TestParametersDictionary="&lt;parms>&lt;parm key='MyParameter' value='ABC' />&lt;/parms>" />
  <TestPackage id="3" Fullname="D:\NUnit4\MyLibrary.Test.Core\bin\Debug\net8.0\MyLibrary.Test.Core.dll">
    <Settings ShadowCopyFiles="True" InternalTraceLevel="Off" MaxAgents="0" ImageRuntimeVersion="4.0.30319" ImageTargetFrameworkName=".NETCoreApp,Version=v8.0" TargetFrameworkName=".NETCoreApp,Version=v8.0" SelectedAgentName="Net80AgentLauncher" />
  </TestPackage>
</TestPackage>
```

@CharliePoole:
Of course we add all those settings to the TestPackage in order that they are passed properly to the NUnit.Engine. And as a consequence it's OK, that these settings are persisted into the tcproj file. However it's a bit confusing because we are saving some settings (ShadowCopyFiles or MaxAgents) already in our TestCentric settings. But I think that's also ok, because some user might configure these values specific for one tcproj and use a different set of values for another tcproj.

If you can confirm my considerations and also think that saving all these settings is correct, then I will check again that when loading a tcproj file, the settings from the tcproj are correctly applied. For example a few days back we thought that the SelectedAgent is always the Default item when loading a project. Now, we should check that the value from the tcproj is properly restored.

CharliePoole:
> I'm curious to see if the idea with the acceptance tests could work. So, I can start an experiment by creating a new test project and adding some tests. Just to get a first impression, if that will cover such use cases.

That would be great. For inspiration, I suggest taking a look at how Terje has implemented acceptance tests as a separate project under the NUnit3 VS Adapter.

Issue #1408, opened 2026-01-01
TopLevel TestPackage settings are not restored when loading tcproj file


rowo360:
**Describe the bug**
This bug description applies only to top level TestPackage settings (for example: **RunAsX86** or **TestParameters**).  If I configure these settings once, save+close my project and load the project again, these settings are lost. However when I open the tcproj file to inspect it's content (txt editor), I can observe that these values are properly saved in the file.

**To Reproduce**
1. Open test assembly
2. Add some test parameters (using dialog invoked from Toolbar)
3. Save + Close testcentric project
4. Load testcentric project
5. Open Test parameter dialog => no parameters are available

**Expected behavior**
Test parameters should be restored when loading a test project

**Environment (please complete the following information):**
 - TestCentric Version: 2.0 beta (current main branch)

PR #1409, opened 2026-01-01
Fix applying package settings

### LINKED ISSUE #1407: Change package settings - subview is not updated
**Describe the bug**
The TestPackage subview displays all current settings with their values. Here's a screenshot:

However when changing one of these package settings in the settings dialog, this subview isn't updated but still displays the old value. Here's a list of the relevant values: MaxAgent, ShadowCopy and PrincipalPolicy.

**To Reproduce**
1. Load one test assembly and observe current values in TestPackage subview
2. Open settings dialog and change one value (for example ShadowCopy)
3. Confirm 'reload' message box
4. Observe TestPackage subview again => old value is still displayed

**Expected behavior**
This subview should update it's content to reflect the changed settings

 - TestCentric Version: 2.0 beta (current main branch)

rowo360:
This PR fixes #1407. Although the issue description sounds like a UI update issue, it turned out that the package settings were not properly applied.

The solution consist of two important parts:

- Usage of `PackageSetting` objects instead of <string, object> pairs to apply a setting
- Use of method `TestPackage.AddSetting(TestPackage)` instead of method `Settings.Set(key, value)`

The first part helps to avoid boxing of value types - so that a bool value if properly set as a bool type and not of type object(bool). 
The second part is necessary so that that a setting is not only applied as a top-level setting, but applied also to all subpackages. 
The package setting sub view, in turn, requests the values directly from the subpackages, so it will display the changed values right away.

Some further details are described in the next comments.

### DIFF
--- a/src/GuiRunner/TestModel/GuiOptions.cs
+++ b/src/GuiRunner/TestModel/GuiOptions.cs
@@ -7,10 +7,8 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Reflection;
 using System.Text;
 using System.Text.RegularExpressions;
-using static System.Windows.Forms.VisualStyles.VisualStyleElement.Window;
 
 namespace TestCentric.Gui
 {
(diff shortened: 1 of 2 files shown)


rowo360:
One benefit of using PackSetting objects is that it allows us to call TestPackage.AddSetting(TestPackage). This method applies a setting to all subpackages, so we don't need to perform this task on our own. However, please notice the new keyword - additional to the implementation in class TestPackage we need to set the IsDirty flag.
```
public new void AddSetting(PackageSetting setting)
```

rowo360:
Good suggestions - I've incorporated the code review comments:

- The settings dialog contains two lists now:
`TopLevelPackageSettingChanges` and `SubPackageSettingChanges`
The individual settings pages can add their settings to either of these lists.
- The TestCentricProject class has two methods to apply a setting:
`SetTopLevelSetting(PackageSetting setting)` and `SetSubPackageSetting(PackageSetting setting)`
Depending on the individual setting the appropriate method must be invoked.

rowo360:
I'm going to merge this PR now, so that I can continue with the related task (#1408). Nonetheless I can add additional changes afterwards if requested.

Issue #1214, opened 2025-03-24
Clarify use of Settings vs VisualState file


CharliePoole:
@rowo360 Although we have discussed this before, I ran into some problems while working on issue #1207.

I'm creating this issue first as a "Design Issue", so we can discuss how we want the Settings file to work. If we decide to change something, we'll change it to a feture or enhancement and one of us can take it on.

Here's what my current `TestCentricSettings.xml` file looks like:

```
<NUnitSettings>
  <Settings>
    <Setting name="TestCentric.Gui.MainForm.Location" value="10, 10"/>
    <Setting name="TestCentric.Gui.MainForm.Size" value="700, 400"/>
    <Setting name="TestCentric.Gui.MainForm.SplitPosition" value="290"/>
    <Setting name="TestCentric.Gui.RecentProjects.File1" value="C:\Users\charlie\dev\TestCentric\testcentric-gui\bin\Release\net462\mock-assembly-x86.dll"/>
    <Setting name="TestCentric.Gui.RecentProjects.File2" value="C:\Users\charlie\dev\TestCentric\testcentric-gui\bin\Release\net35\mock-assembly.dll"/>
    <Setting name="TestCentric.Gui.RecentProjects.File3" value="C:\Users\charlie\dev\TestCentric\testcentric-gui\bin\Release\net462\mock-assembly.dll"/>
    <Setting name="TestCentric.Gui.TestTree.DisplayFormat" value="NUNIT_TREE"/>
    <Setting name="TestCentric.Gui.TestTree.ShowFilter" value="True"/>
    <Setting name="TestCentric.Gui.TestTree.ShowNamespace" value="True"/>
    </Settings>
</NUnitSettings>
```

There are two types of settings present:

1. Settings intended to be transparent to the user, which control things like what the position of the main window. They are intended to be used for all projects, whenever the app is opened and cannot be set directly by the user, but only indirectly, e.g. by moving the window.
2. Settings that control the options within the app to be used by default. In this case, the last three settings are of this type. These settings are the ones I'm wondering about.

A bit of history: The V1 GUI as well as the original NUnit GUI itself had a lot of this type of setting. For V2, many of them were removed Currently, it's up to the GUI to pass the correct settings to the engine for how things are run. Some of this comes from VisualState, some from Settings, some from the user and some is hard-coded. It can be hard to figure out where options and settings come from.

[...]

PR #1410, opened 2026-01-02
Restore top level settings when loading package

### LINKED ISSUE #1408: TopLevel TestPackage settings are not restored when loading tcproj file

rowo360:
This PR fixes #1408 by restoring all top level package settings when loading a package.

The core change is located in method `TestCentricProject.Load()`. It contains additional these statements now, which are responsible to apply all top level settings from the loaded package to the current (this pointer) package.
```
foreach (PackageSetting packageSetting in newPackage.Settings)
    Settings.Set(packageSetting);
```

Besides this change I need to slightly adapt the UI as well: the checked state of the RunAsX86 menu item, must be updated from the package settings whenever we load a project. This worked almost straight forward, besides that a reload is invoked whenever the checked state of this menu item is changed. So, I need to add some additional statements to avoid this unnecessary reload operation here.

One remark to the changes in method TestCentricPresenter.ChangePackageSettingAndReload():
I simplified that one: the 'DEFAULT' case could be removed because it's for the AgentSelection, but that is handled by the AgentSelectionController class in the meantime. Also I removed the settings.Remove() case, because we don't remove any settings in other locations, but simply applying the values. Hope, that's ok!

### DIFF
--- a/src/GuiRunner/TestCentric.Gui.Tests/Presenters/Main/CommandTests.cs
+++ b/src/GuiRunner/TestCentric.Gui.Tests/Presenters/Main/CommandTests.cs
@@ -84,7 +84,6 @@ namespace TestCentric.Gui.Presenters.Main
             _model.Received().OpenExistingProject(file);
         }
 
-        [Test]
         [TestCase(null)]
         [TestCase("")]
         public void OpenTestCentricProjectCommand_NoFileSelected_DoesNotCreateProject(string fileName)
@@ -224,8 +223,6 @@ namespace TestCentric.Gui.Presenters.Main
             _model.Received().ReloadTests();
         }
 
-
-        [Test]
         [TestCase(true)]
         [TestCase(false)]
         public void RunAsX86CheckedChanged_SettingIsAppliedToProject(bool isChecked)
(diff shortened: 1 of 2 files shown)


CharliePoole (approved):
Since this solves your issue, I'm OK with it. Let's try not to make too many changes to the design of `TestCentricProject` until we complete #1214. I just noticed I never labeled it to indicate it's importance and did so now. I'd like to resolve the design definitively for beta8 and implement as much of what we decide as possible. I'll add a new comment to sum up and restart the discussion.

PR #1411, opened 2026-01-03
Add separate settings dialog to configure 'Test Run settings' 

### LINKED ISSUE #1406: Settings displayed in 'Assembly load settings' dialog differ from test run
**Describe the bug**
This issue is about some special settings which are stored in the TestCentric settings **and** also in TestPackages (TestCentricProject file (tcproj)). Here's a list of these special settings:
- MaxAgents
- ShadowCopyFiles
- PrincipalPolicy

When loading a tcproj file the values for these settings are applied to the TestPackage settings successfully, but the TestCentric settings remain unchanged. So, the values for these settings differ now and depending which setting is requested a different value is returned. One example, is the settings dialog which shows the values stored in the TestCentric settings. However this might be misleading for the user, because they might differ from the TestPackage settings, which are used for the next test run.

If the user change such a setting in the settings dialog, it will be applied to the TestCentric setting **and** TestPackage setting. So, both settings are in sync again - that's fine! But the initial display in the settings dialog is misleading.

**To Reproduce**
1. Set MaxAgents to 2 
2. Save TestCentric project
3. Create another TestCentric project 
4. Disable MaxAgents (set to 0)
5. Save TestCentric project
6. Load first TestCentric project
7. Open settings dialog => MaxAgents is set to 0; however test run is executed with MaxAgents=2

**Expected behavior**
The settings dialog should display the values used by the TestPackage for the next test run.

 - TestCentric Version 2.0 main branch

rowo360:
This PR fixes #1406.

Bases on the discussion in the issue, I split this task into two commits:

1.  Removal of these TestCentric application settings: ShadowCopy, MaxAgents and PrincipalPolicy
That means there are no user specific default values anymore, when creating new projects. But instead the NUnit default values are always applied to new projects. Only the setting ShadowCopy is always enabled for new projects (NUnit default is disabled). As a consequence, the settings dialog is retrieving the values from the test project now, instead from the application settings.
2. In order to make it clearer to the user when which settings are saved, I split the settings dialog. The 'Assembly load setting' setting page, which configures those three settings, is removed from the general settings dialog. Instead it's displayed via a new menu item 'Test Run Settings..' 

Here are two screenshots:

I can well imagine that we will also transfer the configuration of setting RunAs86 and SelectedAgent to this new dialog. Then these related settings would all be together. But that can be done at a later point in time.

### DIFF
--- a/src/GuiRunner/TestCentric.Gui.Tests/Presenters/Main/WhenTestsAreLoaded.cs
+++ b/src/GuiRunner/TestCentric.Gui.Tests/Presenters/Main/WhenTestsAreLoaded.cs
@@ -49,6 +49,7 @@ namespace TestCentric.Gui.Presenters.Main
         [TestCase("ForceStopButton", false)]
         [TestCase("SaveResultsCommand", false)]
         [TestCase("TransformResultsCommand", false)]
+        [TestCase("TestRunSettingsCommand", true)]
         public void CheckCommandEnabled(string propName, bool enabled)
         {
             ViewElement(propName).Received().Enabled = enabled;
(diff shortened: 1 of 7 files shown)


rowo360:
Ups, because of our exchange, I completely forgot about this PR - so I merge it, now.

CharliePoole:
I'll wait until after we do the PR for #1413. If you want to do that one, you can assign it to yourself. There is some degree of risk of conflict with #1414 but I don't think it's too high as the major changes will in the model rather than the UI. There will be some for the one who merges last of course. :-)

Issue #1413, opened 2026-01-04
Reduce the scope of the Settings Dialog


CharliePoole:
The Settings Dialog provides a way to record user decisions about the functioning of TestCentric, global decisions, which are used for every project. As discussed in #1412, we want to radically reduce the size of this dialog. ..

* Some settings will be removed because we don't actually support them any longer.
* Other settings are no longer needed because NUnit provides satisfactory default values.
* Still other settings are accommodated in other places in the GUI and are redundant.

We want to end up with a Settings Dialog which has only...

1. A small number of settings used only to initialize new projects.
2. A small number of GUI-related settings, which are not convenient to accommodate elsewhere and are most likely to be generally desirable, rather than varying by project.

Settings only used to initialize new projects should be on a separate page and clearly indicated as such.

Specific decisions will be made as the code is written and reviewed in the PR.

rowo360:
Here is a suggestion for the implementation that does not turn everything upside down. It's more of a gradual evolution, and we can stop at different steps. I like that kind of approach much more.
1. The setting class `TestTreeSettings` will contain properties for the initial default values, but also for the actual values. For example, InitialShowCheckBoxes and ActualShowCheckBoxed, InitialDisplayFormat and ActualDisplayFormat or ActualGroupBy (supports no initial default value).
While the properties names 'Initial...' are real setting property in the sense of the ApplicationSettingsBase class with these attributes:

```
        [UserScopedSetting]
        [DefaultSettingValue("false")]
        public bool InitialShowCheckBoxes
```

the properties named 'Actual...' are simple c# properties without these attributes. This means, they will not be stored in the user.config file - that's exactly what I wanted. However, I want that any changes to these simple c# properties still invokes the SettingChanged event. That will not work out-of-the-box, but I think that can be done without much effort (those changes would only be located in the settings classes). The great thing is that all clients of this setting can continue to work as before (set/get setting or listen to SettingChanged event).

2. This step is only a minor modification: in step 1 the class `TestTreeSettings` contains a mixture of real settings and actual values. Maybe it's better to separate them into separate classes: the class `TestTreeSettings` continues to manage the real settings. But we introduce a new class `TestTreeConfiguration` (we can discuss the naming - I just want to avoid the term setting for now) and this class contains all the actual values. However, I would still propose to keep this new class accessible under the Model.Settings properties. The SettingChanged event is still supported. For example, clients can access these properties by:
[...]

PR #1417, opened 2026-01-10
Settings dialog: add configuration of default tree settings

rowo360:
This PR contributes to #1413, but it's currently in draft state to support discussion about the different approaches. The discussion might even result into abonding this PR and try out a different approach.

So, for a start I just changed the UI of the settings dialog. It looks like this:

The user can configure only the "ShowCheckBox" and "DisplayFormat" as his default values for new projects. Everything else (the logic and settings) remain unchanged for now and need to be adapted next.


rowo360:
Here’s the commit containing my proposal to reduce the visual tree settings. I skipped the intermediate steps I tried to describe earlier and went straight to my proposed target state. You can, of course, view the changes line by line, but I will try to summarize the key points in the following comments.

PR #1427, opened 2026-02-07
Add high resolution system icons for message box

### LINKED ISSUE #1426: Blurred icon in MessageBox
**Describe the bug**
The icon in the message box looks blurry. This is especially the case when working with a 4k monitor. Here are two screenshots from a 4k display:

(See also comments in PR #1423)

**Expected behavior**
The icons should be clearly drawn on all kind of displays.

 - TestCentric Version 2.0 beta (current main branch)

rowo360:
This PR fixes #1426 by providing some "high" resolution (48x48) icons for the message box. These icons are embedded resources, so there's no need to copy/deploy them to the target directory. In addition, they are displayed identically on all displays.

These icons were extracted from a sample Windows forms app. That app has High DPI support enabled and requests the SystemIcons.Warning/Errror/.. and save them to disk. So, these icons are original Windows icons and not downloaded from somewhere else.

Here's a screenshot for comparision: the old message box (2 weeks ago) and the current one


src/GuiRunner/TestCentric.Gui/MessageDisplay.cs
```diff
@@ -26,7 +26,7 @@ public class MessageDisplay : IMessageDisplay
         private static readonly IMessageDisplay DefaultDialog = new MessageDisplayForm(DEFAULT_CAPTION);
 
         public static void Error(string message, string caption = DEFAULT_CAPTION) =>
-            DefaultDialog.Info(message, caption);
```
rowo360: I noticed this one: the Error message box showed an Info icon. So, I switched it to an error icon.

rowo360:
I can still change the .Net framework from 4.6.4 to 4.8. However, it was not required anymore, so I keep it unchanged for now.

src/GuiRunner/TestCentric.Gui/MessageDisplay.cs
CharliePoole: Hah! I never noticed that. Good catch.

Issue #1418, opened 2026-01-11
Automatically save TestCentricProject and VisualState at same time


CharliePoole:
The `TestModel` automatically creates a `TestCentricProject` whenever it is created. In the case of a pre-existing project, it is loaded from the saved `.tcproj` file so that the top level `TestPackage`, containing all sub-packages and settings are present. In the case of a new project, an empty `TestCentricProject` and `TestPackage` may be created or it may be populated, depending on the content of the `GuiOptions` and on properties provided b the GUI.

The user is able to save the project manually or may be invited to save it when closing the application if the project content is seen to have changed using the `IsDirty` property. We have discussed offline the fact that the `IsDirty` property is unreliable and cannot easily be made reliable because of how the underlying NUnit engine code works. Some limitations are also imposed by .NET itself.

There are some approaches that could be used to make `IsDirty` work but they are generally quite complex. Before we try that I'd like to experiment with saving at all times automatically, without asking the user, as we already do with the file containing `VisualState` for the project.

Proposal...
1. Save the `.tcproj` file whenever we save the `.VisualState.xml` file.
2. Make this operation transparent to the user.
3. Stop checking `IsDirty` and prompting the user to save when the application closes.

@rowo360 
I'd like to treat this as an experimental feature, which we will add to our main branch. For development purposes, we'll use this code to work on the GUI and see if problems arise. If we hit problems, we'll modify the code and eventually we'll release it in the next beta. While I can imagine situations where some user might prefer to have full control over saving the file, I think they may actually be better off if we save it when it needs to be saved so that the application behavior is completely consistent. What do you think?

Note: This will be a breaking change with respect to version 1 of the GUI.

CharliePoole:
Currently, `VisualState` is saved in the following situations

1. In response to the `TestsUnloading` event, sent when the tests are about to be unloaded.
2. In response to the `TestsReloading` event, sent when the tests are about to be reloaded.
3. In response to the `RunStarting` event, sent when a test run is about to start, 

The unloading event occurs when exiting the application as well as at other times when tests are unloaded through the UI.

The current `TestCentricProject` is saved
1. In response to the user requesting it to be saved through the UI.
2. In response to the user closing the Project.
3. In response to the user closing the application if the project is dirty and the user responds OK to the prompt that is displayed.

To match the proposed actions in the description, the following changes are needed...
1. In response to the `TestsUnloading`, `TestsReloading` and `RunStarting` event2, both files should be saved.
[...]

PR #1429, opened 2026-02-08
Fix crash when applying setting

### LINKED ISSUE #1428: Changing "Shadow copy" setting results into a crash
**Describe the bug**
There's a crash when changing the "Shadow copy" setting.
(Same applies to changing the "Principal policy" setting)

**To Reproduce**
1. Open settings (File => Test run settings)
2. Change "Shadow copy" settings
3. Press OK
=> Crash

 - TestCentric Version: 2.0 beta (main branch)

rowo360:
This PR fixes #1428:

If a setting already exists, we must first remove it before we can set it. There exists already a method `TestCentricProject.RemoveSetting()` for this purpose. It removes the setting from the top level package itself and all subpackages. So, we invoke this method in all `TestCentricProject.AddSetting(...)` methods now.

The method `RemoveSetting()` succeeds even in case the setting is not present.

### DIFF
--- a/src/GuiRunner/TestCentric.Gui.Tests/Dialogs/SettingsDialogBaseTests.cs
+++ b/src/GuiRunner/TestCentric.Gui.Tests/Dialogs/SettingsDialogBaseTests.cs
@@ -33,6 +33,25 @@ namespace TestCentric.Gui.Dialogs
         }
 
         [Test]
+        public void ApplySettings_SettingValueIsChanged_IsAppliedToProject()
+        {
+            // 1. Arrange
+            ITestModel model = Substitute.For<ITestModel>();
+            TestCentricProject project = new TestCentricProject();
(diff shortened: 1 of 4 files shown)


CharliePoole (changes requested):
Currently `TestCentricProject.AddSetting` works just the same as `TestCentricProject.TopLevelSetting.AddSetting`. I did this deliberately to avoid confusion that would arise (for me anyway) if the two `AddSetting` methods were to have different semantics. So I'd prefer not to change it this way. Sorry I didn't think of it in our earlier corrrespondence.

How about a new method, `TestCentricProject.ReplaceSetting` to do what you want?

CharliePoole (approved):
Looks good. Thanks! FYI, IsDirty will probably go away after #1418 is done, but I'll take care of that in the PR.

PR #1431, opened 2026-02-10
Synchronize saving of TestCentricProject and VisualState

### LINKED ISSUE #1418: Automatically save TestCentricProject and VisualState at same time

CharliePoole:
Fixes #1418

### DIFF
--- a/src/GuiRunner/TestCentric.Gui/Presenters/TestCentricPresenter.cs
+++ b/src/GuiRunner/TestCentric.Gui/Presenters/TestCentricPresenter.cs
@@ -9,6 +9,7 @@ using System.Drawing;
 using System.IO;
 using System.Linq;
 using System.Windows.Forms;
+using NUnit;
 using NUnit.Engine;
 using TestCentric.Gui.Dialogs;
 using TestCentric.Gui.Elements;
--- a/build.cake
+++ b/build.cake
@@ -14,7 +14,7 @@ BuildSettings.Initialize(
 	context: Context,
 	title: "TestCentric.GuiRunner",
 	solutionFile: "TestCentric-Gui.sln",
-	githubRepository: "testcentric-gui",
+	githubRepository: "TestCentric-Gui",
 	exemptFiles: ["Resource.cs", "TextCode.cs"]
 );
 
--- a/src/GuiRunner/TestCentric.Gui/AppEntry.cs
+++ b/src/GuiRunner/TestCentric.Gui/AppEntry.cs
@@ -9,8 +9,8 @@ using System.IO;
 using System.Linq;
 using System.Text;
 using System.Windows.Forms;
+using NUnit;
 using NUnit.Engine;
-using TestCentric.Gui.Dialogs;
 using TestCentric.Gui.Model;
 using TestCentric.Gui.Presenters;
 using TestCentric.Gui.Views;
@@ -51,11 +51,11 @@ namespace TestCentric.Gui
                 return 2;
             }
 
-            if (!ValidateLoggingOption(options))
-            {
-                MessageDisplay.Error($"Failed to enable logging. Might be due to missing access rights in folder {Environment.CurrentDirectory}. Please consider to start with admin rights.");
-                return 2;
-            }
+            //if (!ValidateLoggingOption(options))
+            //{
+            //    MessageDisplay.Error($"Failed to enable logging. Might be due to missing access rights in folder {Environment.CurrentDirectory}. Please consider to start with admin rights.");
+            //    return 2;
+            //}
 
             log.Info("Instantiating TestModel");
             ITestModel model = null;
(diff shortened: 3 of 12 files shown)


src/GuiRunner/TestCentric.Gui/Presenters/TestCentricPresenter.cs
```diff
@@ -167,33 +169,33 @@ void UpdateTitlebar()
...
-
-                _view.MessageDisplay.Error("Test assemblies must target a specific platform, rather than .NETStandard.");
-                if (_lastFilesLoaded == null)
-                    _view.Close();
-                else
-                {
-                    _model.UnloadTests();
-                    _model.CreateNewProject(_lastFilesLoaded);
-                }
-            };
```
rowo360: TestLoadFailures: Is this only temporary uncommented? Maybe you can add a comment what needs to be done to enable it again? Just so that we can recall it later on.
Or if this code part not needed anymore, we can remove it right away.

src/GuiRunner/TestCentric.Gui/Presenters/TreeViewPresenter.cs
```diff
@@ -300,8 +300,7 @@ private void WireUpEvents()
```
rowo360: I didn't spot any code which is calling method SaveVisualState() ? Those event handlers in this class (Unloading/Reloading/RunStarting) are invoking _model.SaveVisualState(...) directly. We could either call method SaveVisualState() in these event handlers or remove that method.

src/GuiRunner/TestModel/TestCentricProject.cs
```diff
@@ -23,18 +23,67 @@ public class TestCentricProject
...
-                options = new GuiOptions();
+            Guard.ArgumentNotNullOrEmpty(projectPath, nameof(projectPath));
 
+            ProjectPath = projectPath;
             TestFiles = [.. options.InputFiles];
             TopLevelPackage = new TestPackage(TestFiles);
 
-            // Turn on shadow copy in new TestCentric project by default
-            AddSetting(SettingDefinitions.ShadowCopyFiles.WithValue(true));
+            // TODO: Default setting needs to be moved elsewhere
```
rowo360: I'm bit worried that enabling the 'ShadowCopy' option gets lost in this way. At the moment this means that we disable ShadowCopy by default for all new projects! What about uncomment the AddSetting(SettingDefinitions.ShadowCopyFiles.WithValue(true)) again and keep the TODO statement. Or another option might be to take care about the TODO statement now. Same concern in other constructor.

src/GuiRunner/TestCentric.Gui/Presenters/TreeViewPresenter.cs
CharliePoole: I'll remove it.

src/GuiRunner/TestModel/TestCentricProject.cs
CharliePoole: You're right, I'll remove the commenting and also create an issue for this and other places where we are making policy decisions (e.g. all new projects need shadowcopy enabled) in the constructor. This belongs either in the caller or centralized in the model (probably best) as a list of things we do for all projects.

CharliePoole:
> I finally finished my review - but I already noticed that you started to comment on them 😄 Looks fine - we'll close the gaps with the next tasks.

Yes... tomorrow!

src/GuiRunner/TestCentric.Gui/Presenters/TreeViewPresenter.cs
CharliePoole: Done

PR #1444, opened 2026-02-24
Eliminate grouping on NUnit Tree

### LINKED ISSUE #1436: Drop Grouping from NUnit Tree Display Strategy
Start by removing the menu items and incrementally remove all code no longer used or needed. This functionality will instead be provided under the Test List strategy as part of a separate issue.

CharliePoole:
Fixes #1436


rowo360:
One fun fact which I also didn't recall: 😄 
I rediscovered myself that the NUnit tree groupings were actually internally based on a test list. For each test the appropriate group is determined and the tree hierarchy is build up. I never came up with the idea to sell it to the user as a test list or call it myself a test list.
But never mind, we continue as agreed: if it turns out that some parts can be solved similar for the test list grouping we can use it, if we want to solve some challenges differently, we'll do so. That's all fine 😄

CharliePoole:
Thanks for checking. I'll get rid of those classes as well.

"Test List" started as a criticism of Microsoft, i.e. "They don't understand the hierarchical nature of NUnit tests, the just deal with it like a list of tests." But in the end I thought if people want a list grouped by various properties, we can do that too.

Issue #1438, opened 2026-02-22
Add Show/Hide Assembly and Fixture options to Test List display.


CharliePoole:
These items may be individually checked and are separate from the selection of a grouping.

Grouping items for assembly and fixture will be removed, leaving only ungrouped, category, duration and outcome.

By default, assembly and fixture will not be shown

Tests will be displayed as follows:

Each Assembly (if Assembly is shown shown)
      Each group label (unless ungrouped)
            Each Fixture containing items in the group (If Fixtures are shown)
                Each test contained

@rowo360 Please review to see if this matches your understanding of what we discussed. Of course, it may change as the work proceeds.

PR #1453, opened 2026-03-02
NUnit tree display: add option to show/hide assemblies and fixtures nodes

### LINKED ISSUE #1437: Add Show/Hide Assembly and Fixture options to NUnit Tree display
These will join the already existing Namespace option. All three of these default to showing the related item.

rowo360:
This PR fixes #1437 by providing additional menu items for the nunit tree display.

All these options are enabled by default for new projects, so that by default the entire NUnit tree will always be displayed. 
These options are also stored in the VisualState file so when reloading a project the previous tree configuration is restored.

### DIFF
--- a/src/GuiRunner/TestModel/TestNode.cs
+++ b/src/GuiRunner/TestModel/TestNode.cs
@@ -69,6 +69,9 @@ namespace TestCentric.Gui.Model
         public bool IsProject => Type == "Project";
         public bool IsNamespace =>
             IsSuite && (Type == "TestSuite" || Type == "SetUpFixture");
+        public bool IsFixture => IsSuite && 
+            (Type == "TestFixture" || Type == "GenericFixture" || Type == "ParameterizedFixture" || Type == "Theory");
+
 
         /// <summary>
         /// Controls if the TestNode should be visible or hidden in the TestTree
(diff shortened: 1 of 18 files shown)


src/GuiRunner/TestModel/TestNode.cs
```diff
@@ -69,6 +69,9 @@ public TestNode(string xmlText) : this(XmlHelper.CreateXmlNode(xmlText)) { }
         public bool IsProject => Type == "Project";
         public bool IsNamespace =>
             IsSuite && (Type == "TestSuite" || Type == "SetUpFixture");
+        public bool IsFixture => IsSuite && 
+            (Type == "TestFixture" || Type == "GenericFixture" || Type == "ParameterizedFixture" || Type == "Theory");
+
```
CharliePoole: In the work I'm doing for #1438, I have also added `TestNode.IsFixture`. My definition is simply

```
public bool IsFixture => IsSuite && Type == "TestFixture";
```
Both could be considered "correct" definitions. I eliminated `GenericFixture` and `ParameterizedFixture` because They would each add two extra levels to the TestList display and I only wanted to add the actual instantiated fixture. This makes sense for me, because I'm _adding_ extra levels and your approach makes sense for you because you are _hiding_ levels that are normally displayed. I'll try to resolve this difference when I merge my work with yours.

However, I think that `Theory` is a different matter. In the normal display, the user should only be seeing the Theory, not the test cases beneath it. This is how both NUnit 2 and TC 1 worked. The individual cases are supposed to be hidden unless the user explicitly displays them. I think we may have lost this in V2 however. I'll double-check and create an issue for it.

So, for the moment, I think this is all OK, subject to later change. That's the benefit you get by being quicker than me to make a PR. :-)

Issue #1447, opened 2026-02-24
Cannot compile test assembly while loaded in TestCentric


rowo360:
**Describe the bug**
I develop my NUnit tests by opening the test project in Visual Studio to write test code and by opening the test assembly in TestCentric to run the tests. Whenever some new tests are ready, I want to run them to check that everything is working correctly. However, I cannot compile the test project anymore as soon as the test assembly is loaded in TestCentric. Visual Studio complains that the test assembly is blocked by another process (TestCentric) and cannot replace it. So I need to unload the test assembly in TestCentric, compile in VisualStudio and reopen test assembly in TestCentric again.

**To Reproduce**
1. Load test assembly in TestCentric
2. Load test project in Visual Studio
3. Edit test code and try to compile
4. Compile error is reported that assembly is blocked by another process

**Expected behavior**
I would appreciate it if TestCentric would not block any loaded test assembly. Then I could switch seamlessly between developing and executing tests. I don't expect that a test assembly isn't blocked while tests are executed - at this point of time the blocking is ok. But while the tests are simply displayed in the tree, the test assembly should not be blocked.

**Environment (please complete the following information):**
 - TestCentric Version: TestCentric 2.0 Beta 7
 - Target Runtime for your test assembly: NET Framework and .NET Core

PR #1454, opened 2026-03-04
Dispose runner object after each operation to unblock loaded assemblies quickly

rowo360:
Here's a **draft PR** to share a proof of concept for the idea mentioned in #1447.
We want to try a solution on TestCentric side that ensures a test assembly is not loaded permanently, but only for as long as necessary during an operation. This idea is in contrast to a solution on NUnit side, which  requires more effort and coordination.

So, here's the essential idea:
- When loading a test assembly/project a runner object is created, the tests are explored and finally the runner is disposed right away. So, while the tests are only displayed in the tree, there exists no runner object anymore.
- When starting a test run, a runner object is created anew, it's kept in memory until the test run is finished and finally it's disposed on the TestRunFinished event. That means that during a test run, the test assembly is loaded (and blocked) - but I think that's acceptable for the user.

Overall, this approach works fine so far - I didn't notice any drawbacks. I can observe that the agent process is created while exploring/running, but also that the process is terminated successfully afterwards. All use cases work as intended. Of course, there must be a small performance delay, because we load the test assembly a second time when starting a test run, but I couldn't notice it.

From code perspective I moved the relevant code parts from class `TestModel` to a new class `TestCentricRunner`. (I don't even want to know how many classes you've seen named with the terms ‘test’ and ‘runner’, but I haven't been able to come up with a better name so far. 😄 ) It was simply an opportunity to extract one part from the large `TestModel` class.
The code line in the new class themselves for exploring/running are taken from 'TestModel' class besides that a new runner is created+disposed for every explore operation. And that there's a runner object during a test run which is disposed on the TestRunFinished event.

I was even able to remove the code that took care if the last run was a debug run and forced a reload if required - well, now the tests are always reloaded before a test run, so this check regarding the debug flag is not required anymore. I think that we might remove additional code lines, if we want to continue with this idea: for example, triggering a reload in case the user changes some settings (RunAsX86, SelectedAgent...) is not required anymore - or maybe we can even remove the entire 'ShadowCopy' setting completely. But all that can be considered later.

Overall, I think this idea looks promising without any complex solution. Therefore I want to share the progress with you. We can discuss the pros/cons in one of our next meetings and decide if we want to continue with this idea.

### DIFF
--- a/src/GuiRunner/TestModel/TestCentricRunner.cs
+++ b/src/GuiRunner/TestModel/TestCentricRunner.cs
@@ -8,9 +8,14 @@ namespace TestCentric.Gui.Model
     using System;
     using System.Threading.Tasks;
     using NUnit;
-    using NUnit.Common;
     using NUnit.Engine;
 
+    /// <summary>
+    /// The TestCentricRunner class wraps the NUnit Engine's ITestRunner interface to explore and run tests.
+    /// While an assembly is loaded by the NUnit ITestRunner, it's blocked by the file system and the assembly cannot be modified by the user.
+    /// However, the user should be able to continue developing their tests while the tests are already shown in the GUI runner.
+    /// Therefore, the TestCentricRunner minimizes the time that the ITestRunner is active by creating and disposing it as needed for each test run.
+    /// </summary>
     internal class TestCentricRunner : IDisposable
     {
         static Logger log = InternalTrace.GetLogger(typeof(TestModel));
@@ -70,6 +75,9 @@ namespace TestCentric.Gui.Model
             return loadedTests;
         }
 
+        /// <summary>
+        /// Runs the tests in the specified package.
+        /// </summary>
         public void RunAsync(TestPackage package, NUnit.Engine.TestFilter filter)
         {
             Guard.ArgumentNotNull(package, nameof(package));


src/GuiRunner/TestModel/TestCentricRunner.cs
```diff
@@ -0,0 +1,105 @@
...
+            }
+            finally
+            {
+                runner.Dispose();
+            }
+
+            return loadedTests;
+        }
+
+        public void RunAsync(TestPackage package, NUnit.Engine.TestFilter filter)
```
CharliePoole: Test execution needs to be tested with randomly generated test parameters. I believe you'll find that the reloading of the test package will actually change the values used and, consequently, the names of the tests that use such parameters. This was a problem we had to resolve for the Visual Studio adapter because Visual Studio does very much the same thing you are doing here. If you can verify that this is a problem, i can explain the solution we used for the adapter.

CharliePoole (changes requested):
I think this is a good start. The issue with random parameters is my main concern and I have added a few other questions and comments.

rowo360:
Actually, this setting doesn't seem to work with the NUnit console at the moment either. I'll write a NUnit issue for this:
```
nunit-console.exe MyLibrary.Test.dll -seed 1000
```
Executing this command line generates a result file containing different test parameters.

rowo360:
That random-seed attribute seems to be missing. Just loaded an assembly without any further handling,

rowo360:
Thanks for creating the NUnit issue and sorry, that it requires many iterations for solving this issue.
I'm using method `TestCentricProject.ApplySetting()` to set the RandomSeed setting. Which in turn removes the setting first and apply it to all subpackages afterwards.

I decided to implement this a little differently after reviewing my proposal again a few hours later. 😄 So, I just did a new commit. But the general idea remains unchanged: while loading/reloading the `TestModel `invokes a new method `TestCentricProject.InitRandomSeed()`. A new seed is applied to the settings which will be remain constant for all subsequent test runs. Only when a new project is created, loaded or reloaded a new seed is initialized.

If necessary, we can also add a condition for reloading later.

PR #1465, opened 2026-03-14
Add context menu to remove entries from recent files list

### LINKED ISSUE #1458: Allow deleting entries from the RecentFiles / RecentProjects menus
We already manage the recent files list by deleting non-existing files. It seems desirable to also allow the user to right click on a menu item and delete the entry. While this would be low priority for most users,  I found I really wanted it when testing the application. 

@rowo360 What do you think?

rowo360:
This PR fixes #1458 by providing a context menu for the recent files menu items. Here's a screen shot:

From technical point of view WinForm provides two context menu approaches: the 'old' `ContextMenu` class and the 'new' `ContextMenuStrip` class, I decided to stick with the `ContextMenu` class, because opening a `ContextMenuStrip` forces all other menu items to close (especially the RecentFileList). The context menu looks a little lonely there - there are some tricks to keep the other menu items open, but it makes it more complex. There are no such effects using the `ContextMenu` class.

Another idea was to use the keyboard (Delete key) to remove an entry. However, a ToolStripMenuItem has no KeyDown/Up event. Instead you can register a KeyDown event on a parent element. Now you know which key is pressed on the menu, but you don't know which menu item is pointed with mouse cursor. That doesn't sound like a simple approach either, so I dropped it.

### DIFF
--- a/src/GuiRunner/TestCentric.Gui/Presenters/TestCentricPresenter.cs
+++ b/src/GuiRunner/TestCentric.Gui/Presenters/TestCentricPresenter.cs
@@ -420,16 +420,11 @@ namespace TestCentric.Gui.Presenters
                         entries.RemoveAt(index);
 
                 menuItems.Clear();
-                int num = 0;
-                foreach (string entry in entries)
+
+                for (int i=0; i<entries.Count; i++)
                 {
-                    var menuText = string.Format("{0} {1}", ++num, entry);
-                    var menuItem = new ToolStripMenuItem(menuText);
-                    menuItem.Click += (sender, ea) =>
-                    {
-                        string path = ((ToolStripMenuItem)sender).Text.Substring(2);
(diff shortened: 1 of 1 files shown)


CharliePoole:
I'm not clear why the contextmenu on each recent file can't be set once and for all and left there, rather than creating and removing it while tests are running. Can you explain this a bit more?

CharliePoole:
Strange... this is what I was looking at...

image

As you see, GitHub seems to be saying this code is part of the `OnRunFinished` method. I only glanced at it and didn't understand why you would be changing this at the end of the run. But when I look more closely at the entire file I can see that this executes on popup.

Anyway... this looks good.

CharliePoole (approved):
Thanks, this will be a big help in experimenting with the code, removing test files, etc.

PR #1466, opened 2026-03-15
Remove reload of tests on setting change

### LINKED ISSUE #1464: Remove 'force reload' when changing setting (SelectedAgent, MaxAgent...)
**Describe the bug**
This is a follow-up of #1447 : Test assemblies are now loaded only temporarily, while a test is running or while exploring the tests. For each test run a NUnit test runner is created to execute the tests.

This new approach makes it possible to simplify the setting handling. For certain settings (SelectedAgentName, MaxAgents, ...) we request a reload operation when changed. This reload is not required anymore, because we are reloading implicitly on next test run.

So, the idea is to remove the reload on setting change completely.

rowo360:
This PR fixes #1464.

I don't have much to say about the changes. Once we agree on the issue in general, the changes contain mainly only removing lines of code.


CharliePoole (approved):
Looks good...

PR #1468, opened 2026-03-17
Add ShowAssemblies and ShowFixtures to TestListStrategy

### LINKED ISSUE #1438: Add Show/Hide Assembly and Fixture options to Test List display.

CharliePoole:
Completes and Fixes #1438

### DIFF
--- a/src/GuiRunner/TestCentric.Gui/Presenters/TreeViewPresenter.cs
+++ b/src/GuiRunner/TestCentric.Gui/Presenters/TreeViewPresenter.cs
@@ -568,7 +568,6 @@ namespace TestCentric.Gui.Presenters
             // If a test is already running, no new test run should be started.
             _view.RunContextCommand.Enabled = _model.HasTests && !_model.IsTestRunning;
             _view.DebugContextCommand.Enabled = _model.HasTests && !_model.IsTestRunning;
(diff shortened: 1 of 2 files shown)


src/GuiRunner/TestCentric.Gui/Presenters/TreeViewPresenter.cs
```diff
@@ -270,36 +272,6 @@ private void WireUpEvents()
 
             _view.ResetFilterCommand.Execute += () => ResetTestFilter();
 
-            // TODO: Verify that this is no longer needed and remove code
```
rowo360: I needed to  check how this is solved with these code lines uncommented. But I noticed there's another event registration which handle all of this. So, it's good to remove it now.

PR #1475, opened 2026-03-22
Catch exceptions while reading invalid VisualState files

### LINKED ISSUE #1450: Exceptions thrown reading VisualState
When reading the visual state file, unhandled exceptions sometimes occur. This happens in particular when running the package tests, which may use the same filenames repeatedly. This needs to be fixed in two ways...

1. Identify the source of the errors and fix them.
2. Handle all exceptions that occur by simply logging them and not using the visual state.

rowo360:
This PR fixes #1450.

Basically I added a catch statement to handle all kind of exceptions while reading a VisualState file. In case of an exception the error is logged and null is returned as a return value.

I was tempted to move the `File.Exists(fileName)` statement into the `VisualState.LoadFrom(string)` method. Currently it's called outside before invoking this method. Also, I was tempted to return an empty, default VisualState object in case of an error. So, that callers don't need to care about the null return value. 

But that would require a few more changes – nothing too complicated, but before I get started, I’d like to hear your thoughts about it.

I ran all of our tests (Gui.Model.Tests and Gui.Tests) to detect any exceptions during VisualState file reading. But I didn't spot any occurrence or any workflow which creates corrupt VisualState file and which needs to be fixed.

### DIFF
--- a/src/GuiRunner/TestCentric.Gui.Tests/VisualStateSerializationTests.cs
+++ b/src/GuiRunner/TestCentric.Gui.Tests/VisualStateSerializationTests.cs
@@ -131,6 +131,36 @@ namespace TestCentric.Gui
             }
         }
 
+        [Test]
+        public void LoadFrom_NonExistingFile_ReturnsNull()
+        {
+            var visualState = VisualState.LoadFrom("NotExistingFile.xml");
+
+            Assert.That(visualState, Is.Null);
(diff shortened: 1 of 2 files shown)


CharliePoole (approved):
This looks good to me and I think it's a sufficient fix unless something else comes up.

Thanks.

Issue #1441, opened 2026-02-22
Allow opening a single file without overhead of a project


CharliePoole:
In fact, opening the file is the easiest part. We will open it and create a project named like `test.dll.tcproj`.

Questions that arise include
* Should that project initially be visible to the user in the title bar or via some notification?
* If we don't make it visible in any way, should the user be told that we are saving the project when we close it?
* When a test file is added, we need to create a project automatically. Are there other points where this would happen?

@rowo360 Any thoughts on the above questions? Any others we need to answer?

CharliePoole:
As a part of this, we need to fix a current problem with saving recent files. The following code is found in `TestModel.LoadTests`...

```
            if (TestCentricProject.ProjectPath == null)
                foreach (var subPackage in TopLevelPackage.SubPackages)
                    RecentFiles.Latest = subPackage.FullName;
            else 
                RecentFiles.Latest = TestCentricProject.ProjectPath;
```

However, in the current codebaase, the initial test on `ProjectPath` always fails, so only .tcproj files are saved.

@rowo360 FYI

rowo360:
I noticed that I missed to reply on your proposal in good time - sorry. What’s more, I think every single suggestion is a good one - let's do it this way. 😄 
Some of the items on the list are already working as described:

- [x] 1. When the user opens an assembly, rather than a TC Project, we should create a "secret" project in the form test.dll.tcproj.
- [x] 2. The title bar should show the name of the assembly, i.e. test.dll, rather than the name of the project.
- [x] 3. On closing, we should save both the "secret" project and the visual state without any further input.
- [ ] 4. When the user opens an assembly and the project and/or visual state are present, we should use them.
If the VisualState file is present, it will be used - that's fine. However the presence of a test,dll.tcproj file is not supported yet.
- [ ] 5. On closing, the name of the assembly - not the project - should be added to recent files.
Currently, both files are added to the recent file list: test.dll and test.dll.tcproj
- [ ] 6. When the user adds a second assembly, we should throw them into a creation dialog.
That's missing - a "Not implemented" message box shows up.
- [x] 7. When the user removes assemblies and reduces the project to containing only one assembly, we should NOT revert to a "secret" project but should continue to use the already saved project name.
I think we've not done anything specific to achieve this. But it already works as described. 😄

PR #1479, opened 2026-03-29
Open existing file (dll/exe) - add fileName to recent file list (but not tcproj file)

rowo360:
This PR contributes to #1441, by fixing item no 5:
On closing, the name of the assembly - not the project - should be added to recent files.

So far both file names (assembly name and project name) were added to their recent file list. To ensure that only the assembly name is added in this case, it was sufficient to remove this single line. After that, the RecentFileList is always reliably set in the `TestModel.LoadTests()` method - that code part is unchanged and thus not included in the commit.

Whether a new project is being created, or a project or assembly is being loaded (including via command-line arguments), this method `TestModel.LoadTests()` is always called and the RecentFileList is already updated. There's no need to update it again in method SaveProject() - especially since this method is also called to save the *.dll.tcproj file.

Only when we implement ‘Save as’ later, we need to take care about the RecentFileList handling.

### DIFF
--- a/src/GuiRunner/TestModel.Tests/TestModelTests.cs
+++ b/src/GuiRunner/TestModel.Tests/TestModelTests.cs
@@ -282,6 +282,32 @@ namespace TestCentric.Gui.Model
         }
 
         [Test]
+
+        public void OpenExistingFile_TestCentricProject_IsSet_AsRecentFileLatest()
+        {
+            // Arrange
+            var xmlNode = XmlHelper.CreateXmlNode($"<test-case id='1' name='TestA' />");
+
+            var runner = Substitute.For<ITestRunner>();
+            runner.Explore(null).ReturnsForAnyArgs(xmlNode);
+            var engine = Substitute.For<ITestEngine>();
+            engine.GetRunner(null).ReturnsForAnyArgs(runner);
+            var options = new GuiOptions();
+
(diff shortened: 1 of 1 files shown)


src/GuiRunner/TestModel.Tests/TestModelTests.cs
```diff
@@ -53,22 +55,6 @@ public void RemoveTestPackage_TestCentricProjectLoadedEvent_IsTriggered()
...
-            var options = new GuiOptions("dummy.dll");
-            var model = TestModel.CreateTestModel(engine, options);
-
-            // Act
-            model.CreateNewProject("TestCentric");
-            model.SaveProject("TestCentric.tcproj");
-
-            // Assert
-            Assert.That(model.Settings.Gui.RecentFiles.Latest, Is.EqualTo("TestCentric.tcproj"));
-        }
```
CharliePoole: Do we really want to remove this? Is there another test that shows a `.tcproj` becoming the most recent?

CharliePoole (approved):
Looks good to me, but make sure you really want to delete that one test before merging.

CharliePoole:
FYI, I'm working on item 6 and there is some overlap, but I can pull this into my branch once this is merged. Biggest change is that there is now a separate method for creating a "wrapper" project.

PR #1489, opened 2026-04-02
Remove ShadowCopy option from settings dialog

### LINKED ISSUE #1463: Remove ShadowCopy setting ?
**Describe the bug**
This is a follow-up of #1447: Test assemblies are now loaded only temporarily, while a test is running or while exploring the tests. But they don't stay loaded while tests are displayed in the tree.

We can now decide whether we still need the `ShadowCopy` setting. I think the original idea for ShadowCopy was also to allow the user to edit and recompile while a test is running. Or did I miss something?

rowo360:
This PR fixes #1463 by removing the ShadowCopy settings from the settings dialog.
In addition the settings dialog was improved to hide the settings tree in case only one single setting page is displayed.

Here's are two screenshots (before and after the changes) for comparision:

New version:

I also considered to adapt the window height, but didn't came up with a general solution fitting for all kind of setting pages. So, I decided to keep the blank part for now


rowo360:
Thanks, for the clarification - I'll remove that statement now.

PR #1494, opened 2026-04-03
Add "Save as" functionality

### LINKED ISSUE #1491: Enable "Save as" functionality
**Describe the bug**
The file menu already contains a "Save as" menu item, but it doesn't execute any action so far. 
Now that the general project handling has been sorted out, this functionality should be implemented.

"Save as" is intended to save a copy of a TestCentric project file to a different folder and/or different filename. The original TestCentric project file is kept unchanged.

rowo360:
This PR fixes #1491 by providing the "Save as" functionality.

- The save dialog is initialized with the folder + filename of the current project; a shortcut if the user wants only minor adaptations
- After the save operation the title bar is updated and the new project is added to the recent file list


CharliePoole (changes requested):
Unless there's a bug in `TestCentricProject`, I think there are changes to make... well actually, if there's a bug there are even more changes. :-)

PR #1499, opened 2026-04-07
Display TestGroup full name in properties view

### LINKED ISSUE #1478: Display information about Selected TestGroup in TestProperties view.
When the selected item is a TestNode, the TestProperties view displays a good deal of information but when it is a group, only the name of the group is shown. A bit more information would make sense.

For a TestNode, there are two panels. The info about the package is only visible for top-level Test Files while the properties panel is always shown. I think we need a third panel, visible only for groups, which shows a bit more information about the group, perhaps only a description (which we would have to add  to each grouping) and a count of the tests included in the group.

rowo360:
This PR fixes #1478. 

The only remaining issue was that we wanted to display the full name for a test group. Since the `TestGroup.ParentGroup` property is now available, the full name can be assembled by navigating through the parent hierarchy.
Here's a screenshot:

The root TestGroup node (Category, Duration, Outcome or 'all tests' in case of none grouping) is included in the full name. It's separated by a ':' however.

### DIFF
--- a/src/GuiRunner/TestCentric.Gui/Presenters/TestPropertiesPresenter.cs
+++ b/src/GuiRunner/TestCentric.Gui/Presenters/TestPropertiesPresenter.cs
@@ -124,7 +124,7 @@ namespace TestCentric.Gui.Presenters
 
         private void InitializeTestGroupPropertiesSubView(TestGroup testGroup)
         {
-            _view.TestGroupPropertiesSubView.FullName = testGroup.Name;
+            _view.TestGroupPropertiesSubView.FullName = GetTestGroupFullName(testGroup);
             _view.TestGroupPropertiesSubView.TestCount = testGroup.TestNodes.Count().ToString();
         }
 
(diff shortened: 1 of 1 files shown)


src/GuiRunner/TestCentric.Gui/Presenters/TestPropertiesPresenter.cs
```diff
@@ -204,6 +204,19 @@ private string GetTestProperties(TestNode testNode)
...
 
+        private string GetTestGroupFullName(TestGroup testGroup)
+        {
+            string fullName = testGroup.Name;
+            while (testGroup.ParentGroup != null)
+            {
+                testGroup = testGroup.ParentGroup;
+                string separator = testGroup.ParentGroup != null ? "." : ": ";
+                fullName = testGroup.Name + separator + fullName;
+            }
```
CharliePoole, line 216: Since `testGroup.ParentGroup` is never null in the loop, I'd do...

```
while (testGroup.ParentGroup != null)
    fullName = $"{testGroup.Name}.{fullName}";
```

... or something equivalent.

CharliePoole (changes requested):
Aside from the one change request you could consider adding a `FulName` property to `TestGroup` and using the same logic there. That way it would be available to any caller. Otherwise, I'd do the change as soon as another caller needs it.

No need to re-review this whatever you do.

rowo360:
I simplified the method to determine the full group name.

PR #1504, opened 2026-04-11
Stop removing non-empty groups from tree

### LINKED ISSUE #1503: Outcome/Duration grouping: tests in tree are incomplete after test run
**Describe the bug**
When grouping by outcome and duration, tests must be re-assigned to a new group during the test run. This works fine when all tests are executed. However if only a subset of tests is executed, then some tests are missing in the tree view. 

Here's a screenshot in which 'FixtureWithTestCases' is executed. All other tests from the group 'Not run' are missing afterwards.

**To Reproduce**
1. Select 'test list' view and group by outcome
2. Select a fixture for test execution
3. Run selected tests

CharliePoole:
Fixes #1503


rowo360 (approved):
Great—that was easy to fix! 🚀

PR #1512, opened 2026-04-14
Labels and tooltips to supplement icons

### LINKED ISSUE #1510: Show tooltip when hovering over result icon
For better accessibility, we should show a tooltip when hovering over a result icon, indicating whether it indicates Passed, Failed, etc. This is not needed in cases where the icon is already accompanied by a label, but only where it stands on its own.

These icons occur in various places in the GUI...

- In the tree itself (tooltip needed)
- In the test results tab for suites (already labelled, so no tooltip needed)
- In the status bar (tooltip needed)
- In the test Filter Display (tooltips already present)
- On the Tree Display page of the Settings Dialog (tooltips or labels needed)
- Other?

In the case of the tree, it may be necessary or desirable to apply the tooltip to the node itself. Some experimentation will be needed here.

CharliePoole:
Fixes #1510

### DIFF
--- a/src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
+++ b/src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
@@ -125,6 +125,7 @@ namespace TestCentric.Gui.Presenters
                 foreach (TreeNode treeNode in GetTreeNodesForTest(result))
                 {
                     treeNode.Text = GetTreeNodeDisplayName(result);
+                    treeNode.ToolTipText = result.Label != null ? result.Label : result.Status.ToString();
                     _view.SetImageIndex(treeNode, imageIndex, true);
                 }
             });
@@ -211,9 +212,11 @@ namespace TestCentric.Gui.Presenters
             {
                 case RunState.Ignored:
                     imageIndex = TestTreeView.IgnoredIndex;
+                    treeNode.ToolTipText = testNode.RunState.ToString();
                     break;
                 case RunState.NotRunnable:
                     imageIndex = TestTreeView.FailureIndex;
+                    treeNode.ToolTipText = testNode.RunState.ToString();
                     break;
             }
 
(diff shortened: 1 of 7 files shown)


CharliePoole:
@rowo360 This is complete with one exception...

When the text of a node is too long to fit within the width of the TreeView control, the tooltip showing the the complete name is now replaced by the result tooltip. Only one tooltip at a time is possible. You can see this by hovering over any name that overflows both before running the tests and afterwards.

This can be resolved by either showing both the name and pass/fail status in the same tooltip or by displaying the complete name using an overlaid expanding label or TipWindow, similar to what we do with extra text in the TestProperties tab.

I thought it best to discuss what approach we would like to use before coding it.

rowo360 (approved):
The actual code changes are all fine, and I don't need to comment on anything.

I propose to simply display the result state first and afterwards the node name in the same tooltip in the case the treeview control width is too small.

I've also noticed that I need to get used to the tooltips first. They're perfect for their purpose, especially for people with color blindness. Or by explaining that a test already has a result icon before any test is excuted, becaused it's an "ignored" or "Not runnable" test. But I also wonder whether they annoy me or bother me while navigating through the tree. For example if I want to set the checkbox of a node and the tooltip is in between.

I probably just need to get used to it, but you can keep an eye on it yourself.

CharliePoole:
In case you didn't notice, the tooltip gives more info than the icon in a few cases. So something skipped may say just "Skipped" or possibly "Explicit" if that's the reason it was skipped. I made a lot of little decisions like that one on the fly as I wrote it, so we may want to pay attention to the precise content of the tooltips as well.

Also, both the tooltip with the name and the one I just added make use of the TreeView controls built-in processing of tooltips. We could take over the display of tooltips more completely, as I've done elsewhere. Then we would be able to fine tune the position of the tips, the time delay before displaying and the delay after moving away. So let's watch all this and see how we feel.

For the combined tooltip, I'll see how easy it is to discover whether the name is not fitting. That's another thing done under the covers by the control. If it's not straightforward, I'll merge this and do another PR. I'm feeling good doing things in small pieces right now.

UPDATE: I'm merging this to allow more time to work on the name display problem.

PR #1513, opened 2026-04-16
Use TipWindow in place of built-in tooltip

### LINKED ISSUE #1510: Show tooltip when hovering over result icon

CharliePoole:
Fixes #1510 

This PR introduces use of a TipWindow, overlaid on the tree view, in place of the built-in toolstrip.

After @rowo360 commented that the tooltips seemed distracting, I decided to try this approach. The TipWindow is, on the one hand, more prominent in the display than the tooltip but, on the other hand, seems less distracting to me because...

* It consistently appears in one location.
* It overlays the text of the TreeNode but repeats that text with the addition of a text representation of the test result
* It serves simultaneously as a display of the test result and of the complete text of the node when the node doesn't fit in the client area of the TreeView

The initial commit only works fully for the NUnitTree display. Using the TestList, test cases are correctly displayed but the TipWindow does not appear for Groups. Addition of groups will require some deeper work. We probabl need notifications of when groups start to execute and finish, just as we now have for TestCases and Suites.

@rowo360 I've asked for your review on this. In addition to a code review, I think it would be best if you built the branch and actually exercised it. We should agree on whether this is a good UI approach before commiting to it. Note that the appearance and positioning of the TipWindow is capable of being modified, although I like overlaying it and using the standard background color that appears in other uses of it in the GUI.


rowo360:
I build the branch and still need to take a closer look at it. But at first glance, something doesn't quite seem right:



You don't want the red rectangles around the text either, do you?
And here’s an initial idea: would it help if the tooltip didn’t appear immediately, but only after a delay? For example, if the mouse cursor hovers over a node for 1.5 seconds. That would introduce more complexity, but would improve user experierence.

CharliePoole:
The green and red outlines are in Debug only. There's a comment in the code about why I left them in - i.e. 'cause they are handy in figuring out what's going on. They could be removed  or commented if we go this way.

For timing, there's currently a 500ms delay, which is easy to increase as it's just a property setting. TipWindow is fairly sophisticated BUT adapting it to TreeView took a lot of work because TreeView wants to do much of the tool tip stuff on its own.

I'm a little puzzled by the multiple ToolTips that are visible. I can't make that happen. Did they just appear by hovering and never go away?

UPDATE: I made it happen. If you move to another node, the first one closes. But if you move directly out of the control itself, it remains visible IFF the TipWindow is overlaying the control. This was intended only for use with labels, but was generalized in the  TipWindow, which needs to be changed.

CharliePoole:
In commit 2, I removed the red and green rectangles and fixed the code so that only one TipWindow appears at a time.

Looking at the code again, there's only a built-in delay for _closing_ the window. I have that set to 500ms, although it now gets closed immediately if you hover over a different item. Obviously, the delay before opening the window would have to be in the client that creates the TipWindow, i.e. TreeViewPresenter. Try it some more and let me know if I should add that - it's a complication, but only a small one.

CharliePoole:
Commit 3 adds a delay so that you can select a TreeNode or open the context menu before the TipWindow is displayed.

rowo360:
That feels much, much better now 😄 
This is because, during common tree operations (expand, collapse, double-click, checkbox, context menu), the tooltip does not appear immediately, so you are less likely to be distracted. I still haven't looked at the code at all; I've just been experimenting with the branch.

Here some observations:

- If a tree node is selected (blue box), showing the tooltip will deselect the node. Hiding the tooltip will select it again, That’s a bouncing effect (back and forth). Don't know, if we can keep the node selected.
- The tooltip is shown on top of the node, so the node is not visible anymore. If we can move it a bit right+down, it behaves more like a common tooltip.
- It occurs that the tooltip remains open. I slowly hoovered over several nodes, never stopping. The tooltip gets displayed for some of the nodes - I assume that's already not intended, because my mouse cursor is already further on. On the last node I wait a short period of time, but leave the node before the tooltip gets displayed. The tooltip gets displayed, but my mouse cursor is already somewhere else and the tooltip doesn't get closed automatically. 
I hope you manage to reproduce otherwise I'll create a short video.

But apart from these implementation details, I think the tooltip is fine. We want to display additional information, so the tooltip is a good option. Other options might be a double-click (however already used for running tests) or some entry in the context menu. None of them is better.

On the other hand, one could also argue that we are merely presenting information that is already available elsewhere. Once the node is selected, the test result view contains the node name and the result state. After all, the Result view is actually there to display all the properties of the selected node relating to the test results, isn't it? Hmm, don't know, what's best now?
(There’s no doubt that the tooltip would be just the thing in the MiniView. 😄 )

CharliePoole:
Thanks... Can we now say that this is the better choice as compared to standard tooltips, which we use, for example, in the status bar? I felt something different was needed here because of the information density of the tree. Up to now, I've been treating this as a prototype, so maybe it's better that you didn't look at the code. :-) 

The code for TipWindow is some of the oldest in the codebase. I wrote it for the NUnit GUI more than 20 years ago. Aside from not using modern techniques and current naming conventions, it assumes too much about how it's used. 
It was first written and used for ExpandingLabel, which I wrote at the same time. A bit later it was extended to apply to items in a ListBox. There's still code in there that assumes it is overlaying a Label. I "fixed" some of this by adding adhoc code - some of it way back then and some just now. There's more adhoc code in the timer I just created.

So my next commit will do an overhaul of TipWindow as well as the code added in the tree display, without (hopefully) changing any functionality. At that point, I hope this PR will be ready to review and commit. I'll deal with Groups separately. As I already mentioned, there's a lot involved there... I'm thinking we need new events for Group Starting and Group Finished.

About overlaying the existing entry... two thoughts... (1) When the full text doesn't fit, I think overlaying it makes more sense than offsetting it. It seems better to do the same thing for the result as well. (2) Offsetting it would hide something else.

For your other two points, I'll check it out.

Is this overkill? Well, I don't know. I don't suffer from color-blindness, but I'd hate to tell someone who does that they have to select each node in order to understand the icon. On the other hand, we may be able to make the icons clearer, so who knows. I wouldn't object to a setting that turned the text display of results on and off, so it would only appear when the full name was not visible in the window. But let's get the icons done first before we make that judgement.

CharliePoole:
@rowo360 I think this is now ready for review.

I ended up simplifying TestGroup rather than complicating it! It no longer stores an index or the current group ResultState. Instead, it's all done at the level of the TreeNodes.

src/GuiRunner/TestCentric.Gui/Presenters/TreeViewPresenter.cs
```diff
@@ -280,18 +282,18 @@ private void WireUpEvents()
                 {
                     _mouseHoverDelayTimer.Stop();
                     _mouseHoverDelayTimer.Interval = MOUSE_HOVER_DELAY;
-                    _mouseHoverDelayTimer.Tick += (s, e) => 
-                    {
-                        _mouseHoverDelayTimer.Stop();
-                        OnTreeNodeMouseHover(s, args);
-                    };
+                    _mouseHoverDelayTimer.Tick += (s, e) => OnTreeNodeMouseHover(s, args);
```
rowo360: I think this line will register to the Tick event over and over again - whenever the mouse is hovered above a node. So, when the Tick event is fired, OnTreeNodeMouseHover is executed multiple times. This fits to my observation that there's a lot of flickering when the tooltip shows up.

src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
```diff
@@ -346,6 +345,11 @@ private void ApplyResultsToTree(TreeNode treeNode)
                 ResultNode resultNode = GetResultForTest(testNode);
                 if (resultNode != null)
                     treeNode.ImageIndex = treeNode.SelectedImageIndex = CalcImageIndex(resultNode);
+
+                // NOTE: the loop only executes at the point in the hierarchy where the current
+                // TreeNode represents a TestNode and the parent is a TestGroup;
+                for (var parent = treeNode.Parent; parent?.Tag is TestGroup; parent = parent.Parent)
+                    parent.ImageIndex = parent.SelectedImageIndex = Math.Max(parent.ImageIndex, treeNode.ImageIndex);
             }
```
rowo360, line 344: I assume this will work fast although the ImageIndex of a root group is set over and over again. Of course that's not an optimal approach, but a reliable one. And at least we don't have namespace nodes in the test list, which would enlarge the tree hierarchy. But we should check (after merge), if this cause any performance impact.

In general I like the simplification of the TestGroup class. But in the long run I worry if it's a clean approach. We need to ask some UI elements (TreeNode) to get information about a result state. Unfortunately there's no entity around beside the UI which have all information about the displayed tree structure and provide this kind of information. I've in mind some tree model (not TestNode) for the displayed tree (grouped + filtered). But that's some larger task. 
Overall, we're coupling the result calculation strongly to WinForm. I'm probably not telling you anything new, and I'm sure you've noticed it too. But I wanted to highlight that again. It’s fine for now, but at some point it will need to be separated.

src/GuiRunner/TestCentric.Gui/Presenters/TreeViewPresenter.cs
CharliePoole: Ouch! I'm using the same Timer that `TipWindow` was using the mouse leaving, i.e. `System.Windows.Forms.Timer`! I'll switch all the timers to `System.Timers.Timer` and use AutoReset so they only fire once.

rowo360:
This exception has occurred once or twice:


I haven't been able to reproduce it reliably so far. But I suggest I just test it again after the timer fix.

CharliePoole:
I put in the Guard call to detect unattached nodes because they have a null Parent field. The TreeView property is getting set to null for some reason. 

FYI, I wasn't able to use System.Timers.Timer because of threading issues. We have to stick with Windows Timer because it runs its events on the UI thread, where we need it to be.

The whole thing is a can of worms, so I've lowered the priority but I'll keep working on it.

rowo360:
Oh understand - that's just too bad. I actually wanted to let you go first with this PR, but now, I will send out a review later.

CharliePoole:
I'll look at your review and see if it gives me a new path forward. One thing I'm considering is to clone `TipWindow` as `TreeViewTipWindow`, remove all my tree-related additions from `TipWindow` and focus the new class just on it's use with `TreeView`. That would work for a release and could eventually lead to a base class, possibly generic, with customized derived classes for use with `Label`, `ListBox` and `TreeView`.

CharliePoole:
@rowo360 This is ready for code review and testing. The PR failure appears to be due to an infrastructure problem rather than the code itself. I'll look into it tomorrow.

src/GuiRunner/TestModel/TreeConfiguration.cs
```diff
@@ -41,8 +41,11 @@ public string DisplayFormat
             get => _displayFormat;
             set
             {
-                _displayFormat = value;
-                OnPropertyChanged(nameof(DisplayFormat));
```
rowo360, line 45: That's great that you sorted that out straight away. I also noticed the flickering in the DisplayStrategyDialog and thought to create an issue for that in time. 👍

src/GuiRunner/TestModel/TreeConfiguration.cs
```diff
@@ -41,8 +41,11 @@ public string DisplayFormat
...
+                if (_displayFormat != value)
+                {
+                    _displayFormat = value;
+                    OnPropertyChanged(nameof(DisplayFormat));
+                }
             }
         }
```
rowo360, line 50: Could it be that the DisplayStrategyDialog is now displayed a little further down? However, I didn't spot any obvious code change which is responsible for that.

src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
CharliePoole: Yeah, I had hoped for more separation originally but I think Windows Forms makes that too difficult. Most alternatives we may use in a future release will also push us to use a different UI model anyway, so that's probably the point in time to try for a better separation.

src/GuiRunner/TestCentric.Gui/Presenters/TreeViewPresenter.cs
CharliePoole: As I commented elsewhere, we're stuck with the SWF Timer because of threading issues. But the flickering turned out to come from unnecessary updates to the TreeConfiguration.

src/GuiRunner/TestCentric.Gui/Presenters/TreeViewPresenter.cs
```diff
@@ -271,9 +275,25 @@ private void WireUpEvents()
...
 
+            // Some of our tests don't set the TreeView
+            if (_view.TreeView != null)
+            {
+                _view.TreeView.NodeMouseHover += (s, args) =>
+                {
+                    _mouseHoverDelayTimer.Stop();
+                    _mouseHoverDelayTimer.Interval = MOUSE_HOVER_DELAY;
+                    _mouseHoverDelayTimer.Tick += (s, e) => OnTreeNodeMouseHover(s, args);
+                    _mouseHoverDelayTimer.Start();
```
rowo360: Am I right in thinking that the problem still persists?
It might be that there's less flickering, but overall it results into calling `_tipWindow.Show()` multiple times. And unfortunately it only sums up getting worse when hoovering the mouse cursor over and over again.

Maybe we can create always  a new Timer object? Or register the Tick event only once?

src/GuiRunner/TestModel/TreeConfiguration.cs
CharliePoole: That happened outside of this PR. Commit 93f7fdb1 added positioning relative to the  TreeView's location in the constructor for the dialog.

rowo360 (changes requested):
So overall, things are looking good, but this issue with the numerous `_tipWindow.Show()` calls doesn't seem to have been fully resolved yet.

src/GuiRunner/TestCentric.Gui/Presenters/TreeViewPresenter.cs
CharliePoole: I think the problem here is that each node has to use a different timer because the node is specified in the constructor. I'll do a new commit that changes it so the constructor only specifies the TreeView and the node is a property we set and we don't construct a new one as you move from node to node.

CharliePoole:
I don't think that _tipWindow.Show is the problem. Rather it's the creation of multiple TipWindows. If we keep a single TipWindow and just Hide or Show it, that shouldn't result in flicker.

rowo360:
OK, another consideration:
We are currently using the NodeMouseHoover event to start the timer. Don't we need something like a MouseLeave as well? So, we'll stop the timer as soon, as the mouse move further?
This fits to my observation that a tooltip gets displayed although I moved my mouse cursor already horizontal to the property view. 
And I'm not sure about Tooltip behavior in general : shouldn't a tooltip only be displayed if the mouse cursor remains unchanged at one location? Or might it already get displayed, if I move the cursor on top of the control? For example if I move the mouse cursor within one single node?

CharliePoole:
Microsoft seems to define "Hover" in a much more complicated way. The mouse doesn't have to be stationary but must remain within some distance of it's initial position. Just as for ToolTips, the implementation of NodeMouseHover is hidden from us. We don't know the distance or the time.

I'm working on a new commit now, which creates only one TipWindow and hides and shows it as needed, while also changing it's position each time the node changes. If that's not sufficient, I'll drop use of NodeMouseHover entirely and go back to the old-school way of doing things... i.e. tracking all mouse movements within the TreeView and determining which node we are over.

BTW, Show and Hide only change visibility. They would only cause flickering if we did it very rapidly.

Anyway, this is enough for me right now... I'll sleep on it and come back in the morning.

CharliePoole:
@rowo360 The latest commit rewrites the logic for displaying the TipWindow completely. It's no longer dependent on TreeView's internal ToolTip implementaton but tracks mouse movements and translates them to individual TreeNodes before firing our own TreeNodeMouseHover event. The TreeWiew is now a derived class, TestCentricTreeView.

The merges failed several times until I finally merged in your changes locally and rebuilt it. Right now, it looks as if everything is working except for groups, which worked for me before the merge. I  guess the merge wiped out some changes. So, hopefully, one more commit will put it right. :-)

CharliePoole:
@rowo360 That's it. Let's see if we can get this merged so we are both on the same codebase. I'm in the mood for small, more localized fixes going forward. :-)

src/GuiRunner/TestCentric.Gui/Views/TestTreeView.cs
```diff
@@ -77,6 +90,28 @@ public TestTreeView()
...
+
+                    _mouseHoverDelayTimer.Stop();
+                    _mouseHoverDelayTimer.Interval = MouseHoverDelay;
+                    _mouseHoverDelayTimer.Tick += (s, e) =>
+                    {
+                        _mouseHoverDelayTimer.Stop();
+                        // Possible race conditions
+                        if (_lastNodeHovered != null && !TreeContextMenu.Visible)
+                            TreeNodeMouseHover?.Invoke(_lastNodeHovered);
+                    };
```
rowo360: I think we’re facing the same problem as before: we must not register to the Timer.Tick event within the MouseMove event. This must happen somehow outside - otherwise we are registering over and over again.
That fits to the observation that the application get inoperable when showing/hiding tooltips over some time.

rowo360 (changes requested):
Sorry, I've spotted something else again. Somehow it doesn't quite look finished yet.

src/GuiRunner/TestCentric.Gui/Views/TestTreeView.cs
CharliePoole: I see one thing to change here. I think we have to stop, reset and restart the timer here, but we should not register the handler. That should be a named method registered only once. Let's try again. Then, if it still causes problems, I need a recipe to repro the problem, which I'm not seeing.

CharliePoole:
That's OK. Main problem is that I'm not seeing what you are seeing, else I would have debugged it. I'll make another try and if that doesn't fix it, let's have a session where you show me on your screen.

CharliePoole:
@rowo360 I still have cleanup to do but look at the new commit and try it out if you have time.

rowo360:
I’d already thought that things must be different for you, as you’d surely have noticed otherwise. But here's some good news:  the issue is solved now 👍 No more flickering or delay 😄 

I'll try to add a video for one additional observation:

https://github.com/user-attachments/assets/802ea649-3f8a-42d5-89b0-1f1bdf8efc5d

At the end of the video I didn't move the mouse but the tooltip was continuously showing/hiding automatically.

rowo360:
And another video showing no urgent problem, but some rather unusual behaviour. Whenever the a tooltip is shown, the tree selection is switching to another color (don't know the exact term (inactive control?)). When I move the mouse over the tree, it creates a somewhat restless impression.

https://github.com/user-attachments/assets/b2930f2d-1643-4f68-a74c-032153373f24

rowo360:
Definitely not important either:
I cannot double-click a test case node anymore to execute this test, when the tooltip is shown.

CharliePoole:
@rowo360 

I think this is good enough to merge, allowing us to continue without major changes hanging. See if you agree.

First Video: This took me a while to replicate and understand. Each TreeNode has two different Bounds areas. One covers the entire row, all across the TreeView control. The other is just the text area. I'm turning on the TipWindow if the mouse is in the first, larger area. I'm turning it off when it exits the label, which covers the text area. I need to use the same bounds for both operations. I'll fix this. **FIXED**

Second Video: This is normal behavior when the TreeView loses focus but disconcerting when it loses it to the ToolTip, which is a separate Window. I'll try to take focus away from the TipWindow to fix this. I agree that this is unimportant, however.  **POSTPONED** Requires OwnerDraw to fix I'll create a separate issue.

Double-click... It seems to me that I should be able to forward the double-click to the TreeView. Again, I'll try but its not a priority. **POSTPONED** Existing special handling of windows messages is no longer working. I'll create a separate issue.

One other thing: My changes to TipWindow have broken it for use with ExpandingLabel. For now, I'll fix that in an adhoc way, i.e. by testing to see if the window is being used with a Label. The reorganization I wrote about can wait till after beta8. **ADHOC FIX** For this one, I'll create a long-running tipwindow branch and keep it up to date. Since this is more about cleaning up the architecture, I'd aim to finish it for the final release.

CharliePoole:
"Final" commit restores recursion when calling `TestTreeView.SetImageIndex` in `DisplayStrategy.cs`. This could conceivably break some of the Running icon changes, but I couldn't tell. However, without the recursion, group icons are not set at all for Outcome or Duration groupings.

rowo360:
Thank for the explanation about the different bounds. And indeed the issue of the 1st video is solved, now. But based on your input I tried a similar use case on the right side and unfortunately the issue pop up again.

https://github.com/user-attachments/assets/5e9810d6-8c2e-4fe4-8ba4-216d895862f4

Hmm, maybe you can check if this can easily be fixed as well? On the other hand, we should consider the best way to proceed. Perhaps we could go ahead with the merge and create a new issue for it.

Before that, we could also implement a fallback strategy. This would allow us to simply but effectively disable the tooltip feature completely. (A bool flag disabling all timers and so on.. - hopefully one single, central location). Just in case we encounter further problems or doubts remain.

BTW: I'll have a look to the running icons again after the merge and fix any remaining issues.

CharliePoole:
You're right... and the fix is obvious... I just didn't think about the right-hand side.

A simple flag we can only change in the build is trivial. I'll do that as well.

Can you take a look at the latest code as well - either now or after a new commit?

rowo360 (approved):
I think it looks good and we can go ahead with the merge. Of course, there’s now some specific code for handling tooltips, but the WinForms tooltip just wasn’t up to the job. 😄

CharliePoole:
Thanks for looking. I have a new build, which I'll commit and merge. It fixes the problem with moving off the tooltip to the right and also defines USE_TOOLTIP as a compile-time constant just as we did with DISABLE_MINI_GUI. That will help if we run into problems. My thinking is to get rid of both of them for the final release, probably keeping the TipWindow and dropping the MiniGUI, but it can stay as is for beta8.

I'll also create the additional issues I mentioned after the merge.

Issue #1500, opened 2026-04-07
Filter tree for outcome, category or text is not working


rowo360:
**Describe the bug**
Filtering the tree using the toolbar has no effect anymore.

CharliePoole:
@rowo360 I believe this will be resolved by PR #1513

rowo360:
As well as the work involved in reactivating filtering in the first place, I’d like to start a discussion about the TreeNode icons we’d like to see in a filtered tree. In particular when applying a filter **after** a test run. In this context, I think a discussion is only necessary for nodes of the type Assembly/Fixture/Namespace (i.e. the inner nodes). The icon for test case nodes remains as it is from the test run.

And basically I come up with two options:

- The icon for the nodes remain unchanged as it is from the test run
- The icon for the nodes is adapted to the filtered tree

When I apply a filter, I expect the TreeNode icons to be applied to the filtered data. This allows me to reliably identify, based on an icon in the hierarchy, what kind of test results I can expect in the child nodes and if it's worth to expand the nodes. In other words, for me, the second option is the better choice.
With Option 1, I would be surprised (or annoyed) if a parent node showed “Failed” and, after expanding all child nodes, I only found “Passed” tests. (because all 'failed' child nodes are filtered out).

In practice, this would mean that we cannot simply take the test result directly from the TestNode, but would instead have to ‘calculate’ the result (take most severe) from the TreeNode children. Basically that's similar for the running icon approach.

CharliePoole:
Logically, what you say makes sense. I'm looking at it now and it's a bit confusing. However...

If you fix this as you suggest, a user not terribly familiar with filtering (like me) might get a bit confused as well. Can we do something to make the selection of outcome filters more visible in the toolstrip? Perhaps a greyed or dimmed version of the normal icon?

PR #1518, opened 2026-04-18
Enable filtering toolbar

rowo360:
This PR contributes to #1500, but doesn't completely solve it.

It takes care that the filter toolbar buttons have again any effect on the displayed tree nodes. So, the tree nodes are successfully filtered according to the filter criteria. That's working fine for the NUnit and the test list display strategy.

What is missing and still needs to be restored:

- The tree node icons are not updated to the filtering operation, but still display the unfiltered state
- A test run doesn't consider the filtering. But also filtered out nodes are included in the test run. That applies for 'Run all' but also for a selected test run.

Although the filtering will not completely work right away, I thought it was a good idea to start with a small commit. Especially since we'll discuss about tree node icon state soon and the filtering will add an additional dimension. So, we can try out any effects in practice.

### DIFF
--- a/src/GuiRunner/TestCentric.Gui.Tests/Presenters/TestListDisplayStrategyTests.cs
+++ b/src/GuiRunner/TestCentric.Gui.Tests/Presenters/TestListDisplayStrategyTests.cs
@@ -46,17 +46,19 @@ namespace TestCentric.Gui.Presenters
             _treeNodes = null;
         }
 
-        [Test]
-        public void OnStrategyCreated_OutcomeFilter_IsInvisible()
+        [TestCase(true)]
+        [TestCase(false)]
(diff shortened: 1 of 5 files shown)


rowo360:
The first commit enables the filtering. Whenever the filtering toolbar is changed, the `IsVisible `property of all TestNodes is updated accordingly to the filter criteria. The NUnitDisplayStrategy and TestListStrategy just miss to evaluate this property. This check is added now, so whenever the IsVisible property is false, the strategy will skip the node.

rowo360:
The second commit is about the filtering toolbar in the Test list view. There was still a leftover, when the filtering was completely disabled in the test list view. I removed that part, now.

CharliePoole (approved):
This looks good to me.

Issue #1515, opened 2026-04-17
Create variants of Running icon, showing result so far


CharliePoole:
We need four variants of the existing Running icon, using the small dot design shown in #1508...

* Passed (green) 
* Failure (red)
* Warning (orange)
* Ignored (yellow)

If it's possible in the small size, the dots could contain a check mark, X, ! and ?, respectively. If not that's OK.

The variants should be stored in the same directory as the basic Running and Pending icons.

rowo360:
I tried using question marks and exclamation points. But in the Tree View, you can't recognize them at all anymore. In fact, you don't even notice the yellow color anymore.

Here are the experiments:
<img width="262" height="227" alt="Image" src="https://github.com/user-attachments/assets/81304e57-15db-4647-86f4-81de9fc26f3e" />
<img width="262" height="227" alt="Image" src="https://github.com/user-attachments/assets/e8d8d393-4c80-443e-b73b-485c2ccd312c" />

That's why I'll stick to simple circles.

CharliePoole:
Well, just the dots then...

CharliePoole:
You'll need to assign indices to them and update the list of images at the start of `TestTreeView.cs`.  The code in `TestTreeView.SetImageIndex` is where the parents are currently set to `Failed`, `Passed` etc. That needs to instead use the corresponding `Running` version of the icons. There may be some other places as well.

Once that's functioning and merged, I'll pull it into my working code for #1510.

PR #1519, opened 2026-04-18
Add new tree node icons for running state

rowo360:
This PR adds new icons for tree nodes to indicate different intermediate result states during a test run (Contribute to #1515)
Running_Success
Running_Warning
Running_Failure
Running_Ignored

I tried to use the question / exclamation mark but the result was too tiny, so I stick with the circle.

First, I wonder that we need also an icon for the Inconclusive result state. But noticed soon, that this state is not propagated to the parent nodes. Instead parent nodes report success even if they contain an inconclusive child node. So, it makes senese, that we don't need a running icon for inconclusive state.


rowo360:
There isn't much to review here 😄 
But I'll stick with the process and request a review - so, at least, you are informed that this work is done, now.

CharliePoole (approved):
This looks good so far. They won't be loaded or used because no indices are assigned and the view and presenter don't know that these icons exist. 

If that was your intent, then this is good to merge, otherwise, we can wait for another commit with the code itself.

rowo360:
Yes, of course that's true! 😄 
I thought it was just a matter of creating the icons at first. But without a use for them, it’s not quite finished yet. But whether we do this in this issue or create a new one, either is fine with me. 

I'll merge this PR now, but not close the issue right away. I thought we want to discuss how the icon/result state of a group is handled before starting this part. But I'm also going to take a look at how they icons could be used next and come up with a proposal.

CharliePoole:
Great! I'll comment further on the issue.

PR #1520, opened 2026-04-19
Use different running image indices depending on the test results of the child nodes

### LINKED ISSUE #1515: Create variants of Running icon, showing result so far

rowo360:
I decided to create a PR for #1515 already now, although there's one flaw which must be fixed by another commit. But before I take on too much at once, I’d like to pause for a moment and discuss with you to see if everything’s okay so far.

I extended the TestTreeView ImageIndex numbers and inserted the new values next to existing RunningIndex number. Which means these are low numbers, which is important when discussing about the flaw.

Whenever a a test is starting (e.g. `DisplayStrategy.OnTestStarting`) we determine the exact type of running index by iterating through all child nodes and getting the most severe running image index. This is done recursive to handle the entire parent hierarchy - each parent node is also in running state now, but the exact type of running index might be different.

Actually, this code part works as intended. However, at the end of the test run there are some incorrect image indices applied to the nodes. This only happens in the TestList view; the NUnit view works correctly.
Here's a screenshot from the flaw:

Although there is a failed test case, the image index of the parent nodes are set to success! This happens whether the test list view is grouped or ungrouped.

The root cause of this issue is the method `DisplayStrategy.OnTestFinished()`. It takes the test result of the finished test and applies it to all parent nodes recursively. However, an ImageIndex is only applied if the new value is larger than the existing one. But unfortunately, that's the case now: the image index of the parent nodes is set to Running_Failure, which is a low number. When the last test case finished with 'success', the success image index is larger than Running_Failure and therefore 'success' is applied to all parents.

Therefore I think we should no longer apply the test result recursively to all parent nodes in method `DisplayStrategy.OnTestFinished()`. That can be easily be switched off, but the result is still not correct. Now, the running image indices of the parent nodes are never updated and after a test run the tree would look like this:

Here is my proposed solution:
There's a OnTestFinished event also for the fixture and test assembly node. In the NUnit view this event ensures that the ImageIndex is correctly updated for the fixture/assembly node. And that's exactly how I'd like to handle it in the TestList view as well. Right now, the test list view isn't working because the `_nodeIndex` mapping <TestId, <List<TreeNode>> is incomplete. For all treeNodes created for TestGroups the `_nodeIndex `mapping isn't updated. (see method MakeTreeNode(TestNode) vs MakeTreeNode(TestGroup)). Which basically means, that in the TestList view the OnTestFinished does nothing.
So overall, I want to populate the _nodeIndex mapping for TestGroups as well, so that the OnTestFinished event can detect the relevant TreeNodes. Each TestGroup is representing a certain TestNode which can be used for the mapping. There will be no mapping for the root TestGroups (category, ...)

But before I get started on the implementation, let’s discuss this so that everything goes into the right direction.

### DIFF
--- a/src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
+++ b/src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
@@ -129,7 +129,7 @@ namespace TestCentric.Gui.Presenters
                 {
                     treeNode.Text = GetTreeNodeDisplayName(result);
                     treeNode.ToolTipText = result.Label != null ? result.Label : result.Status.ToString();
-                    _view.SetImageIndex(treeNode, imageIndex, true);
+                    _view.SetImageIndex(treeNode, imageIndex, false);
                 }
             });
         }
@@ -451,11 +451,28 @@ namespace TestCentric.Gui.Presenters
 
         protected void ResetTestRunningIcons(TreeNodeCollection treeNodes)
         {
-            // Only required for exceptional use case 'force stop test run'
+            // Update all running icons to their corresponding non-running icons (required for TestList view)
             foreach (TreeNode treeNode in treeNodes)
             {
-                if (treeNode.ImageIndex == TestTreeView.PendingIndex || treeNode.ImageIndex == TestTreeView.RunningIndex)
-                    _view.SetImageIndex(treeNode, TestTreeView.SkippedIndex);
+                switch (treeNode.ImageIndex)
+                {
+                    case TestTreeView.RunningIndex_Success:
+                        _view.SetImageIndex(treeNode, TestTreeView.SuccessIndex);
+                        break;
+                    case TestTreeView.RunningIndex_Failure:
+                        _view.SetImageIndex(treeNode, TestTreeView.FailureIndex);
+                        break;
+                    case TestTreeView.RunningIndex_Warning:
+                        _view.SetImageIndex(treeNode, TestTreeView.WarningIndex);
+                        break;
+                    case TestTreeView.RunningIndex_Ignored:
+                        _view.SetImageIndex(treeNode, TestTreeView.IgnoredIndex);
+                        break;
+                    case TestTreeView.PendingIndex:
+                    case TestTreeView.RunningIndex:
+                        _view.SetImageIndex(treeNode, TestTreeView.SkippedIndex);
+                        break;
+                }
 
                 ResetTestRunningIcons(treeNode.Nodes);
             }
--- a/src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
+++ b/src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
@@ -10,6 +10,7 @@ using System.Windows.Forms;
 
 namespace TestCentric.Gui.Presenters
 {
+    using System;
     using Model;
     using Model.Settings;
     using Views;
@@ -116,7 +117,7 @@ namespace TestCentric.Gui.Presenters
         public virtual void OnTestStarting(TestNode testNode)
         {
             foreach (TreeNode treeNode in GetTreeNodesForTest(testNode))
-                _view.SetImageIndex(treeNode, TestTreeView.RunningIndex, true);
+                SetRunningImageIndex(treeNode);
         }
 
         public virtual void OnTestFinished(ResultNode result)
@@ -384,6 +385,49 @@ namespace TestCentric.Gui.Presenters
                 }
 
                 _view.SetImageIndex(treeNode, imageIndex);
+            }
+        }
+
+        private void SetRunningImageIndex(TreeNode treeNode)
+        {
+            // Test execution is starting for the tree node, so set the image index to "Running" 
+            // The exact type of the “running” image index is determined by the strictest image index of all child nodes
+            int imageIndex = TestTreeView.RunningIndex;
+            foreach (TreeNode childNode in treeNode.Nodes)
+                imageIndex = Math.Max(imageIndex, GetRunningImageIndex(childNode));
+
+            _view.SetImageIndex(treeNode, imageIndex, false);
+
+            // Determine image index for parent nodes as well
+            if (treeNode.Parent != null)
+                SetRunningImageIndex(treeNode.Parent);
+        }
+
+        /// <summary>
+        /// Map the image index of a tree node to the corresponding "running" image index
+        /// </summary>
+        private int GetRunningImageIndex(TreeNode treeNode)
+        {
+            switch (treeNode.ImageIndex)
+            {
+                case TestTreeView.SuccessIndex:
+                case TestTreeView.SuccessIndex_NotLatestRun:
+                case TestTreeView.RunningIndex_Success:
+                    return TestTreeView.RunningIndex_Success;
+                case TestTreeView.FailureIndex:
+                case TestTreeView.FailureIndex_NotLatestRun:
+                case TestTreeView.RunningIndex_Failure:
+                    return TestTreeView.RunningIndex_Failure;
+                case TestTreeView.WarningIndex:
+                case TestTreeView.WarningIndex_NotLatestRun:
+                case TestTreeView.RunningIndex_Warning:
+                    return TestTreeView.RunningIndex_Warning;
+                case TestTreeView.IgnoredIndex:
+                case TestTreeView.IgnoredIndex_NotLatestRun:
+                case TestTreeView.RunningIndex_Ignored:
+                    return TestTreeView.RunningIndex_Ignored;
+                default:
+                    return TestTreeView.RunningIndex;
             }
         }
 
--- a/src/GuiRunner/TestCentric.Gui.Tests/Presenters/TestTree/WhenTestCaseCompletes.cs
+++ b/src/GuiRunner/TestCentric.Gui.Tests/Presenters/TestTree/WhenTestCaseCompletes.cs
@@ -35,7 +35,7 @@ namespace TestCentric.Gui.Presenters.TestTree
             _view.InvokeIfRequired(Arg.Do<MethodInvoker>(x => x.Invoke()));
         }
 
-        [TestCaseSource("resultData")]
+        [TestCaseSource(nameof(resultData))]
         public void TreeShowsProperResult(ResultState resultState, int expectedIndex)
         {
             var result = resultState.Status.ToString();
@@ -62,7 +62,7 @@ namespace TestCentric.Gui.Presenters.TestTree
             _model.Events.TestLoaded += Raise.Event<TestNodeEventHandler>(new TestNodeEventArgs(testNode));
             _model.Events.TestFinished += Raise.Event<TestResultEventHandler>(new TestResultEventArgs(resultNode));
 
-            _view.Received().SetImageIndex(Arg.Compat.Any<TreeNode>(), expectedIndex, true);
+            _view.Received().SetImageIndex(Arg.Compat.Any<TreeNode>(), expectedIndex, false);
         }
     }
 }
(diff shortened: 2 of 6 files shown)


CharliePoole:
The second of your two images at least carries the correct information... i.e. I can tell that the overall result is a failure because of the red dot. So maybe there is a solution there.

The problem with updating the _nodeIndex mapping to include groups via an "associated TestNode" is that the same assembly or fixture subgroup may appear under multiple top-level nodes. The associated node would only give you the overall result for all tests in that assembly or fixture.

Taken in isolation (i.e. in the context of this issue by itself) the easiest solution may be to update all the nodes at end of run, converting any of the Running indices to a final index. An alternative would be to put a countdown field in `TestGroup`, intializing it to the number of cases when the group starts, decrementing it as each test finishes and converting to the proper final index when it hits zero.

FYI, my work on  #1510 may overlap some of this and possibly change it further.

src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
```diff
@@ -116,7 +117,7 @@ public void OnTestUnloaded()
         public virtual void OnTestStarting(TestNode testNode)
         {
             foreach (TreeNode treeNode in GetTreeNodesForTest(testNode))
-                _view.SetImageIndex(treeNode, TestTreeView.RunningIndex, true);
+                SetRunningImageIndex(treeNode);
```
CharliePoole, line 120: Not clear why you would do this when the test is starting... should it not always be just "Running"  when it starts, with children set to Pending?

CharliePoole (commented):
I don't understand something here. See inline comment.

rowo360:
Yes, that's right:
Even if I had the mapping complete, I cannot use the test result from the Fixture/Assembly node directly. I had in mind to determine the test result by iterating through all child nodes and take the most severe child test result. That would fit nicely to the filtering feature. But actually, a separate discussion was intended for these questions, and it wasn’t the case that everything had to be decided once and for all in this issue. Especially if you are already on overlapping tasks in parallel.

That’s why I think your suggestion is a good one: we update all running indices at the end of the test run. That might not the final beta 8 solution, but it's sufficient to complete this issue. 
Afterwards we can create a new issue and discuss and agree upon, if we want to improve.

src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
rowo360: All tests included in the test run are set to Pending ImageIndex in the `OnTestRunStarting` event. Afterwards the `OnTestStarting` event is invoked for each individual node. This code change is located in method `OnTestStarting` which indicates that the test execution for this node really started, so the ImageIndex is set to running. Which exact type of running (failed, success...) is determined by navigating through all child nodes. In addition the running ImageIndex is determined for all parent nodes as well. So, the test progress is even evident, if the hierarchy is collapsed.

CharliePoole:
Hey, good morning! :-)

Here's a bit of code from the PR I just finished, waiting for your review. See if this approach might help as it updates the image indexes without waiting for the end of run...

```
        private void ApplyResultsToTree(TreeNode treeNode)
        {
            TestNode testNode = treeNode.Tag as TestNode;

            if (testNode != null)
            {
                if (resultNode != null)

                // NOTE: the loop only executes at the point in the hierarchy where the current
                // TreeNode represents a TestNode and the parent is a TestGroup;
                for (var parent = treeNode.Parent; parent?.Tag is TestGroup; parent = parent.Parent)
                    parent.ImageIndex = parent.SelectedImageIndex = Math.Max(parent.ImageIndex, treeNode.ImageIndex);
            }

            foreach (TreeNode childNode in treeNode.Nodes)
                ApplyResultsToTree(childNode);
        }
```

rowo360:
Here's another commit that implements your idea of updating all running icons to the corresponding non-running icons at the end of the test run. This idea fits well to the existing code which already takes care about runing icons at the end of the test run. 

This PR fixes #1515, however overall it's not quite perfect yet. 😄 
The image index of fixture nodes and assembly nodes could be updated much earlier and we don't need to wait until the end of the test run. But all of this can be discussed in a separate issue.

rowo360:
@CharliePoole: I removed the draft status and may I kindly ask for a review of the 2nd commit? Thanks!

src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
CharliePoole: OK... I get it with respect to Pending, but not Running.

Let's say that testNode refers to a fixture. Each treeNode relates to the same fixture. For NUnitTree, there is just one treeNode, for TestList there may be many.

In either case, given that the fixture is just starting, it can only get RunningIndex, without any colored dot. It's only when the first contained test case completes that it can be updated to one of the variants. So at this point (TestStarting) only Running is possible.

Or have I missed something?

CharliePoole (commented):
Sorry, but I'm still not getting what you're doing in OnTestStarting.

rowo360:
Yes, at first I also thought that on TestStarting event the corresponding TreeNodes can only be set to RunningIndex. That's true, if you execute all tests or all tests of a fixture. But if you execute only partially you have to take into account the test results from a previous test run. 
So, the statement is still correct for nodes representing a test case. If such a node gets the TestStarting event, the icon can and will only be set to RunningIndex. For fixture/assembly nodes, there might be already results from previous runs which should be considered for an accurate running index. At least I thought it was a good idea. And don't showing some icons during the run which implies that it will be successful, only to end up showing that there is an error.

Some examples:

- A fixture with two test cases is executed:

1. Event OnTestRunStarting: all nodes in test run are set to PendingIndex
2. Event OnStarting for fixture node: all children have PendingIndex => Fixture get RunningIndex
3. Event OnStarting for 1st test case: RunningIndex for test case node;  No change for fixture node
4. Event OnStarting for 2nd test case: RunningIndex for 2nd test case node; (assuming 1st test case fails); Fixture node is changed to RunningIndex_Failure 
5. Event OnTestRunFinished: Fixture node is changed to Failure

- A fixture with two test cases was executed, 1st fails, 2nd succeed. Now, only 2nd test case is executed again.
1. Event OnTestRunStarting: Fixture node and 2nd test case are set to PendingIndex; 1st test case is set to Failure_NotLatestRun
2. Event OnStarting for fixture node: one child has PendingIndex, other child has Failure_NotLatestRun => Fixture gets RunningIndex_Failure
3. Event OnStarting for 2nd test case: RunningIndex for test case; No change for fixture node, because one child still failed (from previous run)
4. Event OnTestRunFinished: Fixture node is changed to Failure

CharliePoole:
OK... I' glad I asked because it turns out we have different expectations.

Yours... that a fixture containing some not run tests that failed would reflect failure even if all the tests that were executed passed.

Mine... that the same fixture would reflect success, even though it contained some tests that had previously failed, which were not run this time.

I thought my expectations were "obvious" and probably you thought the same about yours. :-)

Since showing past results is a new thing for us, we should talk about the pros and cons of each approach and also consider what other runners do...

My take is that either way is going to be a bit surprising to some people. That is...
* Why is the fixture showing that it failed even when all my tests ran? OR
* Why is the fixture showing that it passed even though some test cases show as failed?

Your thoughts?

rowo360:
It's good to get another point of view to this issue. This helps to check whether you're heading in the right direction.

So, it's all about test results from previous test runs and how to integrate them into the running icon and/or the final icon. And as you said, I’ve been convinced of my approach so far. But I can also see the merit in the other approach.

So, from my point of view I wanted to maintain the rule that a parent node always aggregates the result state from its children. So if a node is collapsed but has a success state, everybody can rely on that all containing tests did pass. A different behavior would be a bit surprising for me.
Moreover I didn't want to restrict these new running states to consider only those nodes from the current test run. The progress bar displays the result state (red,yellow,green) of the current test run (without taking care about previous test runs). So, the user can watch this as an indication. Moreover he can always 'clear all test results' in the tree to get a fresh view.

Here are some screenshots - first from the PR:
TestC is executed and will pass. However, during the test run the parent nodes 'Feature2Tests' and 'MyLibrary.Test.dll' have state running_failed, because TestB failed in previous run. At the end of the test run the failed icon is displayed for fixture and assembly node.


As you mentioned other test runners, I looked to VS and R#.
Well, VS doesn't have different running states, but displays this clock icon. But the final result looks identical to TestCentric.



I wasn't aware, but R# has indeed some different running icons. They are not static icons, but are even spinning, but overall the use also some color coding to integrate the test results from previous runs.



First, the clock icon is displayed, next R# switch to the running icons. Please note, that TestC has a running success icon (hmm, that test is not finished yet, but this seems to be their logic), the fixture and assembly node has a running failure icon. So, they are integrating test results from previous runs also into their running icons. So, that's the identical behavior as in the PR .

CharliePoole:
Taking just the failed case, let's imagine we needed to tell a user what the icon means.

Using my assumption, we could say "It's still running but has some failures and will fail in the end." Using yours, we would need to give a more complicated explanation like "It's still running but has some failures. If the failures are from the current run, it will fail in the end, but if they are all from prior runs, the test could still pass"

One rule of thumb I've used is that something that is complicated to explain is a "smell." That is, it may be OK, but the smell is suspicious. :-) My own feeling is that this introduces unnecessary complexity on several levels...
* For the user, who has to understand it
* In the implementation, which we have to maintain.

So I'd prefer a simpler approach where prior results don't "count" for running higher-level tests. 

That said, if you really love this feature and are willing to deal with some future complexity, we can leave it in.

Regarding the progress bar... now that you mention it, I find it to be an extra confusion if the progress bar doesn't match the top level result icon. :-)

rowo360:
I'm still in favor of the current proposal and doesn't consider the explanation that complicated: "It's still runing, but has some failures (from current run or previous run) and will fail in the end". For all tests run so far, the result for the node has been an error, and that is not going to change. (It might only change, if the test results are deleted and a different set of tests is executed). I'm only talking about assembly/fixture/namespace nodes which have some child nodes. Test case nodes only show the 'neutral' running icon and switch to the final result icon as soon as the test is finished (as you can see in the screenshot from previous post).

I also find it quite reassuring that we’re acting just like the other test runners. This should make it feel familiar to everyone, and we’re not trying to go down a different path. And try to convince the user that we've found a better approach.

Still, what you say is true, of course: “Anything that can only be explained in several sentences is too complicated.” But my feeling is that the other approach will also end up being complicated and hard to understand. Starting a 3rd or 4th test run would require some special logic to update the icons. Currently a passed or failed test is simply mapped to the passed_NotLatest or Failed_NotLatest icon at TestRunStarting event. Wouldn't this end up in a total mess, when keeping this implementation? For example when starting the 3rd run for a complete different fixture?

We'll talk about the icons again soon when we get to the topic of filtering. So maybe we can proceed for now and check if things get complicated there again. In that case we rollback and search for a different solution. OK?

CharliePoole:
OK... to sum up...

You already convinced me that we need to have an indicator of what's running and that higher-level tests (Suites) should show as still running until all results are done.

You also convinced me that the higher level tests that have started to run should show a type of running icon that indicates whether they are likely to fail, pass, etc.

We only disagree on whether previously run child tests should contribute to the higher level prediction. This has some importance, but it's more like a sub-sub-feature than a feature. :-) I dislike it because it could appear to predict (for example) that a fixture will fail but that it will still up passing.

Still, it's (relatively speaking) a minor thing and if you feel it's valuable let's keep it isolated so we can rollback later if necessary.

CharliePoole (approved):
Based on our separate discussion, I'm OK with this one. Thank you.

PR #1522, opened 2026-04-25
Fix checking a TreeNode for a TestGroup

### LINKED ISSUE #1521: Show checkboxes: Toolbar button 'Run selected Tests' remains disabled in list view
**Describe the bug**
When using **checkboxes** in the list tree view, the toolbar item 'Run selected tests' keeps disabled and I cannot run all the checked tests.
This happens only for assembly, fixture or grouping nodes, for test case nodes it's working well.

I assume it's something related to TestGroup vs TestNode.

rowo360:
We recently changed that a `TestGroup` is no longer a TestSelection but aggregates one. We missed to adapt this code part at that point in time.

In case the treeNode.Tag is a TestGroup we need to request the TestSelection from the TestGroup first to Add() all items to the new TestSelection.

Fixes #1521


CharliePoole (approved):
Good bug! Good Fix! 

I'm wondering... Now that we have subclassed TreeView, maybe  we should subclass TreeNode as well and stop messing around with Tags!

PR #1524, opened 2026-04-27
Restore ShowCheckBoxes from VisualState file

### LINKED ISSUE #1523: Show Checkbox is not properly restored from VisualState
**Describe the bug**
If I enable the option to show checkboxes in the tree, this option will be restored when I reload the project/assembly the next time. Indeed, the tree is displayed showing checkboxes - that's fine.

But the context menu is convinced that there are no checkboxes (the menu item is unchecked). I need to check it twice to disable the checkboxes again.
Overall, it seems that some elements are not correctly initialized with the values from the VisualState file.

rowo360:
I tried to explain the root cause already in the issue itself. The WinForm method `toolStripItem.GetCurrentParent()` returns null for a context menu ToolStripItem as long as the context menu was never opened. As a result the `InvokeIfRequired` method fails and won't apply the new value. As soon as the context menu was opened once the `toolStripItem.GetCurrentParent()` returns a valid parent, but that happens without us doing anything and so we can’t control it ourselves either.

One option might be to rollback that mentioned commit, but I assume it was for a good reason. Therefore I propose to handle the checked state of the context menu items differently as before. So far, we tried to update the checked state right away when the underlying value in the TreeConfiguration changed. Now, I propose that we update the checked state for the context menu items only when the context menu is openend.

The first commit extend the existing method InitializeContextMenu(), In addition to all previous task, the checked state is now updated as well.

The second commit is an optimization: 
I want to avoid that change events gets already invoked causing a cascade of updates only because the context menu is initialized.   There might be different options to achieve this, but I simply deregister the events before initialization and reregister again afterwards.

Fixes #1523.


CharliePoole:
@rowo360 can we hold this until the TipWindow stuff is merged? I think conflicts will arise otherwise.

rowo360:
Don't worry - I’d also thought you’d finish the tooltip topic first 😄

CharliePoole:
I'm seeing the same problem with the CI. It fails to find the GUI we just built on myget.org. Of course, we haven't put it there yet and only do so after everything passes, so I think something is wrong  with the script. I'll look at it in the morning.

PR #1529, opened 2026-04-29
Movability of TestCentricProject files

### LINKED ISSUE #1474: Movability of TestCentricProject files
When a TestCentricProject file is moved to a new location, we would like it to continue to function correctly. There are two possible cases...

1. The user moves both the project file and the referenced test files at the same time, keeping all files in the same positions relative to one another.
2. The user moves only the project file.

NOTE: A combination of 1 and 2 is possible, if the user moves only some of the test files.

Of course, we need to decide which scenarios we will support, but I think we can support both cases.

One approach could be...

1. Newly saved projects are stored with the absolute ProjectPath to which they are saved.
2. Test files are stored with a path relative to the ProjectPath.
3. When opening an existing project, we check the ProjectPath to see if it still matches the location from which it was opened.
4. If the location has changed, we check each test file in the old location (based on the ProjectPath) or the new location from which the file was opened. Files remaining at the old location have their relative path changed so they continue to work.
5. The ProjectPath itself is finally changed to the new location.

We might also consider a warning to indicate that test files are still referenced at their old location.

@rowo360 Your thoughts on this?

CharliePoole:
Fixes #1474

### DIFF
--- a/src/GuiRunner/TestModel/TestCentricProject.cs
+++ b/src/GuiRunner/TestModel/TestCentricProject.cs
@@ -18,6 +18,7 @@ namespace TestCentric.Gui.Model
         public static bool IsProjectFile(string path) => Path.GetExtension(path).ToLower() == ".tcproj";
 
         public string ProjectPath { get; internal set; }
+        internal string OriginalProjectPath { get; set; }
 
         public TestPackage TopLevelPackage { get; private set; }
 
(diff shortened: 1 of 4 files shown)


CharliePoole:
Problems with MyGet connectivity today. Push job has failed four times even though the PR is passing.

src/GuiRunner/TestModel/TestCentricProject.cs
```diff
@@ -126,13 +127,18 @@ private void Load(string path)
...
+                OriginalProjectPath = xmlReader.GetAttribute("OriginalPath");
+
                 while (xmlReader.Read())
                 {
                     switch (xmlReader.NodeType)
                     {
+                        case XmlNodeType.Attribute:
+                            if (xmlReader.Name == "OriginalPath")
+                                OriginalProjectPath = xmlReader.Value;
+                            else
```
rowo360: Did I miss something or is the `OriginalProjectPath` already read in line 130?

rowo360 (approved):
Looks good!

src/GuiRunner/TestModel/TestCentricProject.cs
CharliePoole: You're right. I'll remove it here.

PR #1530, opened 2026-04-30
Eliminate selected node color change with TipWindow

### LINKED ISSUE #1525: Selected Node changes color when tooltips appear
If a `TipWindow` is shown, the selected node in the tree changes from blue to a light grey color. As `TipWindows` appear, close and reappear, the changes in color are a bit disconcerting, producing a somewhat restless impression.

This is caused by loss of focus to the `TipWindow` and is normal behavior when a window loses focus. To provide a better visual cue, the `TreeView` control provides a property, `HIdeSelection`, which defaults to `true`. By setting it to false, the selected node is highlighted even when focus is lost. We use this in our code, setting `HideSelection` to true. Unfortunately, the color shown for a selected node without focus is different from when it is focused, which is the source of this UI bug.

To get the same color in both the focused and unfocused states, we will have to draw the text portion of the nodes ourselves. We can either override `OnDrawNode` in our `TestCentricTreeView` class or use owner draw and handle it in an event. The choice will depend on which one gives us less work to do - i.e. we don't want to draw the graphics as well as the text.

CharliePoole:
Fixes #1525

### DIFF
--- a/src/GuiRunner/TestCentric.Gui/Controls/TestCentricTreeView.cs
+++ b/src/GuiRunner/TestCentric.Gui/Controls/TestCentricTreeView.cs
@@ -3,7 +3,7 @@
 // Licensed under the MIT License. See LICENSE file in root directory.
 // ***********************************************************************
 
-using System;
+using System.Drawing;
 using System.Windows.Forms;
 
 namespace TestCentric.Gui.Controls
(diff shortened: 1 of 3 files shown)


CharliePoole:
@rowo360 I selected a less obtrusive coloring for the selected node so as to be less busy visually. We can easily change the background and text colors, of course. Note that selected item highlighting is now the same in both focused and unfocused states independent of whether the TipWindow is displayed.

rowo360 (approved):
Ok- fine! I notice that we enabled the OwnerDrawMode for the TreeView. But only for the text part which is much simpler.

PR #1542, opened 2026-05-13
Update TestSelection of a all TestGroups when regrouping a test node during a test run

### LINKED ISSUE #1535: Outcome/Duration grouping: TestGroups are not completely updated
**Describe the bug**
When I work with the outcome/duration grouping, so that TestNodes are moved from the "Not run" node to the "Failed", "Passed"... node, the TestGroups are not updated accurately.

**To Reproduce**
1. Open mock-assembly.dll
2. Switch to outcome grouping and show Assembly + Fixtures
3. Run one single test:  MockTestFixture.FailingTest
4. It should look like this afterwards:

5. Execute test for node "Failed" or assembly node => no tests are executed as the TestGroup contains no nodes

And just as bad: the TestNode is still in the ‘Not run’ group and also in the TestGroup  of its assembly node.

**Expected behavior**
When regrouping test nodes all TestGroups in the hierarchy must be updated.

rowo360:
I'm back with this small PR which fixes #1535. I'm curious to know whether the pipeline is up again and running successfully. 😄 

The fix itself takes care that a TestNode is added to all new TestGroups in the hierarchy whenever regrouping is required. And also that it's removed from all old TestGroups in the hierarchy. Previously, the TestNode was only added+removed from one single TestGroup (the direct parent), all other TestGroups in the hierarchy remains unchanged. This works fine if test assembly and fixtures nodes are not shown, because the hierarchy consists only of one level.

### DIFF
--- a/src/GuiRunner/TestCentric.Gui/Presenters/TestListDisplayStrategy.cs
+++ b/src/GuiRunner/TestCentric.Gui/Presenters/TestListDisplayStrategy.cs
@@ -158,17 +158,16 @@ namespace TestCentric.Gui.Presenters
                         newGroup = newGroup.GetOrAddSubGroup(fixtureName);
                 }
                 else // ShowFixtures only
-                {
                     newGroup = topLevelGroup.GetOrAddSubGroup(fixtureName);
-                }
             }
 
             // If the group didn't change, we can get out of here
             if (oldGroup == newGroup)
                 return;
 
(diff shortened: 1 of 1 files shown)


CharliePoole (approved):
This looks good  to me. Thanks for catching this.

CharliePoole:
I'll send you a separate update about the pipeline.

PR #1546, opened 2026-05-22
Show number of test cases in tree node name

### LINKED ISSUE #1545: Tree node names display number of test cases inconsistently
**Describe the bug**
There's a 'feature' from the past which shows the number of containing test cases as a suffix in the tree node name. It was indented to work on all kind of **inner** nodes (assembly, fixture, namespace, category...) so the user has some insights about the amount of tests even when tree nodes are collapsed. Obviously tor a tree node representing a single test case there's no additional suffix displayed. Here's a screenshot:

The screenshot already displays the current issue: this feature isn't working completely anymore. In the screenshot the suffix is missing for the assembly node and the 'Passed' node.

**Expected behavior**
Maybe we should decide first if we want to keep this feature. If not, we should remove the responsible code parts. If yes, we should enable it to work in all use cases and all kind of tree nodes. And if we want to keep it, should we also enable it (again) in the NUnit tree?

rowo360:
Work has continued to keep me very busy, so I haven't managed to carry on working in time. It wasn’t actually that complicated, as there were already methods available for updating a TreeNode or a tree of TreeNodes.

For the TestList view we need to update the tree node names:

- Once after the entire tree has been build up. So, that the number of tests within all TestGroups is finalized and we use the correct numbers.
- On TestFinished event and some regrouping has taken place. All tree nodes which previously contained the test case must be updated (count has decreased) and all tree nodes which newly got the test cases must be updated as well (count has increased).

For the NUnit view the number won't change during a run and can be applied when building up the tree. However, I cannot use the `TestNode.TestCount` property because some parts of the tree might be filtered out (for example outcome filter). So, I added a method to determine the number of test cases. One idea might be to move this method to the TestNode class or even replace the existing `TestCount` property.

It’s already working well, but I wasn’t entirely satisfied with the result. In particular, with generic fixtures, the tree node name contains two items in brackets - I think that makes the meaning unclear. Here's a screenshot.

That is why I suggest we add a brief explanation (‘test or tests’). Here's also a screenshot:

(Fixes #1545)

### DIFF
--- a/src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
+++ b/src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
@@ -218,7 +218,7 @@ namespace TestCentric.Gui.Presenters
 
         private string GetTreeNodeNameCountSuffix(int count)
         {
-            return $" ({count} " + (count > 1 ? "tests)" : "test)");
+            return $" ({count} " + (count == 1 ? "test)" : "tests)");
         }
 
         private string GetTreeNodeDisplayName(TestNode testNode)
@@ -237,8 +237,12 @@ namespace TestCentric.Gui.Presenters
 
         private static int GetTestCount(TestNode testNode)
         {
-            // Check if the node is a test case, and it's not filtered out (e.g. by outcome filter)
-            if (!testNode.IsSuite && testNode.IsVisible)
+            // If TestNode is filtered out (e.g. by outcome filter)
+            if (!testNode.IsVisible)
+                return 0;
+
+            // Check if the node is a test case
+            if (!testNode.IsSuite)
                 return 1;
 
             // Iterate through all child nodes
--- a/src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
+++ b/src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
@@ -128,7 +128,7 @@ namespace TestCentric.Gui.Presenters
 
                 foreach (TreeNode treeNode in GetTreeNodesForTest(result))
                 {
-                    treeNode.Text = GetTreeNodeDisplayName(result);
+                    UpdateTreeNodeName(treeNode);
                     _view.SetImageIndex(treeNode, imageIndex, true);
                 }
             });
@@ -213,17 +213,19 @@ namespace TestCentric.Gui.Presenters
 
         public string GroupDisplayName(TestGroup group)
         {
-            return string.Format("{0} ({1})", group.Name, group.TestNodes.Count());
+            return $"{group.Name}{GetTreeNodeNameCountSuffix(group.TestNodes.Count())}";
         }
 
-        protected virtual string GetTreeNodeName(TestNode testNode)
+        private string GetTreeNodeNameCountSuffix(int count)
         {
-            return testNode.Name;
+            return $" ({count} " + (count > 1 ? "tests)" : "test)");
         }
 
-        protected string GetTreeNodeDisplayName(TestNode testNode)
+        private string GetTreeNodeDisplayName(TestNode testNode)
         {
-            string treeNodeName = GetTreeNodeName(testNode);
+            string treeNodeName = testNode.Name;
+            if (testNode.IsSuite)
+                treeNodeName += GetTreeNodeNameCountSuffix(GetTestCount(testNode));
 
             // Check if test result is available for this node
             ResultNode result = testNode as ResultNode ?? _model.TestResultManager.GetResultForTest(testNode.Id);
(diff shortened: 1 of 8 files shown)


src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
```diff
@@ -233,6 +235,16 @@ protected string GetTreeNodeDisplayName(TestNode testNode)
             return treeNodeName;
         }
 
+        private static int GetTestCount(TestNode testNode)
+        {
+            // Check if the node is a test case, and it's not filtered out (e.g. by outcome filter)
+            if (!testNode.IsSuite && testNode.IsVisible)
```
CharliePoole: This took me by surprise as I didn't remember adding IsVisible to TestNode. I see did it back in 2024. This seems a bit confusing. All the other "IsX" properties represent intrinsic characteristics of a Test Node, i.e. is it a namespace? is it a fixture?, whereas IsVisible means "Given the current filters, should we even include this in the tree?

In the engine, there was a rule (unwritten, of course) that we used properties for intrinsic characteristics and methods for things that could change from run to run... e.g. `GetTestCaseCount(TestFilter filter)`. Can you think of a naming convention for such things in the model? I'm OK with keeping it a property, but to me "IS" generally means something intrinsic to the instance, not something changeable. Picky, right? :-)

Also, this begs the question... if it's filtered out, should this method even be called?

src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
```diff
@@ -233,6 +235,16 @@ protected string GetTreeNodeDisplayName(TestNode testNode)
...
+                return 1;
+
+            // Iterate through all child nodes
+            return testNode.Children.Sum(GetTestCount);
```
CharliePoole, line 249: This line will execute for invisible test cases.

src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
```diff
@@ -213,17 +213,19 @@ protected void AddTestNodeMapping(TestNode testNode, TreeNode treeNode)
...
         {
-            return string.Format("{0} ({1})", group.Name, group.TestNodes.Count());
+            return $"{group.Name}{GetTreeNodeNameCountSuffix(group.TestNodes.Count())}";
         }
 
-        protected virtual string GetTreeNodeName(TestNode testNode)
+        private string GetTreeNodeNameCountSuffix(int count)
         {
-            return testNode.Name;
+            return $" ({count} " + (count > 1 ? "tests)" : "test)");
```
CharliePoole: Suggest `count == 1 ? "test" : "tests"` , giving "0 tests" rather than "0 test".

CharliePoole (approved):
Comments are all suggestions. Thank you.

src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
rowo360: Ok, I'll adapt it.

src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
rowo360: True, it's not necessary to iterate through any children, if the node itself is already hidden. I'll adapt it.

src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
rowo360: Yes, I introduced the `IsVisible `property sometime back when implementing the filtering. Your explanation about the "IsX" properties and their intrinsic behavior sounds reasonable. I want to support and continue it. 😄 
(I’ve probably seen so many "IsX" properties in the past that I’ve stopped noticing the subtle differences.)

So, I can do a separate commit to change this property to a method and adapt the entire code base. Here's the proposal:
- void SetFilteredOut(bool filterOut);
- bool GetFilteredOut();

Of course, we can continue to use the term visible (SetVisibility / GetVisibility). But maybe 'FilteredOut' is more descriptive.

CharliePoole: I find that "Filter" is already too overloaded. That is, a test can be excluded from the tree (and therefore the  test run) or it can be in the tree but excluded from the test run. This arises from the fact that we have two different kinds of filters... our own and NUnit's.

"Visible" is similarly ambiguous... i.e. a test is not visible if it's not in the tree at all, or if it's parent node is not expanded or if it has simply scrolled out of sight.

Here (I think) we are talking about test nodes that are not included in the tree at all. So something using Include/Exclude or maybe Show/Hide may make sense.

A separate commit makes sense.

rowo360:
@CharliePoole :
I adapted the code parts from the review and also renamed the `TestNode.IsVisible` property to `TestNode.FilteredOut`. Does this name fit better? Or do you have any further suggestions? Otherwise I would continue with the merge 😄

CharliePoole:
@rowo360 I'm OK with this. Thanks.

PR #1548, opened 2026-05-27
Tree node icons - ignore test case nodes without any test run result

### LINKED ISSUE #1536: Test list view: "Not run" tree node gets test result icon
**Describe the bug**
After a test run the "Not run" tree node suddenly gets a test result icon. I observed it at least for the outcome/duration grouping, but might also happen for the category grouping.

**To Reproduce**
1. Load test assembly mock-assembly.dll
2. Switch to outcome grouping in test list view
3. Run one single test: for example MockTestFixture.FailingTest
4. Observe test result icon for "Not run"

**Expected behavior**
I think that test result icon for "Not run" should be kept unchanged after the test run. We apply test result for the TestNode MockTest4 and MockTest5  already from the start because they are of type 'Ignored' or failing. But we don't propagate this state to the parent hierarchy.

rowo360:
This issue is caused by certain TreeNodes which get an ImageIndex by default when loading the test assembly. For example, Ignored test cases will get a dedicated ImageIndex and also NotRunnable nodes. This should help the user to quickly identify such kind of tests without even starting a test run.
However, in the NUnit view such kind of nodes doesn't contribute to the overall test result of any parent nodes as long as they were never executed.

We recently added that we calculate the icon of a TreeNode based on the ImageIndices of his child TreeNodes. This calculation considers only the WinForm TreeNodes hierarchy and ImageIndex values, but doesn't consider if an ImageIndex was set by default without any test run. As a result, these default ImageIndices were propagated to the parent nodes, leading to an inconsistent behavior compared with the NUnit view. This happens when calculating the 'running image index' and also when switching between NUnit and list view.

I fixed this by checking if any test result is present for a TestNode. If no test result is present, the node is ignored for the calculation. (Fixes #1536)

### DIFF
--- a/src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
+++ b/src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
@@ -345,12 +345,14 @@ namespace TestCentric.Gui.Presenters
             {
                 ResultNode resultNode = GetResultForTest(testNode);
                 if (resultNode != null)
+                {
                     treeNode.ImageIndex = treeNode.SelectedImageIndex = CalcImageIndex(resultNode);
 
-                // NOTE: the loop only executes at the point in the hierarchy where the current
-                // TreeNode represents a TestNode and the parent is a TestGroup;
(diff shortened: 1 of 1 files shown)


CharliePoole (approved):
Very neat how the addition of the braces works. Thank You.

PR #1549, opened 2026-05-29
Issue 1500

### LINKED ISSUE #1500: Filter tree for outcome, category or text is not working

rowo360:
This PR fixes #1500. The filtering in general was already available, the remaining part was to update the tree icons based on the active filter.

Actually this worked already as expected for the TestList view and I just took care that the approach is applicable for the NUnit view as well. The relevant code is located in one single method DisplayStrategy.ApplyResultsToTree(): this method is called when the tree is build up (Project loaded, switch between TestList/Nunit view, but also when applying a filter). 

Only the leaf nodes (e.g. test case nodes) are associated with TestNodes in the TestList view (all other tree nodes are associated with TestGroups). If such a leaf node has a test result, it's applied to the node and afterwards the ImageIndex of all nodes in the parent hierarchy is calculated.

That's the existing approach for the TestList view. I adapted the code slightly, so that it's now a approach working for TestList and NUnit view. 

The 2nd commit is a bit off topic:
I noticed some kind of flickering when executing tests. It turns out that it's caused by the call `UpdateTreeNodeName` in  `OnTestFinished()` - we already noticed it sometime back that WinForms tree is suffering if we update a node's name too often.  Even if we're not changing anything at all, but just set the identical name anew. So, I first added a check that the name really changed at all. But finally dropped the entire call because the names kept constant here.

### DIFF
--- a/src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
+++ b/src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
@@ -127,10 +127,7 @@ namespace TestCentric.Gui.Presenters
                 int imageIndex = CalcImageIndex(result);
 
                 foreach (TreeNode treeNode in GetTreeNodesForTest(result))
-                {
-                    UpdateTreeNodeName(treeNode);
                     _view.SetImageIndex(treeNode, imageIndex, true);
-                }
             });
         }
 
--- a/src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
+++ b/src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
@@ -339,18 +339,17 @@ namespace TestCentric.Gui.Presenters
 
         private void ApplyResultsToTree(TreeNode treeNode)
         {
-            TestNode testNode = treeNode.Tag as TestNode;
-
-            if (testNode != null)
+            // Consider only treeNodes representing a test case. Test results from fixtures/assemblies might be inaccurate
+            // if some filters are applied (e.g. outcome filter) and some test case nodes are currently not displayed in the tree.
+            if (treeNode.Tag is TestNode testNode && !testNode.IsSuite)
             {
                 ResultNode resultNode = GetResultForTest(testNode);
                 if (resultNode != null)
                 {
                     treeNode.ImageIndex = treeNode.SelectedImageIndex = CalcImageIndex(resultNode);
 
-                    // NOTE: the loop only executes at the point in the hierarchy where the current
-                    // TreeNode represents a TestNode and the parent is a TestGroup;
-                    for (var parent = treeNode.Parent; parent?.Tag is TestGroup; parent = parent.Parent)
+                    // Calculate the ImageIndex of the parent hierarchy 
+                    for (var parent = treeNode.Parent; parent?.Tag is ITestItem; parent = parent.Parent)
                         parent.ImageIndex = parent.SelectedImageIndex = Math.Max(parent.ImageIndex, treeNode.ImageIndex);
                 }
             }


src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
```diff
@@ -127,10 +127,7 @@ public virtual void OnTestFinished(ResultNode result)
                 int imageIndex = CalcImageIndex(result);
 
                 foreach (TreeNode treeNode in GetTreeNodesForTest(result))
-                {
-                    UpdateTreeNodeName(treeNode);
                     _view.SetImageIndex(treeNode, imageIndex, true);
```
CharliePoole, line 130: Isn't the call to UpdateTreeNodeName needed if elapsed time is being displayed?

src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
```diff
@@ -339,18 +336,17 @@ protected void ApplyResultsToTree()
...
                 ResultNode resultNode = GetResultForTest(testNode);
                 if (resultNode != null)
                 {
                     treeNode.ImageIndex = treeNode.SelectedImageIndex = CalcImageIndex(resultNode);
 
-                    // NOTE: the loop only executes at the point in the hierarchy where the current
-                    // TreeNode represents a TestNode and the parent is a TestGroup;
-                    for (var parent = treeNode.Parent; parent?.Tag is TestGroup; parent = parent.Parent)
+                    // Calculate the ImageIndex of the parent hierarchy 
+                    for (var parent = treeNode.Parent; parent?.Tag is ITestItem; parent = parent.Parent)
```
CharliePoole, line 349: The test in this loop combined with the new test you added above makes this equivalent to the original test I had here, so that for the TestList strategy, we are only exercising this at the boundary between actual TestNodes and groups. It's clever but non-obvious... took me about 10 minutes to work it out. Can you add a bit more commentary here explaining how it works for both strategies?

Also, since the Tag is __always__ an ITestItem if non-null, perhaps change the test to `parent is not null`.

CharliePoole (approved):
See my inline suggestions. Thank You!

src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
rowo360: I wanted to start with some minimal modification, so it's clear that the original approach is still applied. But, I also noticed that it's always a ITestItem and the loop only terminates because the parent is null at some point in time. So, I can adapt it, to make it clearer.

But first back to some general explanation:
The crucial part is that the calculation only starts for test case nodes (leaf nodes)
`if (treeNode.Tag is TestNode testNode && !testNode.IsSuite)`

The TestList view uses TestNodes only for leaf tree nodes, all other tree nodes use TestGroups. So indeed, the TestList view started the calculation only for test case nodes already in the past. The NUnit view uses TestNodes for all kind of tree nodes, so I added the new check `!testNode.IsSuite` which checks that only test cases nodes passes and are processed further. So overall, for TestList and NUnit view the calculation process only starts for test case nodes, now.

Next, we check if the test case node has some test result from a previous test run. Only in that case we process further. That's also unchanged.

If some test result is present, we navigate through the entire parent hierarchy and calculate the ImageIndex based on the ImageIndex of all child tree nodes (max of all values). One drawback might be, that this is done over and over again. So, when the next test case of a fixture is processed, the entire parent hierarchy is updated again. However, I didn't notice any performance issue yet - unlike setting the tree node name over and over again. And well, that's exactly how the TestList view already did it previously - that's right, isn't it?
I think this detail can be optimized, but I didn't want to replace the entire method with some different implementation from the start. But instead, I wanted to show that the approach hasn’t changed much.

One additional remark:
Test results from fixture/assembly nodes are not used anymore in this calculation at all. We are starting at the leaf nodes and take their ImageIndices to determine the parent ImageIndices recursively.

src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
rowo360: I remember that it was precisely because of this flickering that we didn’t want to trigger showing the elapsed time at OnTestFinished anymore. Instead, we have accepted that all elapsed times are only displayed at the end of the test run.

src/GuiRunner/TestCentric.Gui/Presenters/DisplayStrategy.cs
CharliePoole: That's basically what I meant by "equivalent to the original test. I understood all of that after looking at it for a bit (~10 minutes), which suggests the need for a small comment. The fact that you or I can figure it out is not a good criterion for general use, I dare say. :-)

I do like the idea that you are not looking up the results of suites if it works. But that could be bad for separation of concerns if it means that the UI code has to understand the inner workings of each different type of Suite. So let's keep an eye on this. I think the type of test that may break it is  a Theory.

Thinking in a theoretical sense, there are two sorts of suites... generally, higher level suites like namespaces simply aggregate what they contain, basically using the same rules we use for TestGroups. But lower-level suites (TestFixtures, Theories, Parameterized methods) have the ability to define their own rules. A Theory is the best practical example of this at the moment.

Anyway, I think this is still a good PR.

PR #1550, opened 2026-06-03
Enable test duration features (show and sort) for TestList view

### LINKED ISSUE #1460: Clarify details of ShowTestDuration and Sorting by Duration
The 'ShowTestDuration' and 'SortBy... Duration' menu items appear in the tree context menu for both `NUnitTreeDisplayStrategy` and `TestListDisplayStrategy`.

For `NUnitTreeDisplayStrategy` it is implemented correctly, with the possible exception of the sort. It's not clear how or even if the sort should apply to suites, i.e. TestFixtures and above in the hierarchy.

For `TestListDisplayStrategy` the menu items are present but don't seem to work correctly, primarily when assemblies or fixtures are displayed.

We should clarify just how this should all work in this issue and implement any changes required.

rowo360:
This PR fixes #1460 by enabling test duration features (show duration and sort by duration) also for the TestList view.

Actually we started a discussion in the issue whether we want to support test duration at all in the TestList view, but we haven't made (or wrote down) a final decision yet. We should continue this discussion, but as you can probably guess, I'm in favor to support it, if I'm already coming in with a PR. 😄 

Overall, I think the test duration is just as helpful in the TestList view as in the NUnit view. Regardless if it's the simple list view or if it's the duration grouping, showing the test duration or even sorting by duration gives a quick feedback which test cases run the slowest of all. So, after identifying the relevant tests quickly, the user can take care about these tests and try to optimize them.

From a technical point of view, the challenge is, of course, that we don’t have a test duration for TestGroups out-of-the-box. I try to solve it identically as we did with test results lately: iterate through all child tree nodes and sum up their test duration. That means that only leaf nodes (test case nodes) contribute with their test duration. The test duration of fixture or assembly nodes are not considered at all. I'm aware that the resulting number if not 100% accurate or identical to the NUnit view: for example if a [Setup] method contributes to the test duration of a fixture, it's considered and displayed in the NUnit view. But that number is completely ignored in the proposed approach for the TestList view. Nonetheless I think that the proposed approach will be helpful to identify the slowest test cases.

The test durations for TestGroups are calculated only at the end of a TestRun and when loading the tree.

### DIFF
--- a/src/GuiRunner/TestCentric.Gui/Presenters/TestListDisplayStrategy.cs
+++ b/src/GuiRunner/TestCentric.Gui/Presenters/TestListDisplayStrategy.cs
@@ -63,7 +63,12 @@ namespace TestCentric.Gui.Presenters
             visualState?.ApplyTo(_view.TreeView);
 
             ApplyResultsToTree();
+            SetTestDurationOfGroups();
             UpdateTreeNodeNames();
+
+            // TestGroup duration is only available after entire tree is build up => sort tree now
+            if (_view.SortCommand.SelectedItem == TreeViewNodeComparer.Duration)
+                _view.Sort();
 
             _view.EnableTestFilter(true);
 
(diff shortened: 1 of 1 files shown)


CharliePoole:
Let's hold up a bit on this while we resolve the question in the issue itself. I'll post there because I think there are some serious questions about what timing means if applied to groups.

CharliePoole (approved):
This gets several of the use cases we outlined working so let's go ahead and merge it. We can do new issues for any use cases that we feel should be addressed. Thanks for your patience.

PR #1552, opened 2026-06-15
Recognize use of --Trace:Debug and save logs in correct location

### LINKED ISSUE #1544: Option --trace:Debug not passed to agent
When running tests using `build -t Package --trace:Debug`, only the internal log for the GUI itself appears in the work directory. The corresponding agent log shows up in the root directory with zero length. Since this works correctly when running tests of the NUnit ConsoleRunner, which uses the same agents, the working theory is that the GUI is not passing the correct arguments.

CharliePoole:
Fixes #1544


CharliePoole:
@rowo360 Lots of little structural changes as I tried to understand why this wasn't working.

PR #1559, opened 2026-07-07
Align progress bar width to window size

### LINKED ISSUE #1558: Text in progress bar is not visible
**Describe the bug**
The width of the progress bar doesn't fit into the window width. It seems that the width is somehow larger. As a result the text in the progress bar is not completely visible.

rowo360:
I simply restored the width of the progress bar which were present sometime back. This value is aligned to the size of the window and other controls. Now, the text in the progress bar is centered again.

I assume that the width was accidently changed in commit 
https://github.com/TestCentric/TestCentric-Gui/commit/93f7fdb1a180f41bbe79ba40744e5bbace5c7742

This PR fixes #1558.


rowo360:
@CharliePoole:
There is really not much to review here and it feels a bit odd to request a review. 
But I prefer to stick to the process - at least you get aware of the issue. 😄

CharliePoole (approved):
Looks good. Thanks.

PR #1563, opened 2026-07-22
Issue 1497

### LINKED ISSUE #1497: Determine how to support projects with multiple configurations
Both NUnit and Visual Studio projects support multiple configurations. In the NUnit V2 and TestCentric V1 GUI, it was only possible to load a single project at one time. We allowed selecting the configuration for that project from a menu.

TestCentric supports multiple projects open at one time. We need to determine how to support project configurations.

### Option 1 - Top Level Menu

This sounds simple but only works smoothly if all projects contain the specific config chosen. We would either need to ignore the config for projects without it - or without any configs in the case of an NUnit project or simply disable the menu item when all projects did not have the same set of configurations.

### Option 2 - Context menu item on a per-project basis.

This would be simpler in many ways. The menu item would only be active for nodes representing projects and could be populated dynamically using the specific configs used in that project. The downside is that it would require configuration of each project by the user. But it would also allow use of different configs for each project, e.g. use of debug for the project currently being debugged.

This issue is blocked until we add support for one project type extension, but we can decide which way to go in any case.

rowo360:
This PR adds a new main menu item to select a configuration present in the any subpackage.
The menu item is disabled in case there's no configuration available.

Fixes #1497

### DIFF
--- a/src/GuiRunner/TestCentric.Gui/Presenters/ConfigurationSelectionController.cs
+++ b/src/GuiRunner/TestCentric.Gui/Presenters/ConfigurationSelectionController.cs
@@ -1,0 +1,90 @@
+// ***********************************************************************
+// Copyright (c) Charlie Poole and TestCentric contributors.
+// Licensed under the MIT License. See LICENSE file in root directory.
+// ***********************************************************************
+
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Windows.Forms;
+using NUnit.Common;
+using NUnit.Engine;
+using TestCentric.Gui.Elements;
+using TestCentric.Gui.Model;
+using TestCentric.Gui.Views;
+
+namespace TestCentric.Gui.Presenters
+{
+    internal class ConfigurationSelectionController
+    {
+        private ITestModel _model;
+        private IMainView _view;
+
+        public ConfigurationSelectionController(ITestModel model, IMainView view)
+        {
+            _model = model;
+            _view = view;
+        }
+
+        public void PopulateMenu()
+        {
+            IPopup configurationMenu = _view.SelectConfigurationMenu;
+            configurationMenu.MenuItems.Clear();
+
+            string activeConfig = GetActiveConfiguration();
+            foreach (string configName in GetPackageConfigurations())
+            {
+                var menuItem = new ToolStripMenuItem(configName);
+                menuItem.Tag = configName;
+                menuItem.Checked = configName == activeConfig;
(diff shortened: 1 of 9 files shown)


CharliePoole:
@rowo360 
Code looks great!

I'd like to revise some of the nomenclature seen by the user and perhaps some internal naming as well.

We have the TestCentric project, which contains Test Files. Test Files may be either assemblies or supported projects. This is potentially confusing for the user and for us writing code... i.e. when we see a "project" mentioned, are we talking about the TC project or one of those "projects" under it. Here's my suggestion...

1. Test Files are either assemblies or supported "sub-projects", like NUnit or VS projects.
2. The menu entry should refer to  Sub-Project Configuration.
3. As for agents, we need a Default entry on the sub-menu, for when there is no selected configname. In that case, the ActiveConfig in the file rules.
4. I think we will eventually want to change a lot of names internally as well, just for clarity. Do that where you think it's needed in the files you are changing.

src/GuiRunner/TestCentric.Gui/Presenters/ConfigurationSelectionController.cs
```diff
@@ -0,0 +1,92 @@
...
+            return configurations.Where(c => !string.IsNullOrEmpty(c)).Distinct().ToList();
+        }
+
+        private string GetActiveConfiguration()
+        {
+            if (!_model.IsProjectLoaded)
+                return string.Empty;
+                        
+            TestPackage package = _model.TopLevelPackage.SubPackages.FirstOrDefault();
+            return package?.Settings.GetValueOrDefault(SettingDefinitions.ActiveConfig);
```
CharliePoole, line 69: What if we changed this setting to simply "ConfigName" reserving "ActiveConfig" for the setting that's inside the NUnitProjectFile? We are overriding the ActiveConfig here, not changing it. I don't see the property referenced in too many places.

CharliePoole (changes requested):
Thanks for this! Please see my general comment on naming and use your judgement for what to do right now vs later.

rowo360:
Sorry, it took a long time - too long. :worried:
Actually, I kept thinking that I'd be able to get back to it soon, but my work has kept me way too busy. But the good news is that I'll definitely have time for TestCentric over the next few days and can work on the next assignments. 😄 

I'm merging this PR now. I appreciate your idea about using unambiguous names for projects and subprojects. So, it' clear what we're talking about when using one or the other term. However, I didn't start any action right now.

Issue #1487, opened 2026-04-02
Fix or remove the font change menu items


CharliePoole:
The Font change menu items and dialog don't work fully, for two reasons...

1. The correct font is not propagated to all windows we create in our code.
2. Where it is passed, the field sizes do not always porperly accomodate larger fonts.

For this issue, we will first examine the possibility of eliminating this feature entirely and instead relying on facilities provided by Windows or any other host OS to accomplish the same purpose. If this works, we can remove both the menu items and the dialog code. If not, we will need to do some design work to try to make our own code work properly. This implies that most windows would need to adjust to Font changes dynamically so it may end up being postponed beyond the 2.0 release.

CharliePoole:
Moving from beta8 to release milestone. I'm leaning toward dropping the font feature entirely.

rowo360:
The font selection menu items are no longer displayed to the user (PR #1568). We want to observe if the font size scaling provided by Windows is sufficient. If successfully, we can remove the entire code and close this issue.

PR #1568, opened 2026-08-05
Disable user defined font selection

rowo360:
This PR contributes to #1487 by disabling the menu items to select a user defined font for the UI.
I decided to take a soft approach by introducing a compiler constant (DISABLE_FONT_SETTING), similar to what we did with MiniGui. This means that all the FontSetting code is still there, and I haven't removed it yet. The compiler constant controls the visibility of the corresponding menu items.

All of this assumes that we still do not want to support individual font settings. However, the user can still use the general Windows setting 'scaling' to adapt the font size to his display.


CharliePoole (approved):
Thanks for doing this... let's merge but not close the issue.

We can use it for a while and make sure that the system-provided resizing does the job fully and then remove it permanently before closing the issue.

CharliePoole:
This reminds me that we need to make a final decision about the mini-gui. Let's talk. :-)

PR #1570, opened 2026-08-05
Improve test run performance for NUnit tree display

### LINKED ISSUE #1569: Improve test execution performance for the NUnit tree display
**Describe the bug**
This issue is targeting the performance when executing tests. It's about quick wins for the NUnit tree display. The performance delay using the Test list display is more severe but needs some more preparation. I'll write a separate issue about it.

I assume that our general tree display and grouping options are settled and it's not wasted to have a look to the performance aspect now. 😄 

Here's some measurement:
I use a sample test project with 50 fixtures, each of them containing 100 test cases, resulting into 5000 test cases overall. That's not super huge, but larger than those projects we usually work with. The test cases just contain one single Assert statement, which means they are executed really fast.

Test execution time:
NUnit console: 17s
TestCentric NUnit display: 83s
TestCentric Test list display + outcome grouping: I decided to terminate the run after 5min

**Expected behavior**
Well, I don't expect that we execute as fast a NUnit console, because the UI might add some additional delays. But we should remove all obvious and easy bottlenecks.

rowo360:
This PR resolves #1569 by adding two quick wins for the test run performance.
It turns out that all performance delay was caused by updating the tree nodes icons over and over again, We do this repeatedly when starting a test case (recently added 'running icon') and repeatedly when finishing a test case.

So, besides reducing the flickering, also the execution time drops:
from original 83s, to 53s by applying the first fix and finally to 18s by applying the 2nd fix. 😄 

The fix in method `TestTreeView.SetImageIndex()` probably needs no further explanation. I actually would have expected WinForms to have implemented this on its own.

The 2nd fix in method `DisplayStrategy.OnTestFinished()` is specific to the NUnit tree display. I recall that we needed to set the applyToParent to true for the Test list display, to update the top most grouping nodes and also to update the running icons. But for the NUnit tree display this parameter can be set to false. And as a result, far fewer TreeNode icons are updated, improving the performance further.
But I expect that this solution is not final. I assume when analyzing the Test list display performance we'll stumple over this issue again. Maybe we find a solution to set the parameter in general to false or find another approach.
But that's beyond my current idea of a 'quick win' and want to tackle the Test list performance on a separate issue.

### DIFF
--- a/src/GuiRunner/TestCentric.Gui.Tests/Presenters/TestTree/WhenTestCaseCompletes.cs
+++ b/src/GuiRunner/TestCentric.Gui.Tests/Presenters/TestTree/WhenTestCaseCompletes.cs
@@ -63,7 +63,7 @@ namespace TestCentric.Gui.Presenters.TestTree
             _model.Events.TestLoaded += Raise.Event<TestNodeEventHandler>(new TestNodeEventArgs(testNode));
             _model.Events.TestFinished += Raise.Event<TestResultEventHandler>(new TestResultEventArgs(resultNode));
 
-            _view.Received().SetImageIndex(Arg.Any<TreeNode>(), expectedIndex, true);
+            _view.Received().SetImageIndex(Arg.Any<TreeNode>(), expectedIndex, false);
         }
     }
 }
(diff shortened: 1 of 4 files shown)


CharliePoole (approved):
This is great! Big performance win for just a few lines changed.

PR #1572, opened 2026-08-10
Update NUnit dependencies

CharliePoole:
This includes a breaking NUnit change.


CharliePoole:
@rowo360 I usually just merge this sort of dependency update without a review. But, in this case, I'm concerned about creating conflicts with your work on nullability. Please take a look and let me know...

1. If you're OK with my merging this before you
2. If you think I should make further changes up the call chain as mentioned in my comment on TestCentricRunner.cs.

rowo360 (approved):
No problem, feel free to merge the branch. I'll rebase afterwards.
But actually, it would be convenient for me if we kept the StopRun method unchanged with the force parameter for a short period of time. I expect that I'll create the PR for the Nullability tomorrow.

CharliePoole:
Merged. After you merge, I'll try to quickly change some of the calls before you proceed with the next assembly.

Issue #1492, opened 2026-04-02
Enable nullability in TestCentric projects


CharliePoole:
We've discussed this but never had time to implement it while also working on new features. I keep running into bugs that would never get past the compiler if we had implemented nullability in our projects.

It appears now that we may have time between beta.8 release and the first rc, while we are waiting for fixes to the NUnit engine and agents, so I propose that we work on nullability at that point.

We may also think of other changes that would improve the build process to schedule at the same time.

For nullability, I think it should be done one assembly at a time. Initially we could implement a minimum level of compiler checks beyond the basics but we might want to add additional analyzers over time.

rowo360:
@CharliePoole:
I could take on this task next—it works out pretty well at this point in time, doesn't it?
I would start with project TestCentric.Gui.Model, enable the nullability with `<Nullable>enable</Nullable>` and fix all resulting compiler errors. A quick preview shows that it will result into ~100 compile errors. Many of them complaining that some members are not initialized in the constructor. I think all of them can be fixed easily, but of course it involves that several classes will get changed. That will make the code review more intense.

I also notice that some attributes for handling nullability are not yet available in .Net 4.6 (or 4.8). For example MemberNotNull, NotNull or NotNullWhen. So, I need to apply a different solution. I also need to do some further research because in one article it was mentioned that the Language version needs to be increased to enable the feature: `<LangVersion>8.0</LangVersion>`
Did you also have to take such peculiarities regarding .NET 4.6 into account when using NUnit?

rowo360:
I figured it out: 😄 
There's this nuget package 'Nullable' which allows to use .Net nullable attributes in older target frameworks. That's exactly what is needed and which is also used by the NUnit console project. So, I would do exactly the same thing.
Changing the LangVersion beyond 7.3 is not required or recommended for .Net 4.6 projects.

CharliePoole:
Yes! You answered yourself before I  could. :-) 

Shall we do separate PRs for each project and merge them as we go?

rowo360:
I would handle it on a project-by-project basis with separate PRs rather than enabling it globally right away. Simply to keep the number of changes manageable. In addition, I'll split it into several commits so that it's easier to keep track of everything during the review.
In most cases, I'll probably add the `? operator`. Only in exceptional cases—and when it's unambiguous—will I include a `Guard` statement (or similar). I would also point that out during the review.

rowo360:
A brief update:
[...]

PR #1573, opened 2026-08-11
Activate nullability checks for project TestCentric.Gui.Model

rowo360:
This PR contributes to #1492 by activating nullability checks for project `TestCentric.Gui.Model`
It includes many code changes. Therefore, I splitted my work into separate commits to keep track of everything during the review.
Here some comments:

1. Commit:
Enabling Nullability and add nuget package Nullable

2. Commit:
Contains all simple fixes for nullability compile errors. Mainly by applying the `?` operator. The only exception is class `AssemblyResolver` in which I also used the `!` operator.

3. Commit:
All of the following commits are not just local fixes; they also affect other classes. This commit starts with adapting class `TestCentricRunner`, but that leads to property `TestModel.LoadedTests` which also needs to be adjusted. In some cases I applied a `Guard.OperationValid()` statement, in which I expect that LoadedTests should never be null.

4. Commit:
The actual code change is in the last file `XmlHelper.cs`. All other changes are resulting adjustments. For example, this includes adjustments to class `TestNode`. However, this commit was before your suggestion about moving the assignment and check into the constructor. So, I applied your suggestion in a subsequent commit.

5. Commit:
I took care about class TestResultManager in this commit. One idea for further improvement with respect to nullability might be method GetResultForTest. Currently the signature is
`ResultNode? GetResultForTest(string id)`
That means that every client needs to check for null. Using the nullability annotations it can be improved to:
bool TryGetResultForTest(string id, [NotNullIWhen(true)] out ResultNode resultNode)

6. Commit:
In class `TestCentricProject` I need to check the `subPackage.Fullname` property for null. My assumption is that it cannot be null at those locations.

7. Commit:
Finally class TestModel was in focus. Here I used the nullability attributes MemberNotNullWhen.
One remark regarding property TestModel.ActiveTestItem. While checking the code base, if this property might be null or not, I noticed that this property is no longer set anywhere. Therefore I suggest to remove it - but didn't want to mix this commit with other stuff at this point in time. 

8. Commit:
After fixing all compiler errors, I noticed that some tests are failing due to null checks in class TestNode. I needed to adapt test code to create accurate TestNode instances.

9. Commit:
Here I applied your suggestion to class TestNode - moving the assignment and null checks into the constructor. Afterwards quite a lot of tests needs to be adapted further to create accurate TestNodes again 😄

### DIFF
--- a/src/GuiRunner/TestModel/AssertionResult.cs
+++ b/src/GuiRunner/TestModel/AssertionResult.cs
@@ -4,6 +4,7 @@
 // ***********************************************************************
 
 using System.Xml;
+using NUnit;
 
 namespace TestCentric.Gui.Model
 {
@@ -17,12 +18,12 @@ namespace TestCentric.Gui.Model
 
         public AssertionResult(XmlNode assertion)
         {
-            Status = assertion.GetAttribute("label") ?? assertion.GetAttribute("result");
+            Status = (assertion.GetAttribute("label") ?? assertion.GetAttribute("result")).ShouldNotBeNull();
             Message = assertion.SelectSingleNode("message")?.InnerText;
             StackTrace = assertion.SelectSingleNode("stack-trace")?.InnerText;
         }
 
-        public string? Status { get; }
+        public string Status { get; }
         public string? Message { get; }
         public string? StackTrace { get; }
     }
--- a/src/GuiRunner/TestModel/AssertionResult.cs
+++ b/src/GuiRunner/TestModel/AssertionResult.cs
@@ -22,7 +22,7 @@ namespace TestCentric.Gui.Model
             StackTrace = assertion.SelectSingleNode("stack-trace")?.InnerText;
         }
 
-        public string Status { get; }
+        public string? Status { get; }
         public string? Message { get; }
         public string? StackTrace { get; }
     }
--- a/src/GuiRunner/TestModel/AssertionResult.cs
+++ b/src/GuiRunner/TestModel/AssertionResult.cs
@@ -23,7 +23,7 @@ namespace TestCentric.Gui.Model
         }
 
         public string Status { get; }
-        public string Message { get; }
-        public string StackTrace { get; }
+        public string? Message { get; }
+        public string? StackTrace { get; }
     }
 }
--- a/src/GuiRunner/TestModel/GuiOptions.cs
+++ b/src/GuiRunner/TestModel/GuiOptions.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
 using System.Text;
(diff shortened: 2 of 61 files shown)


src/GuiRunner/TestModel/GuiOptions.cs
```diff
@@ -128,8 +129,8 @@ private void ParseArguments(string[] args)
             for (int i = 0; i < args.Length; i++)
             {
                 string arg = args[i];
-                Option option;
-                string val;
+                Option? option;
```
CharliePoole: Suggest deleting declarations of `Option` and `val` and changing the next line to

```
if (TryParseOption(arg, out Option option, out string val)
```

src/GuiRunner/TestModel/GuiOptions.cs
```diff
@@ -152,7 +153,7 @@ private void ParseArguments(string[] args)
             }
         }
 
-        private bool TryParseOption(string arg, out Option option, out string value)
+        private bool TryParseOption(string arg, [NotNullWhen(true)] out Option? option, out string? value)
```
CharliePoole, line 154: See above comment...

```
private bool TryParseOption(string arg, [NotNullWhen(true)] out Option? option, [NotNullWhen(true) out string? value)
```

This assumes the overload is only used with options where value must always be non-null. If cases exist where ti may be null, then we need separate overloads, with and without value and you may want to postpone this change.

src/GuiRunner/TestModel/AssertionResult.cs
```diff
@@ -22,8 +22,8 @@ public AssertionResult(XmlNode assertion)
             StackTrace = assertion.SelectSingleNode("stack-trace")?.InnerText;
         }
 
-        public string Status { get; }
-        public string Message { get; }
-        public string StackTrace { get; }
+        public string? Status { get; }
```
CharliePoole: I believe that Status must logically be non-null in all cases. If so,  then a null status could fail in the constructor. Of course, just as for `TestNode` this might cause a lot of tests to break and I'm OK with saving this for a later pass if you prefer.

src/GuiRunner/TestModel/Services/AssemblyWatcher.cs
```diff
@@ -12,16 +12,11 @@ namespace TestCentric.Gui.Model.Services
...
 
-        protected System.Timers.Timer timer;
-        protected string changedAssemblyPath;
-
-        protected FileInfo GetFileInfo(int index)
-        {
-            return files[index];
-        }
+        protected System.Timers.Timer? timer;
+        protected string? changedAssemblyPath;
```
CharliePoole, line 19: This class could do with a rewrite, especially now that we have enabled nullability. Let's save it for later, however. :-)

rowo360:
@CharliePoole:
This is only a friendly reminder about the review - just in case you missed the reviewnotification in your stream of incoming messages. If you are busy with NUnit issues, it's no problem - take your time 😄

CharliePoole (approved):
Comments are all optional or can be postponed.

CharliePoole:
Good you reminded me. I did the review right away but it was pending.

rowo360:
Thanks for the review. I'll apply your suggestions regarding `GuiOptions` and `AssertionResult` - fortunately, it doesn't affect any tests this time. But I will class `AssemblyWatcher` unchanged.

After merging I'll continue with nullability for assembly TestCentric.Gui - but not pushing anything, but wait for your adaptations first.

CharliePoole:
I'll merge my PR in the morning, when my brain will probably be more alert. :-)

PR #1574, opened 2026-08-14
Activate nullability checks for project TestCentric.Gui

rowo360:
This PR contributes to #1592 by activating nullability checks for project `TestCentric.Gui`
While doing this task I had the impression that some WinForm concepts doesn't fit best to nullable checks. Or some further refactorings would be required to improve the code. Here some two examples:
`TestNode testNode = treeNode.Tag as testNode;`
`ToolStripMenuItem item = sender as ToolStripMenuItem;`

Both statements needs (minor) adjustments with nullable checks enabled. However, we know in those examples that the Click event is triggered from a ToolStripMenuItem or that only TestNode or TestGroup instances are set as Tag properties of a TreeNode. Sometimes I didn’t need to make many further adjustments. Where necessary, I skipped the execution, but without triggering any messages or exceptions.

Another repeating issue was about WinFrom Dialogs or Forms which do their initialization not in the constructor, but in a separate method: Init() or Activate(). For example, the TestModel or TestCentricProject is passed into those classes in these methods. So, the nullable checks complains that not all nullable members are initialized properly in the constructor. These kind of compile errors could be fixed by adding some dummy initialization in the constructor or annotating that the null value is OK for now.

Besides WinForm also our TestModel events were interesting: the compiler doesn't know that after a TestModel.OnLoaded event the TestModel.TestCentricProject won't be null anymore. Or if a save command is executed, that a project should be loaded.
In all these cases I added `Guard.OperationValid` statements, to check the conditions. If in the future there's some wrong with our logic or events are triggered in an unexpected order, we'll get triggered now.

One remark to my 2nd commit:
Certain files are ignored for nullability by default - for example WinForm designer files. So, I created all missing designer files and moved the members and the InitializeComponent method to that file. Otherwise, I would get a lot of compiler errors that all those members are not initialized properly in the constructor.

Overall quite a lot of code changes, but I tried to keep the adjustments simple.


src/GuiRunner/TestCentric.Gui/Controls/ExpandingLabel.cs
```diff
@@ -35,6 +36,7 @@ public class ExpandingLabel : Label
         /// Returns true if the label is expanded, i.e. showing it's TipWindow.
         /// </summary>
         [Browsable(false)]
+        [MemberNotNullWhen(true, nameof(_tipWindow))]
```
CharliePoole, line 39: This does half the job... i.e. `_tipWindow` may be non-null if the window was created but is not expanded. How about using `[MemberNotNull]` on `Expand()`?

CharliePoole (changes requested):
I added one small inline suggestion. Thanks for doing this!

PR #1575, opened 2026-08-14
Eliminate vestiges of StopRun(bool) left from PR 1573

CharliePoole:


src/GuiRunner/TestModel/TestCentricRunner.cs
```diff
@@ -96,21 +96,28 @@ private void OnTestRunFinished(TestResultEventArgs args)
...
-                    ActiveTestRun?.ForcedStop();
-                else
+            if (ActiveTestRun is not null)
+                // Async to avoid blocking the main thread for incoming test events in between
+                Task.Run(() =>
+                {
                     ActiveTestRun?.RequestStop();
-                ResetActiveTestRun();
-            });
+                    ActiveTestRun?.Dispose();
```
rowo360: Existing method `ResetActiveTestRun` executes exactly these two lines: Dispose() + set to null. Maybe we should invoke this method instead?

src/GuiRunner/TestModel/TestCentricRunner.cs
```diff
@@ -96,21 +96,28 @@ private void OnTestRunFinished(TestResultEventArgs args)
...
+        }
+
+        public void ForcedStop()
+        {
+            if (ActiveTestRun is not null)
+                // Async to avoid blocking the main thread for incoming test events in between
+                Task.Run(() =>
+                {
+                    ActiveTestRun?.ForcedStop();
+                    ActiveTestRun?.Dispose();
```
rowo360: Again method `ResetActiveTestRun` can be used here.

rowo360 (approved):
Looks fine - the names matches now with the names used by NUnit. That avoids any deviation.

I suggest that you merge your PR first. Afterwards I update my PR and check if any further nullable checks must be added.

src/GuiRunner/TestModel/TestCentricRunner.cs
CharliePoole: Ah... that's a refactoring I started and left unfinished. I was going to start by removing that method, creating a bit of duplication, which I would then remove by doing some re-design of how we do disposal. "Creating duplication before you removing it" was a part of my training and coaching practice, in competition with the more simplistic slogan "Don't repeat yourself" some others use.

Anyway, I think I'll put it back the way it was for now, since this refactoring might have a lot of other effects and doesn't really relate to the purpose of this PR.

FWIW, when I see `Dispose` followed by setting null, it makes me suspect that disposal isn't properly implemented in a class. The "smell" is that the null value of the object is apparently being used as a flag, which is usually not a good idea in modern C#. That seems to be true here but it isn't causing any active problems so I'll come back to it another time. You have to stop somewhere!

PR #1577, opened 2026-08-19
Add nullability checks in project NUnit.UIException

rowo360:
Next project `NUnit.UIException` with enabled nullable checks - contributing to #1492

In most locations, I have inserted the `?` or `!` operator. That said, I took a cautious approach overall so as not to disrupt any behaviour. I even disabled nullable checks in the parser files because I didn't manage to get a good overview about the target state for certain properties and members (null allowed or not). However, I think that's fine for now. This code part hasn't changed for a while. If for some reasons new features need to be implemented, it's a good chance to activate those null checks at that point in time.

### DIFF
--- a/src/GuiRunner/NUnit.UiException.Tests/CodeFormatters/TestPlainTextCodeFormatter.cs
+++ b/src/GuiRunner/NUnit.UiException.Tests/CodeFormatters/TestPlainTextCodeFormatter.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) Charlie Poole and TestCentric contributors.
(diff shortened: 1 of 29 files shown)


CharliePoole (approved):
Looks good for a first pass. I think we should plan on looking at this again when we have time. The guy who wrote this made extreme efforts to ensure that accidentally passing null would not break anything but be ignored and you preserved most of that behavior in this conservative approach. I think you made the right because it's complex code that neither of us is very familiar with.

Even so, most of these calls with null arguments are due to mistakes we can catch at compile time and we should do that eventually. But it would take some time to figure out which of them are mistakes, so leaving it as is until we need to change the code is probably the best call.

PR #1579, opened 2026-09-22
Update refs to latest NUnit assemblies

CharliePoole:
Updates TestCentric to use NUnit.Engine 4.0.0-beta.4.10 as well as the latest versions of dependencies NUnit.Engine.Api, NUnit.Extensibility and NUnit.Extensibility.Api, all of which are now versioned independently from NUnit.Engine.


CharliePoole:
@rowo360 I'll be unavailable until the evening, so feel free to make any needed changes and commit.

Note that the NUnit interface IRuntimeFramework has changed, among other things.

